PR #4, opened 2025-08-27
http: use python3.

vDorst:
Fix compile issue where system don't have python2. I get error below.

```
/bin/sh: line 1: ./treatasm.py: cannot execute: required file not found
make[1]: *** [Makefile:15: httpd.rel] Error 127
```

Debian has python3 a long time so use it.
See https://wiki.debian.org/Python

### DIFF
--- a/httpd/treatasm.py
+++ b/httpd/treatasm.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python3
 import re
 import sys
 


logicog:
You are correct. It is better to use python3 directly. It only works for me because I have the `python-is-python3` package installed.

BTW: The script is merely a workaround for what I believe to be a bug in sdcc, but which I did not have the time and knowledge to fix. The issue is that a statement such as in `uip.c`

```
#pragma codeseg BANK1
...
UIP_LOG("icmp: not icmp echo.");
```

produces the following uip.asm file when compiled:
```
        .area BANK1   (CODE)
        .area CONST   (CODE)
___str_4:
        .ascii "icmp: not icmp echo."
```
The linker will then put the constant string into the CONST segment, which is in BANK0 (0x0000 - 0x3fff), which quickly runs over with data that should be in banked code memory. There is plenty of banked code space but only very little in the common code segment. The script merely removes the `.area CONST (CODE)` line, because I believe sdcc should never have produced this anyway. But maybe I am wrong and there is a different way within sdcc itself to prevent this from happening. Alternatively, it could just be a bug.

PR #5, opened 2025-08-30
fix: comp. warning 283: function declarator with no prototype.

vDorst:
Fix compiler warning 283: function declarator with no prototype.

### DIFF
--- a/httpd/httpd.c
+++ b/httpd/httpd.c
@@ -91,7 +91,7 @@ uint8_t parse_short(register uint16_t *n, __xdata uint8_t * __xdata p)
 }
 
 
-void send_not_found()
+void send_not_found(void)
 {
 	slen = strtox(outbuf, "HTTP/1.1 404 Not found\r\nContent-Type: text/html\r\n\r\n");
 	print_string("slen: "); print_short(slen); write_char('\n');
--- a/rtl837x_port.c
+++ b/rtl837x_port.c
@@ -367,7 +367,7 @@ void port_l2_learned(void) __banked
 /*
  * Basic L2 configuration such as time to forget an entry
  */
-void port_l2_setup() __banked
+void port_l2_setup(void) __banked
 {
 	print_string("\nport_l2_setup called\n");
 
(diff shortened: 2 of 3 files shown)


logicog:
You are absolutely right. I merged your PR, thanks! The output of the build process has become a bit messy and one does not always see all the warnings.

I am using several Hasivo and Keeplink unmanaged switches (you never know what you will really get ordering on AliExpress), and all of them have 2MBit and most of them even 4MBit flash memories, so that no soldering is needed (I am using a SOIC-8 clamp). All the managed version I have so far all use only 2MBit, probably because the Unique ID in the Flash chips makes them expensive and 4MBit without UID probably is the sweet-spot for pricing.

While waiting, you may want to try out the http-simulator in the tools directory: run it under Linux in the html directory and it will to some extent simulate the switch. The web-pages are already much nicer than the ones from Hasivo and co and there is dynamic updating of the throughput, plus as a little detail, the SVG images of the ports in the status line show the same speed-indicators as  the actual switch. If you want to play with that, there is one very important point: No file served can be larger than 2KB at this point (this is not a hard limit, just something to make sure one does not go into wasting valuable RAM in the SoC from the start).

At this point there is something important missing in the firmware: It is not possible to save the configuration in the flash. If you add a config.txt file in the main directory, it will however end up in the image and be called when the switch boots. Anything like the vlan configuration commands are then run automatically.
Mine looks llike this:
```
~/RTLPlayground$ cat config.txt
ip 192.168.2.2
gw 192.168.2.22
netmask 255.255.255.0
```
 I know, I should be working on that with the highest priority, but at this point I am having fun with implementing STP (I figured out how to send a packet from just one port), and anything to do with string-operations are a nightmare without a proper string-library. I don't want to use the one from sdcc, because it uses general pointers and the code produced uses plenty of stack compared to explicitly handling memory spaces.

Can you make your "replace_phy" code a PR? This looks impressive and anything that reduces stack usage is important! I have one comment: You probably looked into `rtl837x_regs.h`. There, depending on REGDBG, different versions of e.g. REG_SET are defined, so that the ifdefs don't end up making the entire inline functions such as `void phy_read_print_args` empty. Do you think it is possible and would make sense to do it again this way? I don't have a strong opinion, I just had initially trouble understanding what was going on with the PHY_READ macro.

PR #6, opened 2025-08-31
Combine two sfr-pairs into a single sfr16 define, so make it less error prone to use and optimize better.

vDorst:
This PR combines two SFR pair into one sfr16.
Instead doing convert two SFR pair into a single uint16_t by hand like:
```c
uint16_t pval = SFR_DATA_8;
pval <<= 8;
pval |= SFR_DATA_0;
```
Now we can use.
```c
uint16_t pval = SFR_DATA_U16;
```

Makes it way less error prone, more readable and also saw that the compiler optimized better.
See some snippets in the commits.

You can define your own order.
let's say `aa = reg_h` and `bb = reg_l`.
* Big endian notation = `__sfr16 __at(aabb) NAME_U16`
* Litte endian notation = `__sfr16 __at(bbaa) NAME_U16LE`

I even think you can use any sfr combination.

Please review `rtl837x_phy.c` closely.
I needed to convert many and/or-operations from 8 to 16-bit.

### DIFF
--- a/rtl837x_phy.c
+++ b/rtl837x_phy.c
@@ -69,10 +69,11 @@ void rtl8224_phy_enable(void) __banked
 	// p001e.0a90:00f3 R02f8-000000f3 R02f4-000000fc P000001.1e000a90:00fc
 	print_string("\r\nrtl8224_phy_enable called\r\n");
 	phy_read(0, 0x1e, 0xa90);
-	uint16_t pval = SFR_DATA_8;
-	pval <<= 8;
-	pval |= SFR_DATA_0;
+	uint16_t pval = SFR_DATA_U16;
 
+
+
+	
 	// PHY Initialization:
 	REG_WRITE(0x2f8, 0, 0, pval >> 8, pval);
 
@@ -83,9 +84,8 @@ void rtl8224_phy_enable(void) __banked
 	phy_write(0x1, 0x1e, 0xa90, pval);
 
 	phy_read(0, 0x1e, 0xa90);
-	pval = SFR_DATA_8;
-	pval <<= 8;
-	pval |= SFR_DATA_0;
+	pval = SFR_DATA_U16;
+
 
 	delay(50);
 	print_string("\r\nrtl8224_phy_enable done\r\n");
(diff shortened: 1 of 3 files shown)


vDorst:
> The SFRs work more like pipes between CPU and ASIC. Also, for example SFR &= 0x01 does not 
work, the CPU needs to do proper moves from and to the register.

This is good to know. Next PR I am using this, I shall change that.

vDorst:
Rebased it on to the latest main.

vDorst:
@logicog, https://github.com/logicog/RTLPlayground/blob/250beae09b9246b726ba4264d808999106572304/rtl837x_phy.c#L83 can be replaced with `phy_write(0, 0x1e, 0xa90, pval);`?

Also you read phy again but don't do anything with the result.
Maybe add comment about the reason.

logicog:
I believe this can be replaced. And then also the reading step can be removed. What is done is only mimicking traces of the ASIC register operations and PHY operations code that I did. I do not always know why something is done. For example the original firmware could be checking the value that has been read. The trace in this case is at the top of the code: "// p001e.0a90:00f3 R02f8-000000f3 R02f4-000000fc P000001.1e000a90:00fc". p means phy-read. P is a write. r is a register read, R is a write. q would be a SerDes read, Q a SerDes write.

vDorst:
Did you try a full rebuild?

```bash
make clean
make
```

I also check it but I don't see something wrong.

logicog:
I retried again after `make clean` and I get the same result. Did you manage to run it on a switch without problems?

vDorst:
Found it.

Was using the wrong data SFR registers.
`SFR_DATA_U16` must use `SFR_DATA_8,a6` and `SFR_DATA_0,a7` not `SFR_DATA_24,a4`, `SFR_DATA_16,a5`.
I think a copy and paste error.

logicog:
Yes, it works! I did some tests like changing port-speed plug/unplug and the likes and it looks really good! I will also test on an 8 + 1 SFP+ port device which has a combination of RTL8224 behind 4 of the ports, just to be sure to test also that code-path. But the sfr16 definitions are definitely working. Great job!

PR #7, opened 2025-08-31
bit_mask lookup before calling phy_write() is expensive.

vDorst:
`bit_mask`-lookup before calling `phy_write()` is expensive.
Move the lookup inside a new function.

* Rename `phy_write()` to `phy_write_mask()` so it can still be used.
* Create new `phy_write()` which accepts `phy_id` and does the phy_mask/bit_mask lookup inside the function.
* Replace all the `phy_write_mask(bit_mask[<var>], ...)` with `phy_write(<var, ...)`.

### DIFF
--- a/rtl837x_common.h
+++ b/rtl837x_common.h
@@ -60,7 +60,7 @@ void print_short(uint16_t a);
 void print_byte(uint8_t a);
 void print_sfr_data(void);
 void print_phy_data(void);
-void phy_write(uint16_t phy_mask, uint8_t dev_id, uint16_t reg, uint16_t v);
+void phy_write_mask(uint16_t phy_mask, uint8_t dev_id, uint16_t reg, uint16_t v);
 void phy_read(uint8_t phy_id, uint8_t dev_id, uint16_t reg);
 void reg_read(uint16_t reg_addr);
 void reg_read_m(uint16_t reg_addr);
--- a/rtl837x_common.h
+++ b/rtl837x_common.h
@@ -61,6 +61,7 @@ void print_byte(uint8_t a);
 void print_sfr_data(void);
 void print_phy_data(void);
 void phy_write_mask(uint16_t phy_mask, uint8_t dev_id, uint16_t reg, uint16_t v);
+void phy_write(uint8_t phy_id, uint8_t dev_id, uint16_t reg, uint16_t v);
 void phy_read(uint8_t phy_id, uint8_t dev_id, uint16_t reg);
 void reg_read(uint16_t reg_addr);
 void reg_read_m(uint16_t reg_addr);
--- a/rtl837x_phy.c
+++ b/rtl837x_phy.c
@@ -105,7 +105,7 @@ void phy_config(uint8_t phy) __banked
 	pval = SFR_DATA_8;
 	pval <<= 8;
 	pval |= SFR_DATA_0 & 0xfe;
-	phy_write_mask(bit_mask[phy], 0x1e, 0x75f3, pval);
+	phy_write(phy, 0x1e, 0x75f3, pval);
 	delay(20);
 
 //	p081e.697a:ffff P000100.1e00697a:ffc1 / p031e.697a:0003 P000008.1e00697a:0001
@@ -114,7 +114,7 @@ void phy_config(uint8_t phy) __banked
 	pval = SFR_DATA_8;
 	pval <<= 8;
 	pval |= SFR_DATA_0 & 0xc0 | 0x01;
-	phy_write_mask(bit_mask[phy], 0x1e, 0x697a, pval);
+	phy_write(phy, 0x1e, 0x697a, pval);
 	delay(20);
 
 //	p031f.a432:0811 P000008.1f00a432:0831
@@ -123,7 +123,7 @@ void phy_config(uint8_t phy) __banked
 	pval = SFR_DATA_8;
 	pval <<= 8;
 	pval |= SFR_DATA_0 | 0x20;
-	phy_write_mask(bit_mask[phy], 0x1f, 0xa432, pval);
+	phy_write(phy, 0x1f, 0xa432, pval);
 
 //	p0307.003e:0000 P000008.0700003e:0001
 	// EEE avertisment 2 register MMMD 7.0x003e, set bit 0: 2.5G has EEE capability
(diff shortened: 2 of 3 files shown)


logicog:
I believe this works. Shown is the memory useg in IRAM and CODE after the change and before:
```
RTLPlayground$ cat rtlplayground.mem
Internal RAM layout:
      0 1 2 3 4 5 6 7 8 9 A B C D E F
0x00:|0|0|0|0|0|0|0|0|a|a|a|a|a|a|a|a|
0x10:|c|c|c|d|d|d|d|d|d|d| | | | | | |
0x20:|B|T|b|b|b|b|b|b|b|b|b|b|b|b|b|b|
0x30:|b|b|b|b|e|e|e|e|e|e|e|e|f|f|f|f|
0x40:|f|f|f|f|f|g|g|g|g|g|g|g|g|g|g|g|
0x50:|g|h|h|h|h|h|h|h|h|h|h|h|h|i|i|i|
0x60:|i|i|i|i|i|i|i|j|j|j|j|j|j|j|j|j|
0x70:|j|j|j|j|j|Q|Q|Q|Q|Q|Q|Q|S|S|S|S|
0x80:|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|
0x90:|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|
0xa0:|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|
0xb0:|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|
0xc0:|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|
0xd0:|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|
0xe0:|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|
0xf0:|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|
0-3:Reg Banks, T:Bit regs, a-z:Data, B:Bits, Q:Overlay, I:iData, S:Stack, A:Absolute

16 bit mode initial stack starts at: 0x7c (sp set to 0x7b) with 132 bytes available.
The largest spare internal RAM space starts at 0x1a with 6 bytes available.

Other memory:
   Name             Start    End      Size     Max     
   ---------------- -------- -------- -------- --------
   PAGED EXT. RAM                         0      256   
   EXTERNAL RAM     0x0001   0x1ad8    6872    16777216
   ROM/EPROM/FLASH  0x0000   0x1c333  45577    16777216


RTLPlayground$ cat rtlplayground.mem_BEFORE
Internal RAM layout:
      0 1 2 3 4 5 6 7 8 9 A B C D E F
0x00:|0|0|0|0|0|0|0|0|a|a|a|a|a|a|a|a|
0x10:|c|c|c|d|d|d|d|d|d|d| | | | | | |
0x20:|B|T|b|b|b|b|b|b|b|b|b|b|b|b|b|b|
0x30:|b|b|b|b|e|e|e|e|e|e|e|e|f|f|f|f|
0x40:|f|f|f|f|f|g|g|g|g|g|g|g|g|g|g|g|
0x50:|g|h|h|h|h|h|h|h|h|h|h|h|h|i|i|i|
0x60:|i|i|i|i|i|i|i|j|j|j|j|j|j|j|j|j|
0x70:|j|j|j|j|j|Q|Q|Q|Q|Q|Q|Q|S|S|S|S|
0x80:|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|
0x90:|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|
0xa0:|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|
0xb0:|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|
0xc0:|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|
0xd0:|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|
0xe0:|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|
0xf0:|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|
0-3:Reg Banks, T:Bit regs, a-z:Data, B:Bits, Q:Overlay, I:iData, S:Stack, A:Absolute

16 bit mode initial stack starts at: 0x7c (sp set to 0x7b) with 132 bytes available.
The largest spare internal RAM space starts at 0x1a with 6 bytes available.

Other memory:
   Name             Start    End      Size     Max     
   ---------------- -------- -------- -------- --------
   PAGED EXT. RAM                         0      256   
   EXTERNAL RAM     0x0001   0x1ad8    6872    16777216
   ROM/EPROM/FLASH  0x0000   0x1c535  46066    16777216
$ ping 192.168.2.2
PING 192.168.2.2 (192.168.2.2) 56(84) bytes of data.
64 bytes from 192.168.2.2: icmp_seq=1 ttl=64 time=66.9 ms
64 bytes from 192.168.2.2: icmp_seq=2 ttl=64 time=32.7 ms
```

489 bytes of FLASH have been saved and the code is much cleaner.

Ping also works and the PHYs are working as expected. Note that ping is slow because I am debugging RX/TX.

There is one remark that I need to make: Flash memory is not really the bottleneck. The bottleneck is the IRAM. That's why there is often in the code repetition instead of using e.g. loops, because it saves the loop variable in IRAM. The exception to that is maybe that there is only 16k of the common code area CODE0. The flash itself will mostly be eaten up by HTML and javascript. A simple `var bgs = psvg.contentDocument.getElementsByClassName("bg");` costs 60 bytes in the flash and I do not really see how this can be shorted. So one really gets mad at the designers of JavaScript for using such long variable and function names. Still, my impression is that FLASH is a lower priority problem.

logicog:
I believe there is a bit of information on the memory in the section on understanding the image using ghidra on the main page. The 8051 CPU is somewhat generic, meaning there are plenty of versions but they have many things in common. I did not have the time to properly document that part.

The best documentation I found is here:
(8051 Memory Spaces)[https://github.com/contiki-os/contiki/wiki/8051-Memory-Spaces]
and
[8051 Code Banking](https://github.com/contiki-os/contiki/wiki/8051-Code-Banking)

CODE0 just means the common code area in the first 16k of Memory, that will allways be in an executable code RAM of the RTL837x processors. Everything that is banked (CODE1 in flash 0x4000-0xffff and CODE2 0x10000-0x1a000 and so on) is read on-the-fly from the flash via SPI. I do not know how much is being cached, but listening on that allowed the cryptographic code verification and dongling to the flash to be hacked here:
https://github.com/up-n-atom/SWTG118AS/issues/4

I have not had any experience on the 8051 either before starting this and not much experience with machine code apart maybe some things on the MSP430. But the simplicity and the fact that one can still do so many things with the 8051 are a lot of fun!

Compression and marking with a different content type is an excellent idea! You could try that with the simulator tool. Unfortunately the simulator does not read the flash image, only what is in the html directory, but maybe you have an idea. This could be the key to a really cool Web-Interace!

PR #15, opened 2025-09-06
Docs: Add SWTG024AS-v2.0-unmanaged version

vDorst:
Pinout Spreedsheet, photos and wiki about my findings for the SWTG024AS-v2.0-unmanaged.

GPIO number may be has endian problem! Maybe have to fix the code of the wiki.

What do you think about the layout?


logicog:
I like the layout a lot, the amount of information gathered is amazing. Also the use of the sub-directories is cool. I just fear that I might now be expected to document the devices I have similarly well...

vDorst:
Now it is ready. 
Updated the GPIO numbering and put it in a table.

PR #16, opened 2025-09-06
Refactor GPIO defines and add cmd `regget`, `regset`, `gpio`.

vDorst:
Refactor the GPIO defines, according to this [rtl-sdk rtl8373 sourcecode](https://gitee.com/hanwckf/rtl8372/raw/master/src/rtk-api/dal/rtl8373/rtl8373_reg_definition.h)
We have only two 32-bit registers to control the 64 gpios of the SOC.
Not three.

Also added serial commands
* `regget`: read a register.
* `regset`: write hex-value to register.
* `gpio`: show current value and changes with the previous readout.


logicog:
It looks very good! I did not test it on a device so far.

PR #17, opened 2025-09-08
STP: initial work

logicog:
This adds initial support for the (R/M)STP protocols. It adds commands to enable and disable STP In the enabled state, incoming STP packets are parsed and the root-bridge state is updated. Additionally, every 2 seconds HELLO packets are sent which contain the current belief of the STP.

A lot of the code for a proper STP implementation is missing such as all the state machines. The purpose of this early PR is therefore more to show how incoming packets can be handled with what UIP calls a device driver and that incoming and outgoing packets have a different frame formats due to RTL837x adding and requiring different kinds of tagging. Also shown is how to send packets only on 1 port, which is needed because STP-CNF packages refer to the port they have been sent on.

Note that the central tick timer's frequency seems to be off. This PR only handles the current situation, it does not try to fix it.

### DIFF
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -14,6 +14,7 @@
 #include "rtl837x_phy.h"
 #include "rtl837x_regs.h"
 #include "rtl837x_sfr.h"
+#include "rtl837x_stp.h"
 #include "uip/uip.h"
 
 #pragma codeseg BANK1
@@ -23,6 +24,7 @@ extern __xdata uint8_t maxPort;
 extern __xdata uint8_t nSFPPorts;
 extern __xdata uint8_t isRTL8373;
 extern __xdata uint16_t mpos;
+extern __xdata uint8_t stpEnabled;
 
 extern volatile __xdata uint32_t ticks;
 extern volatile __xdata uint8_t sfr_data[4];
@@ -573,6 +575,17 @@ void cmd_parser(void) __banked
 				port_l2_forget();
 			else
 				port_l2_learned();
+		}
+		if (cmd_compare(0, "stp")) {
+			if (cmd_words_b[1] > 0 && cmd_compare(1, "on")) {
+				print_string("STP enabled\n");
+				stpEnabled = 1;
+				stp_setup();
+			} else {
+				print_string("STP disabled\n");
+				stp_off();
+				stpEnabled = 0;
+			}
 		}
 		if (cmd_compare(0, "pvid") && cmd_words_b[1] > 0 && cmd_words_b[2] > 0) {
 			__xdata uint16_t pvid;


rtlplayground.c
```diff
@@ -54,6 +54,10 @@ __xdata uint8_t isRTL8373;
 volatile __xdata uint32_t ticks;
 volatile __xdata uint8_t sec_counter;
 volatile __xdata uint16_t sleep_ticks;
+__xdata uint8_t sfp_clock;
```
vDorst: seems a typo.
sfp_clock. I think you mean stp_clock?
So second letter should be a `t` instead of a `f`.

rtl837x_stp.c
```diff
@@ -103,4 +197,55 @@ void stp_cnf_send(uint8_t port) __banked
...
+		}
+	}
+}
+
+
+void stp_setup(void) __banked
+{
+	print_string("Enabling STP: ");
+	// Read STP port states: 00 disable, 01 blocking, 10 learning, 11 forwarding
+	reg_read_m(RTL837X_MSTP_STATES);
```
vDorst: Do you need to read the register? If so, maybe add a note about it.
Because line below you clear all the data.
Same in `stp_off()`.

rtlplayground.c
vDorst: BTW: Same for define `SFP_TICK_DIVIDER`.

rtl837x_stp.c
```diff
@@ -103,4 +197,54 @@ void stp_cnf_send(uint8_t port) __banked
...
+			stp_cnf_send(i);
+		}
+	}
+}
+
+
+void stp_setup(void) __banked
+{
+	print_string("Enabling STP: ");
+	// Read STP port states: 00 disable, 01 blocking, 10 learning, 11 forwarding
```
vDorst: the comment `// Read ...` has less of a meaning because `reg_read_m()` is removed.
Same also happens in `stp_off()`.

Maybe demagic some values.

rtl837x_stp.c
```diff
@@ -103,4 +197,54 @@ void stp_cnf_send(uint8_t port) __banked
...
+}
+
+
+void stp_off(void) __banked
+{
+	// Read STP port states: 00 disable, 01 blocking, 10 learning, 11 forwarding
+	reg_read_m(RTL837X_MSTP_STATES);
+	sfr_data[0] = sfr_data[1] = sfr_data[2] = sfr_data[3] = 0;
+	for (uint8_t i = minPort; i <= maxPort; i++) {
+		uint8_t bit_mask = 0b11 << ( (i << 1) & 0x7);
```
vDorst, line 247: deadcode? bit_mask not used.

logicog:
Ouch, should have tested a bit more thoroughly. I moved the comments and they should now be more meaningful. Also fixed the forgotten use of the bit_mask.

logicog:
Fixed the appearance of SFP_ instead of STP_.

PR #19, opened 2025-09-09
Fixes and code size reduction.

vDorst:
fixes:
* itoa_html(): does not print the middle zero.
* REG_SET(): compiler warning.

Code gen optimize
* don´t inline the push to html buffer functions, saves around 2.5kbytes.


logicog:
I see that you modify files repeatedly. You can avoid that this becomes part of the PR:
First, commit  a fix to an existing commit.
Say, you want to fix the last 4 commits, you do: `git rebase -i HEAD~5` It is 5 because you want to include your fix.
In the list of commits you can now in the editor re-order the commits and especially move your fix below the faulty commit. Change "pick" to "f" (for fix).
Now you can force-push to the PR so that the fix becomes invisible, and all commits look clean:
`git push -f origin <pr-branch>`

PR #20, opened 2025-09-10
WIP: Add initial installer code

logicog:
This is a work in progress to add support for installing RTLPlayground on a managed switch via the Web-Interface.
Currently an updatebuilder binary is provided which can create update images which are accepted by an existing OEM firmware as valid firmware update images. This has been validated.
The image consists of installer code as found in installer.c to be located at 0x1002 in the flash (0x1000 in the CODE segment) plus the actual RTLPlayground firmware image.
The actual image is supposed to be written to 0x001DF000 in flash when the update is installed by the firmware updater.
When the switch reboots after the update, the installer will be called and then will move the actual update image from0x001DF000 to 0x0000 in flash. This part is not yet implemented.


vDorst:
Thanks, I also did a full upgrade path and it works.

PR #21, opened 2025-09-14
WIP: Add a web-updater

logicog:
This is a WIP-PR on top of PR #20
It adds code to create update functionality, so that the RTLPlayground image can be updated over http.
The work will require to:
- Consolidate the flash writing code
- Document the existing OEM update logic, explain RTLPlayground strategy
- Add upload functionality to the httpd server
- Add upload functionality in the web-pages
- Add update logic into the code-base.


logicog:
I fixed the circular dependency in the Makefile and incorrect invocation of the updatebuilder. Fixed spelling error in SFR_FLASH_DUMMYCYCLES

logicog:
I learned something strange:
```
void flash_write_bytes(__xdata uint32_t addr, __xdata uint8_t *ptr, __xdata uint16_t len)
{
...
}

__xdata uint32_t addr = 0x2000;
__xdata uint8_t *ptr = 0x0000;
flash_write_bytes (addr, ptr, 1);
```
Does not work, because sdcc is not able to properly initialize the `len`variable in __xdata when calling `flash_write_bytes`. Very strange. Workaround is to use a variable `__xdata uint16_t len` and set it to `1`in the code (not initialize it to 1 since that also does not appear to work).

logicog:
The PR is now ready for review. It is rather Minimum Viable, so feedback is most welcome!

logicog:
Rebased to main and all commits cleaned up.

html/update.html
```diff
@@ -0,0 +1,20 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <link rel="stylesheet" href="style.css">
+    <title>Firmware update</title>
+  </head>
+  <body>
+     <ul> <li><a href="index.html">Overview</a></li> <li><a href="stat.html">Port Statisitcs</a></li>
```
vDorst: NIT: space between <ul> and <li>

httpd/httpd.c
```diff
@@ -98,28 +143,177 @@ void send_not_found(void)
...
+
+__xdata uint8_t *scan_header(__xdata uint8_t *p)
+{
+	content_type = 0;
+
 	while (*p != '\r' || *(p + 1) != '\n' || *(p + 2) != '\r' || *(p + 3) != '\n') {
 		write_char(*p);
 		if (!*p++)
 			break;
+		if (*p == '\n' && is_word(p + 1, "Content-Type:"))
```
vDorst: This can be just `if (is_word(p, "\nContent-Type:"))`

httpd/httpd.c
```diff
@@ -77,6 +92,36 @@ char strcmp(__xdata uint8_t * __xdata c, __code uint8_t * __xdata d)
...
+
+	if (d[i])
+		return 0;
+	if (c[i] != ' ' && c[i] != '\t' && c[i] != ':' && c[i] != '?' && c[i] != '=' && c[i] != '\n' && c[i] != '\r' && c[i])
+		return 0;
+	return 1;
+}
+
+
 uint8_t parse_short(register uint16_t *n, __xdata uint8_t * __xdata p)
```
vDorst: Swap the two parameters and put n also in __xdata saves three sram bytes.
pointer `p` doesn't need to be on xdata. Because calling convention already put 16bit value in the DPTR register which can be used in the function itself.

This also applies to other function, you don't have to put the first xdata ptr argument also in xdata.

Also make the `vlan`  value static in xdata and directly use it in the function also saves 1 byte.

Like this.
```patch
+__xdata uint16_t vlan_parse;
 
-uint8_t parse_short(register uint16_t *n, __xdata uint8_t * __xdata p)
+uint8_t parse_short(__xdata uint8_t *p)
 {
        uint8_t err = 1;
-       *n = 0;
+       uint8_t c = 0;
+       vlan_parse = 0;
 
-       while (*p >= '0' && *p <= '9') {
+       while(1) {
+               c = *p++ - '0';
+               if (c > 9) { break; }
                err = 0;
-               *n = (*n * 10) + *p - '0';
-               p++;
+               vlan_parse = (vlan_parse * 10) + c;
        }
        return err;
 }
@@ -401,9 +404,8 @@ void httpd_appcall(void)
                        if (!strcmp(q, "/status.json")) {
                                send_status();
                        } else if (!strcmp(q, "/vlan.json")) {
-                               __xdata uint16_t vlan;
-                               parse_short(&vlan, q + 15);
-                               send_vlan(vlan);
+                               parse_short(q + 15);
+                               send_vlan(vlan_parse);
                        } else if (!strcmp(q, "/counters.json")) {
                                send_counters(q[19]-'0');
                        } else {
```

httpd/httpd.c
```diff
@@ -77,6 +92,36 @@ char strcmp(__xdata uint8_t * __xdata c, __code uint8_t * __xdata d)
 }
 
 
+char is_word(__xdata uint8_t * __xdata c, __code uint8_t * __xdata d)
```
vDorst: My feeling is that this function is just heavy for most checks.

I was thinking about `starts_with()`, instead of adding extra length argument.
Put the length in the first byte of array `d`.
But maybe that is over optimizing things.

The funcion
```c
// check if array c starts with array d.
// First byte of d is the length byte.
// return 0 when dont match else the len value.
char starts_with(__xdata uint8_t * c,  __code uint8_t * __xdata d)
{
	uint8_t len = *d++;

	do {
		if (*c++ != *d++) {
			return 0;
		};
		len--;
	} while(len);

	return len;
}
``` 

Use in the code
```c
if (starts_with(request_path, "\x03cmd")) ....
```
But feel free to ignore this.

rtl837x_flash.c
```diff
@@ -408,18 +408,18 @@ void flash_sector_erase(uint32_t addr)
 	while (flash_read_status() & 0x1);
 
 	flash_configure_mmio();
-
 }
 
 
-void flash_write_bytes(__xdata uint32_t addr, __xdata uint8_t *ptr, uint16_t len)
+void flash_write_bytes(__xdata uint32_t addr, __xdata uint8_t *ptr, __xdata uint16_t len)
```
vDorst: Like `sfr_data`, place the `flash_address` also in a global __xdata value.
Update that value before every `flash_xxx()`.

I compiled tested these five functions and saves me 10 bytes of sram.

```patch
-void flash_read_bulk(register __xdata uint8_t *dst, __xdata uint32_t src, register uint16_t len);
-void flash_write_bytes(__xdata uint32_t addr, __xdata uint8_t *ptr, __xdata uint16_t len);
-void flash_find_mark(__xdata uint32_t src, __xdata uint16_t len, __code uint8_t *mark);
+void flash_read_bulk(__xdata uint8_t *dst, __xdata uint16_t len);
+void flash_write_bytes(__xdata uint8_t *ptr, __xdata uint16_t len);
+void flash_find_mark(__code uint8_t *mark, __xdata uint16_t len);
-void flash_dump(register uint32_t addr, register uint8_t len);
+void flash_dump(register uint8_t len);
-void flash_read_security(uint32_t addr, uint8_t len);
+void flash_read_security(uint8_t len);
```

Maybe change `flash_address` to a struct with address and length.
Uploaded the playground test code. See https://github.com/vDorst/RTLPlayground/tree/web-updater1

rtl837x_flash.c
```diff
@@ -408,18 +408,18 @@ void flash_sector_erase(uint32_t addr)
...
 {
-	uint8_t exit_loop = 0;
+	__xdata uint8_t exit_loop = 0;
 
+	write_char('>'); print_long(addr); write_char(':'); print_short(len); write_char('-'); print_byte(*ptr); write_char('\n');
 	while(1) {
 		flash_write_enable();
 		SFR_FLASH_CMD = CMD_PAGE_PROGRAM;
-		SFR_FLASH_TCONF = 0x40 | 8 | 4; // Bytes written is is 4, 8 enables write, 0x40 is unkown
+		SFR_FLASH_TCONF = 0x40 | 8 | 4; // Bytes written is 4, 8 enables write, 0x40 is unknown
```
vDorst, line 345: I am very sure that bit `0x40` tells the SPI peripheral that this is not the last data. So the hardware keeps the chipselect line low. Also line 425 seems to confirm this. Because bit 0x40 is not set and also len maybe low then 4.

httpd/httpd.c
```diff
@@ -98,28 +143,177 @@ void send_not_found(void)
 	slen += strtox(outbuf + slen, "<!DOCTYPE HTML PUBLIC>\n<title>404 Not Found</title>\n<h1>Not Found</h1>\n");
```
vDorst: Seems all static data, why not precompute it to one line of static data?

httpd/httpd.c
```diff
@@ -98,28 +143,177 @@ void send_not_found(void)
...
 }
 
-void handle_post(void)
+
+void send_bad_request(void)
 {
-	__xdata uint8_t *p = uip_appdata;
+	slen = strtox(outbuf, "HTTP/1.1 400 Bad Request\r\nContent-Type: text/html\r\n\r\n");
+	print_string("slen: "); print_short(slen); write_char('\n');
+	slen += strtox(outbuf + slen, "<!DOCTYPE HTML PUBLIC>\n<title>400 Bad Request</title>\n<h1>Bad Request</h1>\n");
```
vDorst: Same here, seems static data.

httpd/httpd.c
vDorst: `parse_short` result if never checked.

httpd/httpd.c
logicog: Yes. Well spotted! Will fix as suggested.

httpd/httpd.c
logicog: WOW! Will fix as suggested.
```
$ diff output/rtlplayground.mem output/rtlplayground.mem_old
10c10
< 0x70:|j|j|j|j|Q|Q|Q|Q|Q|Q|Q|S|S|S|S|S|
---
> 0x70:|j|j|j|j|j|j|j|j|Q|Q|Q|Q|Q|Q|Q|S|
21c21
< 16 bit mode initial stack starts at: 0x7b (sp set to 0x7a) with 133 bytes available.
---
> 16 bit mode initial stack starts at: 0x7f (sp set to 0x7e) with 129 bytes available.
28,29c28,29
<    EXTERNAL RAM     0x0001   0x1c87    7303    16777216
<    ROM/EPROM/FLASH  0x0000   0x1b929  46662    16777216
---
>    EXTERNAL RAM     0x0001   0x1c89    7305    16777216
>    ROM/EPROM/FLASH  0x0000   0x1b95c  46713    16777216
```

html/update.html
logicog: I don't see what you mean. Can you write out differently what you suggest to change?

rtl837x_flash.c
logicog: I will confirm this with a logic analyzer. The writes are also very inefficient at the moment, since at the beginning of the loop
```
flash_write_enable();
```
is called.
This first checks whether the flash is busy (which it sometimes is after the writes, I can see that in the SPI traces) and then checks the write latch, which is always on. In theory it should be sufficient to check whether the write is enabled before the while loop and at the end of the loop whether the chip is still busy.

httpd/httpd.c
logicog: Since it is a web-page, I could store it in a file and use the fileadder. For the HTTP header I was thinking of this recipe: https://stackoverflow.com/questions/5022067/strlen-in-the-c-preprocessor
I need to test.

rtl837x_flash.c
logicog: Need to test. But why don't you do a PR with the code you already have?

html/update.html
vDorst: Lol, I see now that github interprets the html code.
There is a space between `<ul> <li>`.

rtl837x_flash.c
vDorst: I was just freewheeling while reviewing code.
I don't want to interfere with your work while you're still working on it.
Cross PRs are just difficult and only cause confusion.
The same thing happens with the “declutter” PR, where I've already modified code that's still being worked on.

I've uploaded my current freewheel code so you can see what I've done and you can choose for yourself what you want to take out, which also saves some work.

rtl837x_flash.c
vDorst: Also setting the `SFR_FLASH_ADDR{16,8,0|` every cycle should not be necessary.
It is only using on the start of the SPI transmission and the memory auto increments this value internally.

This can also applied to `flash_write_bytes()`.

From the W25Q32JV datasheet, you can program 256 bytes in one transaction.
``` 
The Page Program instruction allows from one byte to 256 bytes (a page) of data to be programmed at previously erased (FFh) memory locations
```

httpd/httpd.c
vDorst: In the meantime it can just be written as
```c
slen = strtox(outbuf, "HTTP/1.1 404 Not found\r\nContent-Type: text/html\r\n\r\nslen: 0x39\n<!DOCTYPE HTML PUBLIC>\n<title>404 Not Found</title>\n<h1>Not Found</h1>\n");
```

httpd/httpd.c
```diff
@@ -98,28 +143,177 @@ void send_not_found(void)
...
+		}
+		// Have we reached the end of the part?
+		if (!boundary[bindex]) {
+			s->tstate = TSTATE_NONE;
+			print_string("len 2: "); print_short(write_len); write_char(' ');
+			flash_write_bytes(uptr, flash_buf, write_len);
+			uptr += write_len;
+			write_len = 0;
+			// TODO: This is a bit premature, what about a nice web-page saying the device will reset???
+			reset_chip();
```
vDorst, line 227: Maybe add a debug console print "Flashing done, reset device".

httpd/httpd.c
```diff
@@ -98,28 +143,177 @@ void send_not_found(void)
...
+		if (bptr >= uip_len) {
+			s->tstate = TSTATE_POST;
+			return 1;
+		}
+		// Have we reached the end of the part?
+		if (!boundary[bindex]) {
+			s->tstate = TSTATE_NONE;
+			print_string("len 2: "); print_short(write_len); write_char(' ');
+			flash_write_bytes(uptr, flash_buf, write_len);
+			uptr += write_len;
```
vDorst, line 223: Add a check that checks the flash status that it is done writing.
Also clear the write enable bit.

html/update.html
logicog: Ahh. ;-) Will fix as suggested.

httpd/httpd.c
logicog: I believe the write done check is part of what flash_write_bytes does. At the end it does:
```
	while (flash_read_status() & 0x1);
	flash_configure_mmio();
```
Also, the datasheet you mention above states: "After the Page Program cycle has finished the Write Enable Latch (WEL) bit in the Status Register is cleared to 0.". But probably this would need to be managed if the feature of writing up to 256 bytes consecutively is used. The firmware flashing is very inefficient at the moment as seen tracing the SPI communication and could be considerably sped up. This will need more work.

httpd/httpd.c
logicog: Will fix as suggested. Plus maybe add the web-page already.

httpd/httpd.c

httpd/httpd.c
logicog: Or that. Will fix as suggested.

rtl837x_flash.c
vDorst: With our last email discussion I think what I said above about SPI transfer `0x40` bit.
I think this is incorrect, it seems for now that we only can send up to 4 bytes at the time as a full SPI transfer. So chipselect low, CMD, ADDRESS, optional dummycycles, to up 4 bytes, chipselect high.

config.txt
logicog, line 1: We will need to mention these defaults in the documentation.

PR #23, opened 2025-09-27
Declutter building artifacts

vDorst:
Declutter sourcecode from building artifacts.
Every "project" puts the build artifacts into relative `BUILDDIR`.

Currently this is on top of PR#21.


vDorst:
> Removing rtlplayground.mem

Still think it is a good metric to see mem-usage change.

I wanted html_data.{c,h} also in `BUILDDIR` but that is not easy because of `fileadder` see the outputname as one thing. Which is fine I think.

Also passing the `BUILDDIR` to other Makefile would be great. But I don't really know how.
For now it is very static path, so I think it is good enough otherwise feel free to improve it.

logicog:
I tested building the flash image and the update from OEM image then cleaning and building again in various sequences. It works and is much cleaner. Great work!

PR #36, opened 2025-10-31
WIP: System settings and configuration management

logicog:
This is some initial code for system settings and configuration management so that the concept can be discussed

With the current code the IP settings of the device can be changed with the new System tab and the initial idea for how to manage writing the configuration into Flash is started in code.

The idea for persisting the configuration to flash is as follows:
- The System page downloads the current configuration file (config.txt) from the device flash as a plain file (code is there for the simulator)
- The file is parsed into a configuration structure (partially implemented)
- The current configuration according to the web-session, which is stored by all the tabs into a client-side JS SessionStore, is modified by the configuration structure, taking in particular account of removing obsolete or overwritten configurations, i.e.
  - `mirrorr off` overwrites a mirror config command and nothing is stored in persistent configuration because no mirror is the default state
- The configuration structure is marshalled into a new configuration text-file, which gets uploaded (possibly after displaying it?) to the flash configuration area, reusing the code for the firmware image upload

This means, that when e.g. the VLAN is configured, the tab will create/update all the VLAN settings in the SessionStore client-side storage in addition to sending the relevant commands to the device. The session store will therefore reflect all the currently un-applied but active settings on the device which can be commonly edited by all setting tabs in the web-client. That structure is then persisted in the System tab after it is applied on top of the current configuration file of the device.

The SessionStore will store a JSON structure of the settings.
The SessionStore JSON will have the same format as the temporary parsed data from the current configuration file.
The JSON will be an ordered list/array of entries for a configuration that should make it easy to verify whether a new command/setting obsoletes or modifies a current one.

There is currently a hard limit of about 2.2kB on the size of a single file served by the device (e.g. a .js file). It should be possible to implement chunking of these files (not the dynamic ones, i.e. the .json!) in order to support more complex .js needed on the client side for the tabs.

### DIFF
--- a/httpd/httpd.c
+++ b/httpd/httpd.c
@@ -8,7 +8,7 @@
 #include "uip.h"
 #include "../html_data.h"
 
-#define SESSION_ID "1234567890ab"
+#define SESSION_ID_LENGTH 12
 #define SESSION_TIMEOUT 200
 
 // Upload Firmware to 1M
@@ -23,6 +23,7 @@
 #pragma constseg BANK1
 
 extern volatile __xdata uint8_t sfr_data[4];
+extern __code uint8_t * __code hex;
 extern __code struct f_data f_data[];
 extern __code char * __code mime_strings[];
 extern __xdata struct flash_region_t flash_region;
@@ -53,7 +54,7 @@ __xdata uint32_t max_upload;
 __xdata uint16_t short_parsed;
 
 __xdata char passwd[21];
-__xdata char session_id[13];
+__xdata char session_id[SESSION_ID_LENGTH + 1];
 __xdata uint8_t authenticated;
 __xdata uint32_t now;
 __xdata uint8_t *timeptr;
@@ -217,7 +218,7 @@ __xdata uint8_t *scan_header(__xdata uint8_t *p)
 			session = p + 17;
 	}
 	if (content_type && is_word(content_type, "multipart/form-data; boundary")) {
-		print_string("\nFound multiplart\n");
+		print_string("\nFound multipart\n");
 		content_type += 30;
 		uint8_t i = 0;
 		while (content_type[i] != '\r' && content_type[i] != '\n') {
@@ -232,23 +233,33 @@ __xdata uint8_t *scan_header(__xdata uint8_t *p)
 		boundary[i + 4] = 0;
 	}
 
-	reg_read_m(RTL837X_REG_SEC_COUNTER);
-	timeptr = (uint8_t*)&now; // Now is Little endian
-	timeptr[0] = sfr_data[3]; timeptr[1] = sfr_data[2]; timeptr[2] = sfr_data[1]; timeptr[3] = sfr_data[0];
+	read_reg_timer(&now);
 
 	if (session) {
-		print_string("Session: "); print_string_x(session); print_string(", time now "); print_long(now);
-		print_string(" last session use: "); print_long(last_session_use); write_char('\n');
 		if (now - last_session_use > SESSION_TIMEOUT) {
 			print_string("Session expired\n");
 		} else {
-			if (is_word(session, SESSION_ID))  // TODO: is_word_x
+			if (is_word_x(session, session_id))
 				authenticated = 1;
 			else
 				print_string("Invalid session cookie!\n");
 		}
 	}
 	return p;
+}
+
+
+void gen_random_bytes(__xdata uint8_t *b, uint8_t bytes)
+{
+	__xdata uint8_t i = 0;
+	while (bytes) {
+		if (!i)
+			get_random_32();
+		b[--bytes] = itohex(sfr_data[i]);
+		if (!bytes) { break; }
+		b[--bytes] = itohex(sfr_data[i] >> 4 | sfr_data[i] << 4);
+		i = (i + 1) & 0x3;
+	}
 }
 
 
@@ -373,12 +384,14 @@ void handle_post(void)
 		p += 8; // Read also over "pwd="
 		if (is_word_x(p, passwd)) {
 			print_string("Password accepted!\n");
-			reg_read_m(RTL837X_REG_SEC_COUNTER);
-			timeptr = (uint8_t*)&last_session_use; // last_session_use is Little endian
-			timeptr[0] = sfr_data[3]; timeptr[1] = sfr_data[2]; timeptr[2] = sfr_data[1]; timeptr[3] = sfr_data[0];
-
+			read_reg_timer(&last_session_use);
+			gen_random_bytes(session_id, SESSION_ID_LENGTH);
+			session_id[SESSION_ID_LENGTH] = '\0';
 			slen = strtox(outbuf, "HTTP/1.1 302 Found\r\nLocation: index.html\r\n" \
-					      "Set-Cookie: session=" SESSION_ID "; SameSite=Strict\r\n\r\n");
+					      "Set-Cookie: session=");
+			for (register uint8_t i = 0; i < SESSION_ID_LENGTH; i++)
+				outbuf[slen++] = session_id[i];
+			slen += strtox(outbuf + slen, "; SameSite=Strict\r\n\r\n");
 		} else {
 			slen = strtox(outbuf, "HTTP/1.1 302 Found\r\nLocation: login.html\r\n\r\n");
 		}


logicog:
I fixed the configuration command filtering and tested with the following configuration:
```
ip 192.168.10.247
gw 192.168.10.1
gw 192.168.10.3
gw 22.22
bla
netmask 255.255.255.0
netm 255.255.255.0
netmask 222.22
ip 10.0.0.1
vlan 3 4t 5
vlan 5 4t 5
vlan 3 2t 4
eee on
eee off
eee 1 off
eee 2 on
eee 1 on
mirror 3 4 2t
mirror 2 5
```
This converts to a filtered configuration for flash writing of:
```
gw 192.168.10.3
netmask 255.255.255.0
ip 10.0.0.1
vlan 5 4t 5
vlan 3 2t 4
eee off
eee 2 on
eee 1 on
mirror 2 5
```

vDorst:
At the first page load, the device gives way too much data.

Logfile [pr36-5.log](https://github.com/user-attachments/files/23756261/pr36-5.log)

Screenshot_20251125_215929

Screenshot_20251125_220250

html/config.js
```diff
@@ -0,0 +1,55 @@
+var configInterval = Number();
+var configuration = [];
+const conf_cmds = [
+  /ip\s+\d+(\d{1,3}\.){3}\d{1,3}/, /gw\s+\d+(\d{1,3}\.){3}\d{1,3}/, /netmask\s+\d+(\d{1,3}\.){3}\d{1,3}/,
+  /eee(\s+\d)?\s+(on|off)/, /mirror(\s+\d|10)(\s+(\d|10)(t|r)?)+/, /vlan\s+(\d{1,4})(\s+(\d|10)(t|u)?)+/
```
vDorst: ip regex accept more than just ip.
`ip 11232347234687648726487242364827.123.123.123` is valid because of the `ip\s+\d+` `\d+` should be removed.

You can check this on `regex101.com`. Select Flavor `Java 8`, set Regex Flags: `globel`.

Same for mirror command, these are all valid.

```
mirror10 1
mirror 1 1t
mirror 2 2r
mirror 3 4t 2t
mirror 2 5
```

I am not sure what to do with `(\d|10)`

logicog:
> At the first page load, the device gives way too much data.

This is the browser. When the switch boots, the link comes up and at that point the browser sends all the penned up status requests which could not be sent before because the network being unavailable at once.

logicog:
I have also seen a first corrupted page once. Lots of strange characters at the end of the page. Need to catch it on wireshark. Maybe it has something to do with the cache of the browser being dirty. I now always use shift-reload on the page once the device has booted.

html/config.js
logicog: > Same for mirror command, these are all valid.

The intention was not to prevent _all_ invalid commands. It is like a unix configuration file: you are allowed to put also incorrect stuff in there. The real test is when the commands are executed by the command interpreter. The verification of the command is more important for the second stage of the configuration writing, namely whether one configuration overwrites a previous one. For that to work, the commands have to be relatively well-formed, and that is why they are checked for following a certain form, not to prevent any malformed command.
Maybe that is not the best idea, but for me it was important that the configuration only always has the last valid configuration of a property, i.e. eee on overwrites eee off.

logicog: > I am not sure what to do with `(\d|10)`

Port 10 is the CPU port. You can mirror packets to the CPU port, for example to do statistics on the device itself. Of course this may be very dangerous as the device could DoSed, but there is a possibility to rate limit and sample mirrored packages, so it may actually make more sense when this is implemented. Note that the OEM firmware does not allow to do that, but I have seen it among the registers.

cmd_parser.c
```diff
@@ -706,21 +745,55 @@ void cmd_parser(void) __banked
...
+#define FLASH_READ_BURST_SIZE 0x100
+#define PASSWORD "1234"
 void execute_config(void) __banked
 {
-	memcpyc(flash_buf, "test", 5);
-	print_string_x(flash_buf);
+	__xdata uint32_t pos = CONFIG_START;
+	__xdata uint16_t len_left = CONFIG_LEN;
+
+	// Set default password, it can be overwritten in the configuraiotn file
```
vDorst: typo: configuraiotn

html/config.js
```diff
@@ -0,0 +1,55 @@
+var configInterval = Number();
+var configuration = [];
+const conf_cmds = [
+  /ip\s+(\d{1,3}\.){3}\d{1,3}/, /gw\s+(\d{1,3}\.){3}\d{1,3}/, /netmask\s+(\d{1,3}\.){3}\d{1,3}/,
+  /eee(\s+\d)?\s+(on|off)/, /mirror(\s+\d|\s+10)(\s+(\d|10)(t|r)?)+/, /vlan\s+(\d{1,4})(\s+(\d|10)(t|u)?)+/
```
vDorst: `/mirror(\s+\d|\s+10)` to  `/mirror\s+(\d|10)` so it matches other cases and `mirror <num>` is wanted syntax.

html/login.html
```diff
@@ -0,0 +1,30 @@
...
+</head>
+
+<body class="login_page">
+  <div class = "center">
+  <h1> RTL Switch Login</h1>
+  <form method="post" action="login">
+  <div class="txt_field">
+    <input name="pwd" type="password" onclick="removeNote()" required />
+    <span> </span>
+    <label> Password </label>
```
vDorst: Are the extra spaces needed inside the tags-value?

httpd/httpd.c
```diff
@@ -198,9 +232,45 @@ __xdata uint8_t *scan_header(__xdata uint8_t *p)
...
+				print_string("Invalid session cookie!\n");
+		}
+	}
 	return p;
 }
 
+void session_id_new(void)
+{
+	// In order to get a new random numner, this bit has to be set each time!
+	reg_bit_set(RTL837X_RLDP_RLPP, RLDP_RND_EN);
```
vDorst: I am thinking just to make a function that just generate u32 value.
Because is it worth it to read the to extra bytes?

Just make one function to generate u32 random value seems enough?
```c
// Get random u32 value, value is stored in sfr_data[].
void get_random_u32(void) {
    // In order to get a new random numner, this bit has to be set each time!
	reg_bit_set(RTL837X_RLDP_RLPP, RLDP_RND_EN);
    reg_read_m(RTL837X_RAND_NUM0);
}
```

I also wondering if we can make this more general? Two functions that can generate/fill a hex-char-array and use a array.
I think we can use `itohex()` here too .
`itohex()` does the masking for you, just byteswap the value every time. Like `byte_to_html()`.

tools/httpd_sim.c
```diff
@@ -253,13 +293,30 @@ void send_bad_request(int socket) {
...
 char *scan_header(char *p)
 {
+	session = 0;
+	authenticated = false;
 	while (*p != '\r' || *(p + 1) != '\n' || *(p + 2) != '\r' || *(p + 3) != '\n') {
 		if (!*p++)
 			break;
 		if (*p == '\n' && is_word(p + 1, "Content-Type:"))
 			content_type = p + 15;
+		if (*p == '\n' && is_word(p + 1, "Cookie:"))
```
vDorst: Although it is the simulator code.
This can we rewritten as `if (is_word(p, "\nCookie:"))`.
Same as if-statement above.

cmd_parser.c
logicog: Will fix.

html/config.js
logicog: Yes, absolutely, will fix.

html/login.html
logicog: No. Will fix.

httpd/httpd.c
logicog: I'll change the session_id_new to fill a hex-array and add the 32-bit get random as flash space is cheap, plus this can be much more useful in the future.

tools/httpd_sim.c
logicog: Yes, will fix.

logicog:
All fixes done and rebased.

httpd/httpd.c
```diff
@@ -175,16 +205,20 @@ __xdata uint8_t *skip_boundary(__xdata uint8_t *p)
...
+	session = 0;
+	authenticated = 0;
 
 	while (*p != '\r' || *(p + 1) != '\n' || *(p + 2) != '\r' || *(p + 3) != '\n') {
 		write_char(*p);
 		if (!*p++)
 			break;
 		if (is_word(p, "\nContent-Type:"))
 			content_type = p + 15;
+		if (*p == '\n' && is_word(p, "\nCookie:"))
```
vDorst: `*p == '\n'` is redundant.

httpd/httpd.c
```diff
@@ -198,10 +232,40 @@ __xdata uint8_t *scan_header(__xdata uint8_t *p)
 		boundary[3] = '-';
 		boundary[i + 4] = 0;
 	}
+
+	reg_read_m(RTL837X_REG_SEC_COUNTER);
```
vDorst: Maybe convert this like a function like:

```c
void read_reg_timer(uint32_t * tmr)
{
    uint8_t * val = (uint8_t *)tmr;
	SFR_REG_ADDR_U16 = RTL837X_REG_SEC_COUNTER;
	SFR_EXEC_GO = SFR_EXEC_READ_REG;
	do {
	} while (SFR_EXEC_STATUS != 0);
    val++ = SFR_DATA_0;
    /* other values */
}
```

httpd/httpd.c
```diff
@@ -198,10 +232,40 @@ __xdata uint8_t *scan_header(__xdata uint8_t *p)
...
+				authenticated = 1;
+			else
+				print_string("Invalid session cookie!\n");
+		}
+	}
 	return p;
 }
 
 
+void gen_random_bytes(__xdata uint8_t *b, uint8_t bytes)
```
vDorst: maybe call it `gen_random_bytes_string()`, because we add 'null'-terminate.

httpd/httpd.c
```diff
@@ -302,13 +373,40 @@ void handle_post(void)
...
 		if (i)
 			cmd_available = 1;
-	} else if (is_word(request_path, "upload")) {
-		print_string("POST upload request\n");
+	} else if (is_word(request_path, "login")) {
+		print_string("POST login\n");
+		p += 8; // Read also over "pwd="
+		if (is_word_x(p, passwd)) {
+			print_string("Password accepted!\n");
+			reg_read_m(RTL837X_REG_SEC_COUNTER);
```
vDorst: `read_reg_timer()` can be use there too.

tools/httpd_sim.c
```diff
@@ -253,13 +293,30 @@ void send_bad_request(int socket) {
...
 char *scan_header(char *p)
 {
+	session = 0;
+	authenticated = false;
 	while (*p != '\r' || *(p + 1) != '\n' || *(p + 2) != '\r' || *(p + 3) != '\n') {
 		if (!*p++)
 			break;
-		if (*p == '\n' && is_word(p + 1, "Content-Type:"))
+		if (is_word(p, "\nContent-Type:"))
 			content_type = p + 15;
```
vDorst, line 317: maybe add `continue` so we don't wasted time on an extra cookie check we know is going to fail.

vDorst (commented):
Some more comments.
After this I think it is fine to merge.

httpd/httpd.c
```diff
@@ -175,16 +205,20 @@ __xdata uint8_t *skip_boundary(__xdata uint8_t *p)
...
+	session = 0;
+	authenticated = 0;
 
 	while (*p != '\r' || *(p + 1) != '\n' || *(p + 2) != '\r' || *(p + 3) != '\n') {
 		write_char(*p);
 		if (!*p++)
 			break;
 		if (is_word(p, "\nContent-Type:"))
 			content_type = p + 15;
+		if (is_word(p, "\nCookie:"))
```
vDorst: We want the `else` here too like as in did in `tools/httpd_sim.c`.

httpd/httpd.c
```diff
@@ -198,10 +232,38 @@ __xdata uint8_t *scan_header(__xdata uint8_t *p)
...
+				authenticated = 1;
+			else
+				print_string("Invalid session cookie!\n");
+		}
+	}
 	return p;
 }
 
 
+void gen_random_bytes_string(__xdata uint8_t *b, uint8_t bytes)
```
vDorst: On a second thougth, I think the interface not ideal and easy to misuse.
I think it should be like `memset()` or `strset()`, so we provide the real array/string size.

A option is too still treat `session_id` as a fix-length string, and set the zero-byte at startup.
That just use `memset()`-like random the fill the `size_of(session_id) - 1`.
So \0 everytime not set everytime.

httpd/httpd.c

httpd/httpd.c
logicog: What about this:
```
void gen_random_bytes(__xdata uint8_t *b, uint8_t bytes)
{
	__xdata uint8_t i = 0;
	while (bytes) {
		if (!i)
			get_random_32();
		b[--bytes] = i ? itohex(sfr_data[i]): itohex(sfr_data[i] >> 4);
		i = (i + 1) & 0x7;
	}
}

gen_random_bytes(session_id, 12);
session_id[12] = '\0';
```

logicog: Hmm, that should be
```
b[--bytes] = (i & 1)? itohex(sfr_data[i] >> 4): itohex(sfr_data[i]);
```
Or maybe even shorter:
```
b[--bytes] = itohex((i & 1)? sfr_data[i] >> 4: sfr_data[i]);
```

vDorst: Interface looks good.
But `i` and `bytes` are now incremented/decremented at the same rate.

I think to could look like this.
```c
void gen_random_bytes(__xdata uint8_t *b, uint8_t bytes)
{
	__xdata uint8_t i = 0;
	while (bytes) {
		if (!i)
			get_random_32();
		b[--bytes] = itohex(sfr_data[i]);
		if (!bytes) { break; }
		b[--bytes] = itohex(sfr_data[i] >> 4 | sfr_data[i] << 4);
		i = (i + 1) & 0x3;
	}
}
```
We don't care about the order of the bytes nibbles. So we can just print the lower nibble first.
`sfr_data[i] >> 4 | sfr_data[i] << 4` this should result in a single `swap`-instruction.

Can we demagic the number `12`. It is also used in other places in the code.

logicog: Yes, indeed, this is correct, my try was not.

PR #37, opened 2025-11-02
De-Magic I2C configuration for SFP-EEPROM reads

logicog:
De-magic register configuration around I2C reads to the SFP EEPROMS.


vDorst:
Thanks!

PR #39, opened 2025-11-04
Sfp and Link status work

logicog:
This fixes several issues with the existing SFP code and the link status code. It now provides full support for 2 SFP ports and uses also the link status register for ports 8 and 9 correctly.
This was tested on 5+1, 4+2 and 8+1 devices.
An example output now is e.g.:
```
> stat

5       SFP OK  Down    0x00000000      0x00000000      0x00000000      0x00000000
1       On      Down    0x00000000      0x00000000      0x00000000      0x00000000
6       SFP OK  10G     0x0000002d      0x00000000      0x00000000      0x00000000

> sfp

Slot 0 - Rate: 67  Encoding: 06
OEM             SFP-10G-SR      2   

Slot 1 - Rate: 0d  Encoding: 01
OEM             SFP-GE-SX-MM850     
```
For reasons of code complexity (use of additional variables), the ports are not in sequence on RTL8272 devices, only for RTL8273 this is the case.

### DIFF
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -415,6 +415,10 @@ void parse_port(void)
 	uint8_t p = cmd_buffer[cmd_words_b[1]] - '1';
 	p = machine.phys_to_log_port[p];
 	print_byte(p);
+	if (machine.is_sfp[p]) {
+		print_string(" is SFP no PHY information available.\n");
+		return;
+	}
 	if (cmd_words_b[2] > 0 && cmd_compare(2, "10m")) {
 		print_string(" 10M\n");
 		phy_set_speed(p, PHY_SPEED_10M);
--- a/rtl837x_phy.c
+++ b/rtl837x_phy.c
@@ -290,7 +290,6 @@ void phy_show(uint8_t port) __banked
 		print_string("\nForced speed: "); print_short(v); write_char('\n');
 		uint8_t s1 = ((v & 0x40) ? 0x2 : 0x0) | ((v & 0x2000) ? 0x1 : 0x0);
 		uint8_t s2 = (v >> 2) & 0xf;
-		print_string("s1, s2: "); print_byte(s1); write_char(' '); print_byte(s2); write_char('\n');
 		switch(s1) {
 		case 0:
 			print_string("10M\n");


vDorst:
@logicog can you rebase this PR, so I can check it again.

logicog:
On it. Need to push my work on IGMP aside, first.

vDorst:
Website and console don't match.
Console stat is correct.
I do have a SFP, 10Gbit link and RJ45, 1gbit, connected.

Screenshot_20251207_121644

`status.json`
```json
[{"portNum":5,"isSFP":1,"enabled":1,"link":4,"txG":"0x056","txB":"0x0","rxG":"0x033","rxB":"0x0"},{"portNum":1,"isSFP":0,"enabled":1,"link":0,"txG":"0x00","txB":"0x0","rxG":"0x00","rxB":"0x0"},{"portNum":2,"isSFP":0,"enabled":1,"link":2,"txG":"0x0e1","txB":"0x0","rxG":"0x0165","rxB":"0x0"},{"portNum":3,"isSFP":0,"enabled":1,"link":0,"txG":"0x00","txB":"0x0","rxG":"0x00","rxB":"0x0"},{"portNum":4,"isSFP":0,"enabled":1,"link":0,"txG":"0x00","txB":"0x0","rxG":"0x00","rxB":"0x0"},{"portNum":6,"isSFP":1,"enabled":0,"link":1,"txG":"0x00","txB":"0x0","rxG":"0x00","rxB":"0x0"}]
```

Console
```
stat

5       SFP OK  10G     0x00000035      0x00000000      0x0000005a      0x00000000
1       On      Down    0x00000000      0x00000000      0x00000000      0x00000000
2       On      1000M   0x000001ce      0x00000000      0x00000134      0x00000000
6       NO SFP  Down    0x00000000      0x00000000      0x00000000      0x00000000
```

logicog:
I'll add a fix for the website, too, then, now that we know that the new code in principle works. I'll try to do this by tomorrow.

logicog:
Not for me. The SFP-Module is in the _left_ slot of keeplink KP-9000-6XHML-X2:
```
> stat

5       SFP OK  10G     0x00000099      0x00000000      0x00000415      0x00000000
1       On      1000M   0x000001e8      0x00000000      0x00000476      0x00000000
2       On      1000M   0x0000035d      0x00000000      0x00000162      0x00000000
6       NO SFP  Down    0x00000000      0x00000000      0x00000000      0x00000000
```
Screenshot_20251209_064815

Note, that on the web, for you sfp_slot_0, which is the 0th I2C connection is correct, but for me, not. So at least in both cases logical port 5  is associated with I2C- connection 1, So for now, we only have a problem with geometry, not with association of I2C-port to SDS-lane or to the GPIOs.
So what about introducing say a MACHINE_KL_9000_6XHML_X2, and MACHINE_SWGT024_V2_0 which use a different geometry? Something like
```
In common:
#define MACHINE_KL_9000_6XHML_X2
// #define MACHINE_SWGT024_V2_0
#ifdef MACHINE_KL_9000_6XHML_X2
#define GEOMETRY_PHY_3_IS_PORT_5
#endif
In rtl837x_port.c:
#ifdef GEOMETRY_PHY_3_IS_PORT_5
__code uint8_t log_to_phys_port[9] = {
	0, 0, 0, 5, 1, 2, 3, 4, 6
};
#else
__code uint8_t log_to_phys_port[9] = {
	0, 0, 0, 6, 1, 2, 3, 4, 5
};
#endif
```
The Problem is that the web-interface also has such definitions in port.js:
```
const logToPhysPort = [0, 0, 0, 5, 1, 2, 3, 4, 6];
const physToLogPort = [	4, 5, 6, 7, 3, 8];
```
So this is a bigger mess. And I now know why I had a terrible feeling when introducing PHY-port mappings to the web-application.
We could run the javascript through the C-Preprocessor when compiling the image...
Or we could give a geometry hint in the `status.json`...

logicog:
> I can provide the link-changes prints, so we can figureout which port is connected which sds and also document it on the wiki.
At this point I believe this is not the problem, unless one of both of the 2 SFP-ports does not work for you.

I like the structures. This is allows compile-time verification. It may make the code a bit larger, but since we do not have an issue with the flash-size, this is irrelevant.

I will provide a proposal using structures, later today (finally got IGMP working, need to wrap that up while it is fresh in my mind). Note that we will still need some defines, because we cannot auto-detect geometry, this has to be input to the compiler, the structures in the kernel you mention are based on auto-detection of the SoC.  Device-tree btw would probably be the ideal solution, as it creates structures in the build image, which are compile-time dependent, but this comes with an enormous amount of additionally required infrastructure.

Because we go for the structure, I will also add information on geometry to the .json sent to the client which ends up in a struct there.

vDorst:
The "auto-detect" can just be a config-setting.
When the device information is to much we can put it in a bank, and only copy the used settings in to XDATA memory or program it in the "CODE"-bank.
And frequently used settings we can add flags in SRAM, so that the code-size to fetch a setting don't explode or consuming to much cycles.

logicog:
I pushed an initial implementation. It is more for feedback and not very well tested, so far only on the 4+2 device. But it is a bit of work and easy to get wrong, so better get feedback early.
There are lots of things that still can be done, such as using SFP-pins, there is nothing yet about the I2C pins. and looking at #40 there may be the need to have SDS configuration in.

logicog:
Here they come. Sorry!

machine.h
```diff
@@ -0,0 +1,34 @@
...
+
+// #define DEFAULT_8C_1SFP
+
+// #define DEFAULT_5C_1SFP
+
+typedef struct machine {
+	char machine_name[30];
+	uint8_t isRTL8373;
+	uint8_t min_port;
+	uint8_t max_port;
```
vDorst, line 31: I was hoping that 
```
	uint8_t min_port : 4;
	uint8_t max_port : 4;
```
gives a smaller struct and better code gen.
But code-gen is already bad.
I always does load the full pointer over and over again.
Instead just to inc the pointer on the second fetch.

I was also excepting that the compiler could see that most variables are a const.
And that it make use of that trait.

rtl837x_igmp.c
```diff
@@ -25,7 +22,7 @@ void igmp_setup(void) __banked
 	// For now, forward all unkown MC pkts (2 bits per port. 00: flood via floodmask, 01: drop, 10: trap, 11: to rport)
 	REG_SET(RTL837X_MC_LOOKUPMISS_ACTIONS, 0x00000000); //0x4f78
 	// Define ports where unknown MC addresses are flooded to:
-	if (isRTL8373) {
+	if (machine.isRTL8373) {
 		REG_SET(RTL837X_MC_FLOODMASK, PMASK_9);	// R5368-000001f8
 	} else {
 		REG_SET(RTL837X_MC_FLOODMASK, PMASK_6);
```
vDorst: replace with `REG_SET(RTL837X_MC_FLOODMASK, machine.isRTL8373 ? PMASK_9 : PMASK_6)`

machine.h
```diff
@@ -0,0 +1,34 @@
...
+	uint8_t n_sfp;
+	uint8_t log_to_phys_port[9];
+	uint8_t phys_to_log_port[9]; // Starts at 0 for port 1
+	uint8_t is_sfp[9];  // 0 for non-SFP ports 1 or 2 for the I2C port number
+	uint8_t sfp_0_pin;
+	uint8_t sfp_0_los;
+	uint8_t sfp_1_pin;
+	uint8_t sfp_1_los;
```
vDorst: I think we should put this also in a struct `sfp_port` and array.
So it can be expanded later with i2c-settings and other sfp features.

```c
struct sfp_port
  uint8_t pin_detect; // gpio number 0-63, 0xFF = don't have it?
  uint8_t pin_los; // gpio number 0-63, 0xFF = don't have it?
  uint8_t port_map; // replaces uint8_t is_sfp[9];
}

sfp_port sfp_port[2];
```

Maybe add a helper function to return a bool if a port is a SFP port.
I am thinking what information to put here when a single phy is attached to a sds.

httpd/page_impl.c
```diff
@@ -139,6 +134,20 @@ void reg_to_html(register uint16_t reg)
 	sfr_data_to_html();
 }
 
+
+void send_sfp_info(uint8_t sfp)
+{
+	// This loops over the Vendor-name, Vendor OUI, Vendor PN and Vendor rev ASCII fields
+	for (uint8_t i = 20; i < 60; i++) {
+		if (i >= 36 && i < 40) // Skip Non-ASCII codes
+			continue;
```
vDorst, line 143: Maybe replace the NON-ACSII with a `.` or `?`?

machine.h
```diff
@@ -0,0 +1,34 @@
...
+// #define DEFAULT_5C_1SFP
+
+typedef struct machine {
+	char machine_name[30];
+	uint8_t isRTL8373;
+	uint8_t min_port;
+	uint8_t max_port;
+	uint8_t n_sfp;
+	uint8_t log_to_phys_port[9];
+	uint8_t phys_to_log_port[9]; // Starts at 0 for port 1
```
vDorst, line 34: Maybe it is micro optimization that makes it worse.

But I am thinking of replace `log_to_phys_port` and `phys_to_log_port` with:
```c
port_mapping port_mapping[9];

struct port_mapping {
  uint8_t log_to_phy:4;
  uint8_t phy_to_log:4;
}
```
So that the upper-nibble is `log_to_phy` and lower-nibble `phy_to_log`.

But maybe it is terrible and makes the codegen worse?

rtl837x_port.c
```diff
@@ -266,7 +262,7 @@ uint8_t port_l2_forget(void) __banked
 	REG_SET(RTL837x_L2_TBL_FLUSH_CNF, 0x0);
 
 	// Flush L2 table for all ports by setting the ports and the flush-exec bit (bit 16)
-	if (isRTL8373) {
+	if (machine.isRTL8373) {
 		REG_SET(RTL837x_L2_TBL_FLUSH_CTRL, L2_TBL_FLUSH_EXEC | PMASK_9);
 	} else {
 		REG_SET(RTL837x_L2_TBL_FLUSH_CTRL, L2_TBL_FLUSH_EXEC | PMASK_6);
```
vDorst: Can also be replaced `REG_SET(RTL837X....., machine.isRTL8373 ? ... : ...)`

rtl837x_port.c
```diff
@@ -358,14 +354,14 @@ void port_l2_setup(void) __banked
...
 		uint16_t reg = RTL837X_L2_LRN_PORT_CONSTRAINT + (i << 2);
 		REG_SET(reg, 0x00001040);
 
 		// All ports may communicate with each other and CPU-Port
 		reg = RTL837X_PORT_ISOLATION_BASE + (i << 2);
-		if(isRTL8373) {
+		if(machine.isRTL8373) {
 			REG_SET(reg, PMASK_9 | PMASK_CPU);
 		} else {
 			REG_SET(reg, PMASK_6 | PMASK_CPU);
```
vDorst: Can also be replaced `REG_SET(RTL837X....., machine.isRTL8373 ? ... : ...)`

vDorst (commented):
Overall it looks good.

machine.h
logicog: Does the code change if one adds a `const` to `__code const`? My understanding is that it does not change anything because __code is already handled as constant.

The implementation puts the machine-struct into BANK 0, which means it is always cached in the 16k DRAM Code cache of the SoC. At least the speed should not be a problem. The usage of that DRAM is maybe an issue, but we always only have 1 machine struct in memory, which is about 65 bytes long and dominated by the 30-bytes for the machine name. Stuffing more bits into the bytes will not significantly reduce memory size, but I am worried about the code being produced and possible stack usage. To save bytes it would be easier to simply limit the name to say 20 bytes.

BANK-0 is a necessity, because const variables in __code can only be accessed in the same bank or BANK-0  by the code, and these are global definitions.

sdcc needs improvements, but I already looked at the code and it is about as readable as the one from gcc...

rtl837x_igmp.c
logicog: Will fix.

machine.h
logicog: Yes, to the struct. I need to look into replacing is_sfp[9] with port_map, this can break code, but we anyway used IS_SFP before. The problem is that one would now need to look into 2 port_maps for the 2 ports instead of one is_sfp or the previous implementation, which just did heuristics based on the port number. Plus, the looking into the 2nd port map would be conditional. This will become a very ugly `#define` or a relatively heavy helper function.

httpd/page_impl.c
logicog: Yes.

machine.h
logicog: Need to investigate...

rtl837x_port.c
logicog: Yes.

rtl837x_port.c
logicog: Yes.

machine.h
vDorst: I did not test it.
But I am thinking maybe we have to move the data to XDATA.
Downside of code-data is that load a value from code-data is that the instruction always needs `a`-reg, `MOVC A,@A+DPTR`. XDATA instruct don't have this downsize. `MOVX A,@DPTR`
This also can give us the advantage to load machine-profile dynamically.
Machine-name can still point into code-data, bank-switching to get the name is also not a problem. This name is not fetch a lot. Only when loading the page at the first time and maybe in the console at boot.

machine.h
vDorst: > Plus, the looking into the 2nd port map would be conditional. This will become a very ugly #define or a relatively heavy helper function.

I leave it up to you what you think is best.

machine.h
logicog: The code does not change when const is added, e.g.
```
machine.sfp_port[0].pin_detect
still becomes
        mov     dptr,#(_machine + 0x003d)
        clr     a
        movc    a,@a+dptr
instead of just:
        mov a, #1
```
I know gcc does much better.

vDorst: When data is moved to XDATA it will remove the `clr a` instruction.
Doesn't save much.

rtl837x_port.c
```diff
@@ -184,7 +164,7 @@ void vlan_setup(void) __banked
 	vlan_names[0] = 0;
 
 	// Initialize VLAN table for VLAN 1, by disabling that entry
-	if (isRTL8373) {
+	if (machine.isRTL8373) {
 		REG_SET(RTL837x_TBL_DATA_IN_A, 0x0007ffff);
 	} else {
 		REG_SET(RTL837x_TBL_DATA_IN_A, 0x0007e3f8);
```
vDorst: can be replaced by `REG_SET(RTL837x_TBL_DATA_IN_A, machine.isRTL8373 ? 0x0007ffff : 0x0007e3f8);`

rtl837x_port.c
```diff
@@ -231,7 +211,7 @@ void vlan_setup(void) __banked
 	REG_SET(RTL837X_VLAN_L2_LRN_DIS_1, 0);
 
 	// Enable VLAN 1: Ports 0-9, i.e. including the CPU port are untagged members
-	if (isRTL8373) {
+	if (machine.isRTL8373) {
 		REG_SET(RTL837x_TBL_DATA_IN_A, 0x0207ffff);	// 02: Entry valid, 7ffff: membership
 	} else {
 		REG_SET(RTL837x_TBL_DATA_IN_A, 0x0207e3f8);
```
vDorst: Same here `REG_SET(RTL837x_TBL_DATA_IN_A, machine.isRTL8373 ? 0x0207ffff : 0x0207e3f8);`

rtl837x_port.c
```diff
@@ -266,11 +246,7 @@ uint8_t port_l2_forget(void) __banked
 	REG_SET(RTL837x_L2_TBL_FLUSH_CNF, 0x0);
 
 	// Flush L2 table for all ports by setting the ports and the flush-exec bit (bit 16)
-	if (isRTL8373) {
-		REG_SET(RTL837x_L2_TBL_FLUSH_CTRL, L2_TBL_FLUSH_EXEC | PMASK_9);
-	} else {
-		REG_SET(RTL837x_L2_TBL_FLUSH_CTRL, L2_TBL_FLUSH_EXEC | PMASK_6);
-	}
+	REG_SET(RTL837x_L2_TBL_FLUSH_CTRL, machine.isRTL8373 ? L2_TBL_FLUSH_EXEC | PMASK_9 : L2_TBL_FLUSH_EXEC | PMASK_6);
```
vDorst: I was expecting `REG_SET(RTL837x_L2_TBL_FLUSH_CTRL, L2_TBL_FLUSH_EXEC | machine.isRTL8373 ?  PMASK_9 : PMASK_6);`
Not sure if that works in C.

rtl837x_port.c
vDorst: Can also be replaced `REG_SET(RTL837X....., machine.isRTL8373 ? ... : ...)`

rtlplayground.c
```diff
@@ -950,36 +947,36 @@ void handle_sfp(void)
...
-		sfp_print_info(1);
-		sds_config(0, sfp_rate_to_sds_config(rate));
-	}
-	if ((!(sfp_pins_last & 0x10)) && (sfr_data[1] & 0x04)) {
-		sfp_pins_last |= 0x10;
-		print_string("\n<MODULE 2 REMOVED>\n");
-	}
+	if (machine.n_sfp == 2) {
+		reg_read_m(RTL837X_REG_GPIO_32_63_INPUT);
+		if ((sfp_pins_last & 0x10) && (!(sfr_data[1] & 0x04))) {
```
vDorst: I would like to see a helper function to read the gpio input status.
Like 
```c
bool gpio_read_input_state(uint_t pin_number) {

}
```
But I would like to avoid reading the gpio states multiple times from the SOC.
So I would like to propose gpio-input-cache.
So read the both input registers from the SOC once at the begin of the loop and put it in array.
Use the helper function above, to fetch the input state from the cached array.

I think that makes it more readable, I think also faster and better code-gen.

rtlplayground.c
```diff
@@ -1584,21 +1577,17 @@ void rtl8373_init(void)
...
 	// r6330:00015555 R6330-00005555 r6330:00005555 R6330-00005555
 	REG_SET(0x6330, 0x00005555);
-	if (nSFPPorts == 2) {
+	if (machine.n_sfp == 2) {
 		print_string("Configuring 2nd SFP+ port\n");
 		REG_SET(0x6330, 0x00005515);
 	}
 
 	// r6334:00000000 R6334-000001f8  RTL8373: r6334:00000000 R6334-000000ff
 	reg_read_m(0x6334);		// Also in sdsMode_set
```
vDorst: Can we demagic these values?

httpd/page_impl.c
```diff
@@ -358,14 +347,11 @@ void send_status(void)
...
-			itoa_html(log_to_phys_port[i]);
-		else
-			itoa_html(i + 1);
+		itoa_html(machine.log_to_phys_port[i]);
 
-		if (IS_SFP(i)) {
+		if (machine.is_sfp[i]) {
 			slen += strtox(outbuf + slen, ",\"isSFP\":1,\"enabled\":");
 			if (i != 3) {
 				reg_read_m(RTL837X_REG_GPIO_00_31_INPUT);
```
vDorst: helper function `gpio_read_input_state()` can also be used here.
See more information futher in the review.

rtl837x_port.c
logicog: Will fix.

rtl837x_port.c
logicog: Will fix.

rtl837x_port.c
logicog: It does, but I added extra `()` for good measure.

rtlplayground.c
logicog: Bufferring inputs is inviting accidents to happen as the data is by necessity volatile. Usually one only buffers outputs. I implemented:
```
bool gpio_pin_test(uint8_t pin)
{
	reg_read_m(RTL837X_REG_GPIO_00_31_INPUT + (pin > 31 ? 4 : 0));
	return sfr_data[(pin >> 3) & 3] & (1 << (7 - (pin & 7)));
}
```
Which greatly improves readability at the cost of 2 bytes on the stack, but here we are not very deep in the nested functions, so not really a big problem. In `page_impl.c:send_status()` we can go for the cached state in `sfp_pins_last`.

httpd/page_impl.c
logicog: I replaced the entire GPIO testing section with:
```
if (machine.is_sfp[i]) {
	slen += strtox(outbuf + slen, ",\"isSFP\":1,\"enabled\":");
	bool_to_html(!(sfp_pins_last & (0x1 << ((machine.is_sfp[i] - 1) << 2))));
}
```
I hope this does not disappoint, given that I noticed a certain fondness for the Elvis-operator ?: recently, and there would have been a bit of an opportunity, here, too.

httpd/page_impl.c
logicog: Ähh, on second thought, better no. These are often at the end and then it really does not look nice on the web-page.

logicog:
Fixed the above, but still need to fix the javascript-code and transfer of configuration to the web-browser.
I was thinking maybe to add proper LED configuration, too in the machine state. This could help PR #44  These configuration values are not just random numbers, behind them is a model with LED-groups, port numbers and leds-per-port that could be de-magiced.
I also need to go through the magic in the _init() routines you pointed out.

machine.c
```diff
@@ -0,0 +1,58 @@
...
+	.sfp_port[0].sds = 1,
+	.sfp_port[0].i2c = 1,
+	.sfp_port[1].pin_detect = 1,
+	.sfp_port[1].pin_los = 58,
+	.sfp_port[1].sds = 0,
+	.sfp_port[1].i2c = 0,
+	.reset_pin = 49,
+};
+
+#elif MACHINE_SWGT024_V2_0
```
vDorst: this not working
I had to change it to `#elif defined(MACHINE_SWGT024_V2_0)`.

httpd/page_impl.c
```diff
@@ -167,7 +176,16 @@ void send_basic_info(void)
 	byte_to_html(uip_ethaddr.addr[5]);
 	slen += strtox(outbuf + slen, "\",\"sw_ver\":\"");
 	slen += strtox(outbuf + slen, VERSION_SW);
-	slen += strtox(outbuf + slen, "\",\"hw_ver\":\"SWGT024-V2.0\"}");
+	slen += strtox(outbuf + slen, "\",\"hw_ver\":\"SWGT024-V2.0\"");
```
vDorst: here we can output `machine.name`.

httpd/page_impl.c
```diff
@@ -167,7 +176,16 @@ void send_basic_info(void)
...
+	slen += strtox(outbuf + slen, ",\"sfp_slot_0\":\"");
+	send_sfp_info(0);
+	char_to_html('"');
+	if (machine.n_sfp == 2) {
+		slen += strtox(outbuf + slen, ",\"sfp_slot_1\":\"");
+		send_sfp_info(1);
+		char_to_html('"');
+	}
```
vDorst, line 188: can be replaced by.
```c
	if (machine.n_sfp == 2) {
		slen += strtox(outbuf + slen, "\",\"sfp_slot_1\":\"");
		send_sfp_info(1);
	}
    char_to_html('"');
```

rtlplayground.c
```diff
@@ -906,64 +901,60 @@ static inline uint8_t sfp_rate_to_sds_config(register uint8_t rate)
...
-		print_string("\n");
-		sfp_print_info(0);
-		sds_config(1, sfp_rate_to_sds_config(rate));
-	}
-	if ((!(sfp_pins_last & 0x1)) && (sfr_data[0] & 0x40)) {
-		sfp_pins_last |= 0x01;
-		print_string("\n<MODULE REMOVED>\n");
-	}
+	reg_read_m(RTL837X_REG_GPIO_00_31_INPUT + (pin > 31 ? 4 : 0));
+	return sfr_data[(pin >> 3) & 3] & (1 << (7 - (pin & 7)));
```
vDorst: I think this in incorrect.
`return sfr_data[3-((pin >> 3) & 3)] & (1 << (pin & 7));`
I am still testing it, this seems to work.
I think the error is `reg_read_m()` is already converting LE to BE.
And the bits don't have to be swapped. Only the byte order is different.

vDorst:
Hmm it seems that my Dell SFP don't drive the LOS-pin.
I don't see any change with `gpio` command.

rtlplayground.c
logicog: I tested this extensively on the keeplink 4+2 and 5+1 devices. The GPIOs for the sfp slots all work, module detection and LOS detection. Maybe for your device the GPIOs are different?

httpd/page_impl.c
logicog: Fixed. I just added the commit.

vDorst:
I updated to latest commit and added the fix GPIO settings,readout and document to it.
See https://github.com/vDorst/RTLPlayground/commits/sfp_work5/.
I had to swap the sds ports other I did not get a link on right sfp port.

Both modules do detect properly and get a link.
Also 1G-RJ45-SFP module works a 1Gbit.

Screenshot_20251214_113146

> There were a couple of issues in....

Great that you found it!

logicog:
I think I got the latest changes from you into the current commits.

logicog:
Can we get the link speed in? This was already in the code, it was just broken, because of incorrect translation of physical port number to logical port number. OK, I added link partner display. But leaving it without the last commit would keep the code broken. I will start a new PR for the Web display and SFP properties, ok?

vDorst:
Ok, that is fine.

vDorst:
Correct that the SFP module don't show this information yet?
```
> port 6 show

PORT 08
Forced speed: 0x0000
s1, s2: 00 00
10M
Duplex: 0x0000 enabled: no

Link Partner advertises:

> stat

6       SFP OK  10G     0x00000019      0x00000000      0x00000042      0x00000000
1       On      100M    0x00000114      0x00000000      0x000000a7      0x00000000
5       NO SFP  Down    0x0000000e      0x00000000      0x0000000c      0x00000000
```

vDorst:
This means I can't force the SFP port to let say 1gbit or 2.5bit if I want/need to?

logicog:
https://superuser.com/questions/1731301/why-does-auto-port-speed-negotiation-not-work-on-sfp-fiber-uplinks
However, you can force an SFP port sometimes to use a different speed. This works by changing the SerDes frequency. The module needs to support this, even if the frequency is lower than its maximum frequency because of things like PLLs that are involved.

vDorst:
Done, great job!

logicog:
Thanks the great collaboration on this!

Issue #46, opened 2025-12-09
Update documentation


sergewar:
At first I would like to say "Thank you!" for this great project.

I tried to compile project and faced that error

```
make
mkdir -p output/
make -C tools
make[1]: Entering directory '/projects/RTLPlayground/tools'
mkdir -p output/
gcc httpd_sim.c -Wall -o output/httpd_sim -I/usr/include/json-c -ljson-c
httpd_sim.c:10:10: fatal error: json.h: No such file or directory
   10 | #include <json.h>
      |          ^~~~~~~~
compilation terminated.
make[1]: *** [Makefile:23: output/httpd_sim] Error 1
make[1]: Leaving directory '/projects/RTLPlayground/tools'
make: *** [Makefile:40: tools] Error 2
```

after some investigation found that also need to install `libjson-c-dev` library

```
apt install libjson-c-dev
```

also found that there is no package `python`
```
apt install python
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Package python is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  python-is-python3 2to3

E: Package 'python' has no installation candidate
```

logicog:
Thank you for your kind words and thank you for the contribution!

PR #47, opened 2025-12-09
Update requisites for compiling

sergewar:
https://github.com/logicog/RTLPlayground/issues/46


logicog:
I tested it on Debian 12 and 13 and it works!

PR #49, opened 2025-12-11
Add basic IGMP support

logicog:
This adds an implementation for trapping IGMP packets to the CPU which will identify IGMPv1/2/3 packets, but handle only v3. The implementation then inserts/updates/deletes L3 MC entries in the L3 lookup table. The entries consist of an Ipv4 Destination IP (the IPv4 MC address), a Source IP (0.0.0.0) and a Portmask. Note that this implementation is not VLAN aware, as there is no hardware support in the device.
    
An alternative strategy is to control switching of the L2-MC packets  in which the IPv4-MC packets are transported (dst-MaC is  01:00:5e:xx:yy:zz, with xx:yy:zz corresponding to bits in the Ipv4-MC address). This will allow to use VLAN-aware packet switching. While  code support is there for table insert/update/deletes, some further L2 configuration is missing.
    
There is no support for IPv6 MC, yet.

### DIFF
--- a/rtl837x_phy.c
+++ b/rtl837x_phy.c
@@ -282,6 +282,41 @@ void phy_show(uint8_t port) __banked
 {
 	uint16_t v;
 
+	// The actual PHY speed is in a Realtek propriatary register
+	print_string("\nLink speed: ");
+	phy_read(port, PHY_MMD_CTRL, 0xA434);
+	v = SFR_DATA_U16;
+	v = ((v & 0x0600) >> 7) | ((v & 0x0030) >> 4);
+	switch(v) {
+	case 0:
+		print_string("10M");
+		break;
+	case 1:
+		print_string("100M");
+		break;
+	case 2:
+		print_string("1000M");
+		break;
+	case 3:
+		print_string("500M");
+		break;
+	case 4:
+		print_string("10G");
+		break;
+	case 5:
+		print_string("2500M");
+		break;
+	case 6:
+		print_string("5G");
+		break;
+	default:
+		print_string("10M");
+	}
+	if (v & 0x8)
+		print_string(" full duplex");
+	else
+		print_string(" half duplex");
+
 	phy_read(port,  PHY_MMD_AN, 0x00);
 	v = SFR_DATA_U16;
 	if (!(v & 0x1000)) { // AN disabled, we are in forced mode


doc/igmp.md
```diff
@@ -0,0 +1,130 @@
+# IGMP (Internet Group Management Protocol) and MLD (Multicast Listener Discovery)
+IGMP (for IPv4) and MLD (for IPv6) are protocols that control the distribution
+of Layer-3 MC packets on the LAN, which otherwise would be flooded across the
```
vDorst: Tell that MC is MultiCast.

doc/igmp.md
```diff
@@ -0,0 +1,130 @@
...
+
+The RTL8372/3 SoC supports managing IPv4-MC using either Destination-IP (the IPv4
+multicast group address)/Source-IP (typically 0.0.0.0) matching or via controlling
+the switching of the underlying L2-MC packets (i.e. packets in 01:00:5e:xx:yy:zz, where
+xx:yy:zz are the LSBs of the IPv4-MC address). The DIP/SIP-based switching is
+not VLAN-aware, meaning a stream will be available in all VLANs if subscribed to.
+This is not a problem in a typical home network, however. The L2-based method
+is VLAN aware, but currently not supported in the software.
+
+There is presently no support for IPv6/MLD-based MC.
```
vDorst: Assume that only software is the limiting factor.
`no support` to `Although there is hardware support for IPv6/MLD-based MC, the currently software has no support for it.`

doc/igmp.md
```diff
@@ -0,0 +1,130 @@
...
+incoming IGMP packets to the CPU. `RTL837X_IGMP_TRAP_CFG` then is used to configure
+priority and CPU-Port of trapped IGMP/MLD packets. 
+
+Configuration of the IP-MC-forwarding to the listening ports is done by managing the
+forwarding tables of the switch, see [L2 learning](l2.md).
+
+
+## IGMP API
+The code currently provides the following functions:
+```
```
vDorst, line 68: You can tell Markdown that this is C.
So it is highlighted. 
~~~suggestion
```c
~~~

Before
```
void igmp_setup(void) __banked;
```

After
```c
void igmp_setup(void) __banked;
```

You can ignore this if you don't like this or it gets noisy.

doc/igmp.md
```diff
@@ -0,0 +1,130 @@
...
+
+Connect the switch to an MC-aware router (e.g. to your home network). Connect the 2 Linux/Windows
+devices to the switch. The connection to the router makes sure that Linux/Windows will send
+out IGMP messages on the ports connected to the switch, which they will only do if they are aware
+that there is a MC-aware router in the network. Make sure the 2 GUI devices are in the home network
+(e.g. via DHCP).
+
+Start streaming on one of the Linux/Windows machines:
+```
+$ vlc your_video.mp4 --sout="#std{access=udp, mux=ts, dst=239.255.0.1:8090}"
```
vDorst, line 114: Append ` --miface=<ifname>` tells vlc which interface to use.

rtl837x_igmp.c
```diff
@@ -6,52 +6,378 @@
...
 	// Define ports where unknown MC addresses are flooded to:
 	if (isRTL8373) {
-		REG_SET(RTL837X_MC_FLOODMASK, PMASK_9);	// R5368-000001f8
+		REG_SET(RTL837X_IPV4_UNKN_MC_FLD_PMSK, PMASK_9);
+//		REG_SET(RTL837X_IPV6_UNKN_MC_FLD_PMSK, PMASK_9);
 	} else {
-		REG_SET(RTL837X_MC_FLOODMASK, PMASK_6);
+		REG_SET(RTL837X_IPV4_UNKN_MC_FLD_PMSK, PMASK_6);
+//		REG_SET(RTL837X_IPV6_UNKN_MC_FLD_PMSK, PMASK_6);
 	}
```
vDorst: Can be replaced with `REG_SET(RTL837X_IPV4_UNKN_MC_FLD_PMSK, machine.isRTL8373 ? PMASK_9 : PMASK_6)`;

rtl837x_igmp.c
```diff
@@ -6,52 +6,378 @@
...
-	// Drop unknown MC messages
-	REG_SET(RTL837X_MC_LOOKUPMISS_ACTIONS, 0x00015540); //0x4f78
+	// Drop unknown IP-MC packets
+	if (isRTL8373) {
+		REG_SET(RTL837X_IPV4_PORT_MC_LM_ACT, LOOKUP_MISS_DROP_9);
+//		REG_SET(RTL837X_IPV6_PORT_MC_LM_ACT, LOOKUP_MISS_DROP_9);
+	} else {
+		REG_SET(RTL837X_IPV4_PORT_MC_LM_ACT, LOOKUP_MISS_DROP_6);
+//		REG_SET(RTL837X_IPV6_PORT_MC_LM_ACT, LOOKUP_MISS_DROP_6);
+	}
```
vDorst: Can be replaced with `REG_SET(RTL837X_IPV4_PORT_MC_LM_ACT, machine.isRTL8373 ? LOOKUP_MISS_DROP_9 : LOOKUP_MISS_DROP_6)`;

rtl837x_igmp.c
```diff
@@ -6,52 +6,378 @@
...
+
+/*	// Allow all physical ports to be dynamic router ports
+	reg_read_m(RTL837X_IGMP_ROUTER_PORT);
+	if (isRTL8373) {
+		REG_WRITE(RTL837X_IGMP_ROUTER_PORT, PMASK_9 >> 8, PMASK_9 & 0xff, sfr_data[1], sfr_data[0]);
+	} else {
+		REG_WRITE(RTL837X_IGMP_ROUTER_PORT, PMASK_6 >> 8, PMASK_6 & 0xff, sfr_data[1], sfr_data[0]);
+	}
+*/
```
vDorst, line 134: Still need? Maybe add comment or remove.

doc/igmp.md
logicog: yes.

doc/igmp.md
logicog: yes.

doc/igmp.md
logicog: No, this is cool! I did not know this existed.

doc/igmp.md
logicog: Ah. This is interesting. For me this was only necessary if the switch was not connected to the home network. I believe there are router announcement messages coming downstream on that interface which the Linux kernel will interpret and make vlc choose the right interfaces. I'll add a comment.

rtl837x_igmp.c
logicog: Yes.

rtl837x_igmp.c
logicog: yes.

rtl837x_igmp.c
logicog: I would like to leave this in for now. The router port configuration is not yet implemented. We will need a CLI command for this and also support on the web interface. It allows to circumvent IGMP for some ports so that MC messages are flooded even if the port is not in a MC group. This would be great in case there is a problem with equipment that provides no proper IGMP messages. In the given case IGMPv2 or even IGMPv1 which some old IPTVs sometimes still use. It would be necessary to test that...

logicog:
Rebased to latest main.

rtl837x_igmp.c
```diff
@@ -15,37 +17,361 @@
...
+	REG_WRITE(RTL837X_IGMP_ROUTER_PORT, sfr_data[0], sfr_data[1], pmask >> 8, pmask & 0xff);
+}
+
+
+/*
+ * IGMP show the current entries and state
+ */
+void igmp_show(void) __banked
+{
+	print_string("igmp_show called");
```
vDorst: Missing `\n` at the end
```
> igmp show
igmp_show called3:0x00ff7c15
4:0x00ff7c15
5:0x00ff7c15
6:0x00ff7c15
7:0x00ff7c15
8:0x00ff7c15
```

rtl837x_igmp.c
```diff
@@ -15,37 +17,361 @@
...
 void igmp_setup(void) __banked
 {
 	uint8_t i;
-
-	// For now, forward all unkown MC pkts (2 bits per port. 00: flood via floodmask, 01: drop, 10: trap, 11: to rport)
-	REG_SET(RTL837X_MC_LOOKUPMISS_ACTIONS, 0x00000000); //0x4f78
+	print_string("igmp_setup called\n");
+	// For now, forward all unkown IP-MC pkts (2 bits per port. 00: flood via floodmask, 01: drop, 10: trap, 11: to rport)
+	REG_SET(RTL837X_IPV4_PORT_MC_LM_ACT, LOOKUP_MISS_FLOOD);
+	REG_SET(RTL837X_IPV6_PORT_MC_LM_ACT, LOOKUP_MISS_FLOOD);
```
vDorst, line 90: Many places the `IPV6` settings are commented-out.
What about this one?

Also line 94.

doc/igmp.md
vDorst: Highlight feature from github, see https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks#syntax-highlighting

doc/igmp.md
vDorst: Was useful for me because vlc used the wrong interface.

rtl837x_igmp.c
vDorst: ok

rtl837x_igmp.c
logicog: This should actually also show the entries in the switch, not just the per-port configuration of IGMP. For this to work, searching for all L3 forwarding entries in the table space needs to be implemented. I have not figured out how to do that. Alternatively, and like the OEM firmware, the software itself tracks what is being forwarded. However, in theory this should not be necessary, and the OEM firmware only has to do that because using L2 MC entries to handle L3 MC looses information about the original MC groups, see e.g. https://networklessons.com/multicast/multicast-ip-address-to-mac-address-mapping
So this will need to come later and in connection with displaying that in the web-interface. For now the newline has been fixed.

rtl837x_igmp.c
logicog: I have not been able to test IPv6 streaming in vlc. I could not get it to work at all on the vlc client side. Even though the client saw the packets (i.e. wireshark saw them arrive at the network interface), vlc there did not pick them up.

For now, everything is in a supposedly safe setting for IPv6, i.e. everything is being flooded like for a dumb switch. IPv6 multicast management also uses MLD instead of IGMP and support for the protocol would need to be implemented. Plus, the switch does not support L3 IPv6 multicast switching in HW, so a translation layer IPv6-MC to L2-MC similarly to what the OEM firmware does for IPv4-MC would need to be implemented. (actually the switch HW does not support IPv6 at all in HW anywhere, but admittedly there are anyway only very few L3 features at all, such as L3 unicast/mc P2P-routing in HW). Note that IPv6-MC is not supported for the OEM firmware, either. I do not know what the practical implications are, as the most obvious application is IPTV, but the streaming box usually sits within the home network, where there are enough IPv4 addresses.

vDorst: Ok, then we leave it as is.

rtl837x_igmp.c
vDorst: Thanks for the extra info.

logicog:
Thanks a lot for testing and reviewing this!

PR #50, opened 2025-12-15
Sfp readings

logicog:
Tooltips

This pull request adds support for displaying of SFP dynamic information such as the temperature as tooltips for SFP ports. For Ethernet ports the speed is displayed, but it would also be possible to e.g. add link state information such as advertised speeds.

A simulation of SFP data is added for http_sim.

Because SFF-8472 requires bulk reading of the 2-byte dynamic data such as the module temperature, bulk reading of up to 16 bytes of SFP-module data via I2C bulk transfers is added. This could also be used to bulk transfer e.g. the module name.

Display of the raw SFP module data is added to the CLI. However, no pretty printing of the values e.g. in mW is done because of the requirement for custom tailored conversion routines without the availability of floating-point numbers.

### DIFF
--- a/rtl837x_phy.c
+++ b/rtl837x_phy.c
@@ -348,7 +348,6 @@ void phy_show(uint8_t port) __banked
 				break;
 			default:
 				print_string("Unknown\n");
-				break;
 			}
 			break;
 		default:


httpd/page_impl.c
```diff
@@ -148,6 +148,46 @@ void send_sfp_info(uint8_t sfp)
...
+
+	if (machine.sfp_port[slot].i2c == 0) {
+		reg_read_m(RTL837X_REG_I2C_CTRL);
+		sfr_mask_data(1, 0xfc, SCL_PIN << 5 | SDA_PIN_0 << 2);
+		reg_write_m(RTL837X_REG_I2C_CTRL);
+	} else {
+		reg_read_m(RTL837X_REG_I2C_CTRL);
+		sfr_mask_data(1, 0xfc, SCL_PIN << 5 | SDA_PIN_1 << 2);
+		reg_write_m(RTL837X_REG_I2C_CTRL);
+	}
```
vDorst:
~~~suggestion
sfr_mask_data(1, 0xfc, machine.sfp_port[slot].i2c == 0 ? SCL_PIN << 5 | SDA_PIN_0 << 2 : SCL_PIN << 5 | SDA_PIN_1 << 2 );
reg_write_m(RTL837X_REG_I2C_CTRL);
~~~
This will save 21 bytes.

httpd/page_impl.c
```diff
@@ -148,6 +148,46 @@ void send_sfp_info(uint8_t sfp)
...
 
 
+void sfp_send_data(uint8_t slot, uint8_t reg, uint8_t len)
+{
+	if (reg & 0x80) {	// Configure SFP readings address (0x51) as I2C device address
+		REG_WRITE(RTL837X_REG_I2C_CTRL, 0x00, 0x1 << (I2C_MEM_ADDR_WIDTH-16) | len & 0xf,  0x51 >> 5, 0x51 << 3);
+	} else {
+		REG_WRITE(RTL837X_REG_I2C_CTRL, 0x00, 0x1 << (I2C_MEM_ADDR_WIDTH-16) | len & 0xf,  0x50 >> 5, 0x50 << 3);
+	}
+	reg &= 0x7f;
```
vDorst: I was hoping that the compiler would optimize high bit check in `reg` and clearing that bit in one `JBC` jump-instruction.
I did try many combinations with hope that it optimize to that instruction but it doesn't. 

Like this
```c
if (reg & 0x80) {	// Configure SFP readings address (0x51) as I2C device address
	reg &= 0x7f;
	REG_WRITE(RTL837X_REG_I2C_CTRL, 0x00, 0x1 << (I2C_MEM_ADDR_WIDTH-16) | len & 0xf,  0x51 >> 5, 0x51 << 3);
	} else {
		REG_WRITE(RTL837X_REG_I2C_CTRL, 0x00, 0x1 << (I2C_MEM_ADDR_WIDTH-16) | len & 0xf,  0x50 >> 5, 0x50 << 3);
	}
```

or the more extreme, which gives better code but not the best with the `JBC`-instruction.
```c
uint8_t addr = 0x51 << 3;
if (reg & 0x80) {	// Configure SFP readings address (0x51) as I2C device address
	reg &= 0x7f;
} else {
	addr = 0x50;
}
REG_WRITE(RTL837X_REG_I2C_CTRL, 0x00, 0x1 << (I2C_MEM_ADDR_WIDTH-16) | len & 0xf,  0x50 >> 5, addr);

```

html/main.js
```diff
@@ -72,6 +83,15 @@ function update() {
 	    leds[0].style.fill = "black"; leds[1].style.fill = "black";
 	    psvg.style.opacity = 0.4
 	  }
+	  tt.innerHTML = "Link speed: " + linkS[p.link + 1] + "<br>";
+	  if (p.isSFP) {
+	    tt.innerHTML += "T = " + (Number(p.sfp_temp) >> 8) + "." + ((Number(p.sfp_temp) & 0xff)/256.0 * 100).toFixed(0) + "&deg;C<br>";
```
vDorst: instead of `&deg;C`, `&#8451;` can be used as "Degree Celsius".

html/main.js
```diff
@@ -8,13 +8,21 @@ var pIsSFP = new Int8Array(10);
 var numPorts = 0;
 const logToPhysPort = [0, 0, 0, 5, 1, 2, 3, 4, 6];
 const physToLogPort = [	4, 5, 6, 7, 3, 8];
-
+// <div class="tooltip"><div id="ports"><span class="tooltiptext">The text for the Tooltip.</span></div></div>
```
vDorst: Please remove if not needed.

httpd/httpd.c
```diff
@@ -593,7 +593,7 @@ void httpd_appcall(void)
 
 			slen = strtox(outbuf, "HTTP/1.1 200 OK\r\nContent-Type: ");
 			slen += strtox(outbuf + slen, mime_strings[f_data[entry].mime]);
-			slen += strtox(outbuf + slen, "\r\nCache-Control: max-age=60, must-revalidate\r\n\r\n");
+			slen += strtox(outbuf + slen, "; charset=UTF-8\r\nCache-Control: max-age=60, must-revalidate\r\n\r\n");
```
vDorst, line 596: Still needed, because you are using a HTML-tag for the degree symbol?

html/style.css
```diff
@@ -84,7 +84,7 @@ object {
 .row {display: flex;}
 .rcol {flex: 90%;}
 .lcol {flex: 10%;}
-span::before{ content:''; position: absolute; top: 40px; left:0; width: 100%; height: 2px; background: #aaf;}
+/* span::before{ content:''; position: absolute; top: 40px; left:0; width: 100%; height: 2px; background: #aaf;} */
```
vDorst: Please remove if not needed.

html/main.js
```diff
@@ -72,6 +83,15 @@ function update() {
 	    leds[0].style.fill = "black"; leds[1].style.fill = "black";
 	    psvg.style.opacity = 0.4
 	  }
+	  tt.innerHTML = "Link speed: " + linkS[p.link + 1] + "<br>";
+	  if (p.isSFP) {
+	    tt.innerHTML += "T = " + (Number(p.sfp_temp) >> 8) + "." + ((Number(p.sfp_temp) & 0xff)/256.0 * 100).toFixed(0) + "&deg;C<br>";
+	    tt.innerHTML += "Vcc = " + (Number(p.sfp_vcc) / 10000.0).toFixed(2) + "V<br>";
+	    tt.innerHTML += "TX-Bias = " + (Number(p.sfp_txbias) / 500.0).toFixed(1) + "mA<br>";
+	    tt.innerHTML += "TX-Power = " + (Number(p.sfp_txpower) / 10.0).toFixed(0) + "mW<br>";
+	    tt.innerHTML += "RX-Power = " + (Number(p.sfp_rxpower) / 10.0).toFixed(0) + "mW<br>";
```
vDorst: I seems recommended to put a space/half-space between the value and unit.
`10 mA` instead of `10mA`.

rtlplayground.c
```diff
@@ -745,6 +745,15 @@ void sds_config(uint8_t sds, uint8_t mode)
  */
 uint8_t sfp_read_reg(uint8_t slot, uint8_t reg)
 {
+	if (reg & 0x80) {
+		// Configure SFP readings address (0x51) as I2C device address
+		REG_SET(RTL837X_REG_I2C_CTRL, 0x1L << I2C_MEM_ADDR_WIDTH | 0x51 << I2C_DEV_ADDR);
+		reg &= 0x7f;
+	} else {
+		// Configure SFP EEPROM address (0x50) as I2C device address
+		REG_SET(RTL837X_REG_I2C_CTRL, 0x1L << I2C_MEM_ADDR_WIDTH | 0x50 << I2C_DEV_ADDR);
```
vDorst: Compiler warnings!
```
rtlplayground.c:750: warning 116: left shifting more than size of object changed to zero
-:0: warning 116: left shifting more than size of object changed to zero
rtlplayground.c:754: warning 116: left shifting more than size of object changed to zero
```

httpd/page_impl.c
logicog: Yes, will do.

httpd/page_impl.c
logicog: I'll do the first as it is more readable.

html/main.js
logicog: I'll try that, it might look nicer on the page.

html/main.js
logicog: Will do, this was my template for writing the HTML generating JS code...

httpd/httpd.c
logicog: Yes. Without this, the Degree sign does not work, instead funny characters appear. I thought that the charset directive configures the data being transmitted, but it appears more that it defines what characters can be displayed. UTF-8 is anyway the default nowadays, and we may need this to support translated pages with other European and Asian languages. Note that the charset directive is only for pages served out of the flash. json data is not affected.

html/style.css
logicog: Will do.

httpd/page_impl.c
vDorst: This was more about to show what I tried to get better machine-code output.
Not really a change suggestion

html/main.js
logicog: I will try  a narrow no-break space : &#8239;

rtlplayground.c
logicog: I will replace this with REG_WRITE. I do not know why the compiler complains, it seems to work.

html/main.js
vDorst: Change `T` to `Temp` so it is a bit more clear.
Maybe move the `brand` and `type` from the main page to here.
As an extra add the serialnumber too.

rtlplayground.c
vDorst: I think you have to cast it first to a large type then `uint8_t`.
`I2C_MEM_ADDR_WIDTH` = 20 so shifting it result in a atleast a 24-bit number.

logicog:
I implemented all of  your suggestions.
I also added tabular design for the tool-tips. However, I am not a web-designer and have to pour a lot of time into something that afterwards only looks mediocre. At least, now it is a table, and the rest can be done in style.css.

logicog:
Fixed all of the above and rebased.

PR #51, opened 2025-12-15
[WIP] Move sfr_data to pdata or sram.

vDorst:
I am working on SIPHASH were I need a large memory block of 32 bytes.
But I want to access it easily.
I know you can access lower part XDATA via a single register using the instruction `MOVX @Ri, A`. 
This is called Paged Memory??. Just use `__pdata` to define a variable.
Not sure if paging is controllable.

Because we are using `sfr_data` a lot, it make sense to make it faster to access.
I made a quick test to move `sfr_data` to `pdata`.
Because of some function I also needed to move `linkbits_last`.
See first commit.
As a result it reduced the code size with 350 bytes.
Most of these bytes, I assume, are used to load in the DPTR of calculate it.

Second commit is move `sfr_data` and `linkbits_last` to sram.
I think it make sense to use some extra sram to access sfr_data faster.
As a result it reduced the code size with 650 bytes from the original xdata code.

diff between the two `.mem`-files.
Screenshot_20251215_232627

So I think we can optimize more things by moving them into `__pdata`.

And/Or define a large scrats-memory block in pdata, which I may need for my SIPHASH.
Which can also be used by other code.

### DIFF
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -27,7 +27,7 @@ extern __xdata uint8_t stpEnabled;
 extern __code uint8_t log_to_phys_port[9];
 
 extern volatile __xdata uint32_t ticks;
-extern volatile __pdata uint8_t sfr_data[4];
+extern volatile __data uint8_t sfr_data[4];
 
 extern __code uint8_t * __code greeting;
 extern __code uint8_t * __code hex;
--- a/rtl837x_igmp.c
+++ b/rtl837x_igmp.c
@@ -23,7 +23,7 @@ extern __code struct machine machine;
 #pragma constseg BANK1
 
 extern __xdata uint8_t cpuPort;
-extern __pdata uint8_t sfr_data[4];
+extern __data uint8_t sfr_data[4];
 
 extern __xdata uint8_t uip_buf[UIP_CONF_BUFFER_SIZE + 2];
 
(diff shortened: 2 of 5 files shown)


vDorst:
An other test.
Only strange is that the last b is not overwritten.
This time we own that memory location so no-one else should write there.
```
PAGE 0 READ
aaaaaaaaaaaaaaaa
PAGE 1 READ
bbbbbbbbbbbbbbbb
PAGE 1 READ XDATA
xxxxxxxxxxxxxxxb
PAGE 0 READ
aaaaaaaaaaaaaaaa
PAGE 1 READ
xxxxxxxxxxxxxxxb
```
See commit https://github.com/vDorst/RTLPlayground/commit/b9531cab3c8f67ab13565abf7031d98d81802010

logicog:
Maybe the page and xdata are not completely aligned? Could you read a wider range of XMEM?
At least the PDATA reads/writes are consistent. So it should be usable.

BTW: There is also SRAM in the ASIC for TX and RX of packets, the datasheet says there are 8MBit, but I believe only 512kByte can be addressed (see `nic_rx_packet()` and `nic_tx_packet()`. Not all of it is used. The ASIC can do DMA to and from XDATA to write and read data in XMEM, see the send and receive routines in rtlplayground. Source and destination addresses are put into SFRs and then a transfer is executed. This could be a very fast way of doing memcpy.

PR #53, opened 2025-12-16
Add github actions

sergewar:


.github/workflows/build.yml
```diff
@@ -0,0 +1,37 @@
...
+      - name: Step 6
+        run: objcopy --input-target=ihex -O binary rtlplayground.ihx rtlplayground.img
+      - name: Step 7
+        run: if [ -e rtlplayground.bin ]; then rm rtlplayground.bin; fi
+      - name: Step 8
+        run: |
+          echo "0000000: 00 40" 
+          xxd -r - rtlplayground.bin
+      - name: Step 9
+        run: cat rtlplayground.img >> rtlplayground.bin
```
vDorst: Why are all these `make` steps also in here?
My feeling is that `make` is enough.

sergewar: These all commands are from Compiling section of readme https://github.com/logicog/RTLPlayground?tab=readme-ov-file#compiling

sergewar: Removed all steps, leave only `make`, but anyway it doesn't compile
https://github.com/sergewar/RTLPlayground/actions/runs/20295309590/job/58287643305
```
sdcc -mmcs51 -Ihttpd -Iuip -o output/rtl837x_igmp.rel -c rtl837x_igmp.c
rtl837x_igmp.c:229: error 78: incompatible types
from type 'struct ipmc_table_entry xdata* xdata'
  to type 'unsigned-char xdata* fixed'
make: *** [Makefile:55: output/rtl837x_igmp.rel] Error 1
Error: Process completed with exit code 2.
```

vDorst:
That is an old compiler.
`Debian 13` has version `4.5.0`, which is also the latest release.
So can we just use `Debian 13`?

PR #54, opened 2025-12-18
Add a Port Speed configuration Web-Page

logicog:
port_conf

This adds a port speed configuration page which allows to configure the advertised link speeds and duplex modes advertised by the respective PHYs. Basically, it is a web-frontend to the port command. Support for simulating the page in the http_sim is also added.

In order to be user-friendly, not all combinations of advertised link speeds are supported. When auto is selected, all speeds and duplex modes are advertised. However, there is no way to just advertise e.g. 1000M plus 100M in just full duplex, merely e.g 100M in full or half duplex or any.

Additionally, this PR adds support for machines with 9 ports in the web-frontend and in the simulator.

### DIFF
--- a/html/style.css
+++ b/html/style.css
@@ -160,3 +160,5 @@ margin: 30px 0;
 .tt_table td {
   text-align: left;
 }
+select { text-align-last: right; font-family: monospace}
+option { direction: rtl; font-family: sans-serif}


html/ports.js
```diff
@@ -0,0 +1,90 @@
...
+      + '<option value="2g5"> 2.5GBit</option>'
+      + '<option value="1g"> 1000MBit</option>'
+      + '<option value="100m"> 100MBit</option>'
+      + '<option value="10m"> 10MBit</option>'
+      + '</select>';
+     const dSelect = '<select name="duplex_sel" id="duplex_sel">'
+      + '<option value="any"> Any</option>'
+      + '<option value="full"> Full</option>'
+      + '<option value="half"> Half</option>'
+      + '</select>';
```
vDorst, line 15: Can we combine these option into one like in the original firmware?

Screenshot_20251219_214803

So you can´t select a non-existing option.

rtl837x_phy.c
vDorst, line 1: Commit message doesn´t match the changes.

cmd_parser.c
```diff
@@ -457,6 +457,38 @@ void parse_port(void)
...
+	uint8_t p;
+
+	print_string("\nMTU ");
+	if (cmd_words_b[1] > 0 && cmd_compare(1, "show")) {
+		for (p = machine.min_port; p <= machine.max_port; p++) {
+			reg_read_m(RTL8373_REG_MAC_L2_PORT_MAX_LEN + ((uint16_t) p << 8));
+			print_sfr_data();
+			mtu = SFR_DATA_U16 & 0x3fff;
+			print_string("Port "); print_byte(machine.log_to_phys_port[p]);
+			print_string(" MTU: 0x"); print_short(mtu); write_char('\n');
```
vDorst:
```
> mtu show

MTU 0x0fffffffPort 06 MTU: 0x0x3fff
0x017c85f2Port 01 MTU: 0x0x05f2
0x0fffffffPort 02 MTU: 0x0x3fff
0x0fffffffPort 03 MTU: 0x0x3fff
0x0fffffffPort 04 MTU: 0x0x3fff
0x0fffffffPort 05 MTU: 0x0x3fff
```
Output is a bit noise.

Maybe change it to, also print the mtu in decimal number instead of hex.
```
MTU:
Port x: 1522 (0x017c85f2)
```

vDorst (changes requested):
Overall looks great.

Also noticed that the session-timeout is low ~20sec.
And that this timeout not always reset when interacting with the site.

logicog:
> Also noticed that the session-timeout is low ~20sec.
> And that this timeout not always reset when interacting with the site.

I noticed the same, the low timeout is actually rather annoying when working on the websites. I also understand now the reason: the seconds are much too short. That is strange, because they used to be OK at some point. I investigated, and it seems there are multiple mistakes that to some extend cancel each other out in the timer configuration for the system ticks. The timer/counter is linked to the Frequency of the Crystal and there to only half that frequency. So it is 12.5MHz. I still need to check whether the Clock dividers have an effect on this, though. With 12.5MHz it is not possible to have a 100Hz System Tick IRQ, the counter runs out too fast and would require a counter larger than 16bits. The problem is that if the ticks are too short, the housekeeping routines, i.e. the web-page serving, the link-state handling and the RX of packets like pings and so on, may take longer than the tick-interval. If this happens now and then, this should not be a problem, because the tick-counting is handled in an IRQ. But if this happens for a prolonged time, then the clock goes wild. I believe 200Hz not only fits into the counter, but should also be slow enough for the housekeeping. I need to test more, though.

And then there is a secondary issue with the ticks. They are just a bit too fast, even if mathematically correctly derived from 12.5MHz. They are about the same amount too fast as the UART. Now, this could be the dead cycles in the ISR for the ticks, but that should not cause the problem for the UART, because there the clock is generated in HW from the counter, entirely without an ISR. It looks more like the Crystal is a bit off its official frequency.

I hope I have something tomorrow to fix the fast timeout.

The interactions with the web-page such as fiddling with the pull-down menus do not interact with the server, they cannot prolong the session, because for obvious security reasons the session is handled server-side. What I need to check is whether the configuration commands actually restart the session timeout, and I believe this is not the case, currently the session is restarted when static content is served. Will investigate.

cmd_parser.c
logicog: Oops, will fix.

html/ports.js
logicog: The disadvantage of the solution in the original FW is that it is not possible to just set the speed to e.g. 10MBit without specifying the Duplex. If you set it to 10MBit Half and the other side wants Full, then it will not work. And that is not so theoretical because you can usually see the speed from the LEDs, but not the duplex. The non-existing cases such as 2500Mbit Half duplex are not actually configured. If you try that, you get 2500Mbit full, so at least it works (the SoCs do not support 2.5GBit Half duplex). What I could do is force full duplex on the page if you select 1000MBit or 2500MBit. What do you think?

vDorst: > The disadvantage of the solution in the original FW is that it is not possible to just set the speed to e.g. 10MBit without specifying the Duplex. If you set it to 10MBit Half and the other side wants Full, then it will not work. 

How offend is this going to be a issue.
I think if want to reduce the speed you are doing it for a reason.
But when also showing the `link partner` advertised settings, you can see what it supports.

Made a quick hacking in Typst to make a concept "layout"

Screenshot_20251220_095700

With this you see the link partner settings and select every bit you want.

Other ideas I found.
Mutliselect dropdown boxes:

Screenshot_20251220_091108

and

Screenshot_20251220_091209

> hat I could do is force full duplex on the page if you select 1000MBit or 2500MBit. What do you think?

I am fine with just forcing it to full-duplux.
I also don't see a reason these days why to force to use half-duplux.
If the other-side needs that, it should advertise like that.

rtl837x_phy.c
logicog: I will fix that.

html/ports.js
logicog: I will add a checkbox button to enable/disable the interface and combine the speed and duplex settings into one pull-down option which gets grayed out when the port is disabled. One apply button at the end. For now no FC, as this first needs backend code but could be added as another checkbox. The space should be there.

vDorst:
website sends the right command.

Screenshot_20251223_101823

CLI interpretation seems to go wrong.
```
> stat

6       SFP OK  10G     0x0000001b      0x00000000      0x00000004      0x00000000
1       On      Down    0x00000000      0x00000000      0x00000000      0x00000000
3       On      1000M   0x00000004      0x00000000      0x00000015      0x00000000

> port 3 off

PORT 05 OFF

> stat

6       SFP OK  10G     0x0000001d      0x00000000      0x00000004      0x00000000
1       On      Down    0x00000000      0x00000000      0x00000000      0x00000000
2       Off     Down    0x00000000      0x00000000      0x00000000      0x00000000
3       On      1000M   0x00000004      0x00000000      0x00000017      0x00000000

> port 2 off

PORT 06 OFF

> 
<new link: 2000004000, was 2002004000>
stat

6       SFP OK  10G     0x0000001d      0x00000000      0x00000004      0x00000000
1       On      Down    0x00000000      0x00000000      0x00000000      0x00000000
2       Off     Down    0x00000000      0x00000000      0x00000000      0x00000000
3       Off     Down    0x00000004      0x00000000      0x00000017      0x00000000
``

**Comment by logicog** [comment]
I think the problem with the switched ports is a mistake in the machine configuration:

```
#elif defined MACHINE_SWGT024_V2_0
__code const struct machine machine = {
	.machine_name = "SWGT024 V2.0",
	.isRTL8373 = 0,
	.min_port = 3,
	.max_port = 8,
	.n_sfp = 2,
	.log_to_phys_port = {0, 0, 0, 6, 1, 2, 3, 4, 5},
	.phys_to_log_port = {4, 6, 5, 7, 3, 8, 0, 0, 0},  <<<<<<<<<<<<<<<<<<<<<<<<<<<
```
It should be:
```
#ifdef MACHINE_KP_9000_6XHML_X2
[...]
	.phys_to_log_port = {4, 5, 6, 7, 3, 8, 0, 0, 0},
```
I fixed the MTU printout. Also there was an issue with switching on a port again, the command did not exist on the CLI, the web used auto, which worked.
I also fixed the dropdown for 1000MBit to show full duplex.
And finally, there was a bug in the full/half duplex detection for the CLI printout.

**Comment by logicog** [comment]
I fixed the mtu printout, sorry about that, I did not test it again after taking the `0x` out.
The `Auto`is because this is a speed selector, not an indicator. The column is called "Set Speed". To show the advertised speeds was the reason for that initially added column. Auto is just the default shown by the selector.

File: cmd_parser.c [hunk 3]
```diff
@@ -420,36 +421,47 @@ void parse_port(void)
 		print_string(" is SFP no PHY information available.\n");
 		return;
 	}
-	if (cmd_words_b[2] > 0 && cmd_compare(2, "10m")) {
-		print_string(" 10M\n");
-		phy_set_speed(p, PHY_SPEED_10M);
+	if (cmd_words_b[2] <= 0) {
+		print_string("mtu <port> [show|on|off|10m|100m|1g|2g5] [half|full]");
```
vDorst: word `mtu` is a copy/paste error?
Also two zero are printed in between
`PORT 00mtu <port> [show|on|off|10m|100m|1g|2g5] [half|full]`

vDorst:
Thanks for this great work!

PR #59, opened 2025-12-20
Fix system clock

logicog:
This fixes the system clock.
The system clock is based on timer 0, which is run in 16 bit counter mode. The timer/counter is driven by half the external crystal frequency. When the counter overflows an interrupt is called.  This increases the system ticks. These ticks are now configured to run at 200 Hz, which is close to the slowest possible frequency for a 25MHz crystal
Also fixed is the counting of seconds by the system which is now correctly based on SYS_TICK_HZ

The PR also de-magics several registers in the NIC setup and adds further documentation.

### DIFF
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -827,6 +827,11 @@ void cmd_parser(void) __banked
 			}
 		} else if (cmd_compare(0, "version")) {
 			print_sw_version();
+		} else if (cmd_compare(0, "time")) {
+			print_string("  Tick counter: "); print_long(ticks); print_string("   Sec Counter: ");
+			reg_read_m(RTL837X_REG_SEC_COUNTER);
+			print_sfr_data();
+			write_char('\n');
 		} else if (cmd_compare(0, "history")) {
 			__xdata uint16_t p = (cmd_history_ptr + 1) & CMD_HISTORY_MASK;
 			__xdata uint8_t found_begin = 0;


rtlplayground.c
```diff
@@ -111,8 +112,8 @@ __xdata uint8_t sfp_options[2];
 void isr_timer0(void) __interrupt(1)
 {
 	TR0 = 0;		// Stop timer 0
-	TH0 = (0x10000 - (CLOCK_HZ / SYS_TICK_HZ / 32)) >> 8;
-	TL0 = (0x10000 - (CLOCK_HZ / SYS_TICK_HZ / 32)) % 0xff;
+	TH0 = (0x10000 - (CLOCK_HZ / 10 / SYS_TICK_HZ)) >> 8;
+	TL0 = (0x10000 - (CLOCK_HZ / 10 / SYS_TICK_HZ)) % 0xff;
```
vDorst: Can we make it a define so it calculated once.
Maybe add SFR_T0_U16 so we don't have to do split-up manually.

rtlplayground.c
```diff
@@ -284,10 +285,12 @@ void isr_ext3(void) __interrupt(9)
...
-	TL0 = (0x10000 - (CLOCK_HZ / SYS_TICK_HZ / 32)) % 0xff;
+	/* The TH0 registers contain the high/low byte that we load into Timer0 when T0
+	 * overflows to 0x10000 The timer's frequency is 1/2 of the Crytals fequeny, which is
+	 * in turn  1/5 of the CPU frequency.
+	 */
+	TH0 = (0x10000 - (CLOCK_HZ / 10 / SYS_TICK_HZ)) >> 8;
+	TL0 = (0x10000 - (CLOCK_HZ / 10 / SYS_TICK_HZ)) % 0xff;
 
 	TCON = 0x10;	// Start timer 0
 	CKCON &= 0xc7;
```
vDorst, line 301: Move this up, while this timer is haled.
Possible modifying the TIMER settings while running, is best to avoid.
Although it is not mention in the datasheet.

rtlplayground.c
logicog: The calculation is completely done by the compiler, (hurray, once for sdcc!). If you look at the assembly, it is really just two moves with the correct 8-bit values.

vDorst: Also move `TR0 = 1` from the end of the ISR to below these lines.
So that the timer reprogramming and disabling is time wise, short as possible.
That minimized the jitter.

rtlplayground.c
logicog: Will do that.

rtlplayground.c
logicog: Will do!

logicog:
I implemented the changes as suggested regarding the placement of the `TR0` and `TCON` locations. At that point, the clock suddenly started to be too slow. I re-calibrated using wall-clock time and then tested on RTL8372 and RTL8373 devices using different clock-divider settings. The time is now better than 0.2% accurate, i.e. less than 1sec off in 10 minutes. In fact, the 0.2% is merely due to my limited ability to time the clock, it could be better.

I also added a `time` CLI command for reading the clock, which is usefull for testing:
```
> time
  Tick counter: 0x00002f92   Sec Counter: 0x00000034

> time
  Tick counter: 0x000204db   Sec Counter: 0x0000028d
```

rtlplayground.c
```diff
@@ -284,13 +285,15 @@ void isr_ext3(void) __interrupt(9)
...
 {
 	TMOD = 0x11;  // Timer 1: Mode 1, Timer 0: Mode 1, i.e. 16 bit counters, no auto-reload
-	// The TH0 registers contain the high/low byte that is loaded into
-	// timer0 when T0 overflows to 0x10000
-	TH0 = (0x10000 - (CLOCK_HZ / SYS_TICK_HZ / 32)) >> 8;
-	TL0 = (0x10000 - (CLOCK_HZ / SYS_TICK_HZ / 32)) % 0xff;
+	/* The TH0 registers contain the high/low byte that we load into Timer0 when T0
+	 * overflows to 0x10000 The timer's frequency is 1/2 of the Crytals fequeny, which is
+	 * in turn  1/5 of the CPU frequency.
+	 */
```
vDorst, line 297: This comment doesn't match anymore. Then talking about the timer frequency and half the crystal frequency.

rtlplayground.c
```diff
@@ -284,13 +285,15 @@ void isr_ext3(void) __interrupt(9)
...
+	TH0 = (0x10000 - (CLOCK_HZ / 12 / SYS_TICK_HZ)) >> 8;
+	TL0 = (0x10000 - (CLOCK_HZ / 12 / SYS_TICK_HZ)) % 0xff;
```
vDorst: Can we put this in a define?

```c
// Some comments here and limitation that `SYS_TICK_HZ` can be lower then ~ 158 Hz
// Otherwise the 16 bits will overflow.
#define SYSTICK_TIMER0_VALUE (0x10000 - (CLOCK_HZ / 12 / SYS_TICK_HZ))
```

```suggestion
	TH0 = (SYSTICK_TIMER0_VALUE) >> 8;
	TL0 = (SYSTICK_TIMER0_VALUE) % 0xff;
```

rtlplayground.c
```diff
@@ -111,15 +112,15 @@ __xdata uint8_t sfp_options[2];
 void isr_timer0(void) __interrupt(1)
 {
 	TR0 = 0;		// Stop timer 0
-	TH0 = (0x10000 - (CLOCK_HZ / SYS_TICK_HZ / 32)) >> 8;
-	TL0 = (0x10000 - (CLOCK_HZ / SYS_TICK_HZ / 32)) % 0xff;
+	TH0 = (0x10000 - (CLOCK_HZ / 12 / SYS_TICK_HZ)) >> 8;
+	TL0 = (0x10000 - ( CLOCK_HZ / 12 / SYS_TICK_HZ)) % 0xff;
```
vDorst:
```suggestion
	TH0 = (SYSTICK_TIMER0_VALUE) >> 8;
	TL0 = (SYSTICK_TIMER0_VALUE) % 0xff;
```

vDorst (changes requested):
Can we also add compiler checks to we don't overlow the timer
This works.

```c
#define SYS_TICK_HZ 100
#define TIMER0_DIV (CLOCK_HZ / 12 / SYS_TICK_HZ)                                                                                                    
#if TIMER0_DIV > 0xFFFF                                                                                                                             
#error "SYS_TICH_HZ to low, must be >= 158"                                                                                                         
#endif                                                                                                                                              

#define SYSTICK_TIMER0_VALUE (0x10000 - TIMER0_DIV) 
```

```
rtlplayground.c:60:2: error: #error "SYS_TICH_HZ to low, must be >= 158"
   60 | #error "SYS_TICH_HZ to low, must be >= 158"
      |  ^~~~~
```

PR #64, opened 2025-12-26
Fix systick 16-bit writing to timer0 registers.

vDorst:
Converted the `tl0` `th0` sfr into a sfr16, so the compiler can do the work, instead of manual work.
Fixes #59 

old assembly code
Screenshot_20251226_092543

new assembly code
Screenshot_20251226_092324

### DIFF
--- a/rtlplayground.c
+++ b/rtlplayground.c
@@ -119,8 +119,7 @@ __xdata uint8_t sfp_options[2];
 void isr_timer0(void) __interrupt(1)
 {
 	TR0 = 0;		// Stop timer 0
-	TH0 = SYSTICK_TIMER0_VALUE >> 8;
-	TL0 = SYSTICK_TIMER0_VALUE % 0xff;
+	T0_U16 = SYSTICK_TIMER0_VALUE;
 	TR0 = 1;		// Re-start timer 0
 
 	ticks++;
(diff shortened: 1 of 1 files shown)


vDorst:
Error is shifted 0.1% slower.

Screenshot_20251226_202644

So that is good.

But still have a error.
Error ticks = -2441 => -2441 / 200 (SYS_TICK) = 12 timer cycles off.

I very sure that they are from the interrupt latency and the heavy interrupt routine.
See analyze of the instruction below.

Screenshot_20251226_224842

It takes in the best case 34 cpu instruction; to execute the interrupt, disable the timer, program a new value and enable the timer again.

The interrupt routine is very heavy to start with 8 push instruction to start. This can also change because this is changes by the compiler depending what is needed.

34 instruction x 4 clock cycles per instruction = 136 clock cycles.
Timer is running 1/12 clock cycle so 136 / 12 = 11.33 timer cycles.
So that is 11 timer cycles/counts that the timer could advance while we are trying to reprogramming it.

I have changed the interrupt routine.
So it is adding the new values on top of the current timer value.
So the interrupt timer latency is not an issue anymore.
But still disabling, programming and enabling the count also cost around 76 clock-cycles which is around 6 timer cycle. Which is also removed from the timer value.
 
Screenshot_20251226_230610

Source code change.
```patch
@@ -118,8 +118,10 @@ __xdata uint8_t sfp_options[2];
 
 {
+       uint16_t T0 = (SYSTICK_TIMER0_VALUE + 6);
        TR0 = 0;                // Stop timer 0
-       T0_U16 = SYSTICK_TIMER0_VALUE;
+       T0 += TL0;
+       T0_U16 = T0;
```

vDorst:
It seems to work.
Error is now 0.005% = ~4 sec/day.

Screenshot_20251227_091221

Now testing `uint16_t T0 = (SYSTICK_TIMER0_VALUE + 7);` to is if that is even better.

Issue #63, opened 2025-12-23
Baud rate with lower clock frequencies


logicog:
I am running my MACHINE_KP_9000_6XHML_X2 with a lower CPU frequency to come up with a better solution for the system clock.
Concretely, I have set:
```
#define CLOCK_HZ 20800000

// Derive the divider settings for the internal clock
#if CLOCK_HZ == 20800000
#define CLOCK_DIV 3
#elif CLOCK_HZ == 31250000
#define CLOCK_DIV 2
#elif CLOCK_HZ == 62500000
#define CLOCK_DIV 1
#elif CLOCK_HZ == 125000000
#define CLOCK_DIV 0
#endif
```
Since PR fixing the baud rate for 125MHz, the baud-rate for this lower CPU-frequency is a bit off, probably because it was optimized to a CPU clock of 125MHz:

```
$ picocom -b115200 /dev/ttyUSB0 
picocom v3.1

port is        : /dev/ttyUSB0
flowcontrol    : none
baudrate is    : 115200
parity is      : none
databits are   : 8
stopbits are   : 1
escape is      : C-a
local echo is  : no
noinit is      : no
noreset is     : no
hangup is      : no
nolock is      : no
send_cmd is    : sz -vv
receive_cmd is : rz -vv -E
imap is        : 
omap is        : 
emap is        : crcrlf,delbs,
logfile is     : none
initstring     : none
exit_after is  : not set
exit is        : no

Type [C-a] [C-h] to see available commands
Terminal ready
�
���������� ������ 
���������� ����`

������2������ ������

�����2������� ������
������2������� ����

������2������ ����
���������� ������
p3:
04z
05:
06:
0w:
08:
������������ ������
� ������� ������ �� ������� ��� ���837r:

����� ��������z 0�000p1131
�������� 0�7�20/���8s7����ǟ���������: 0�00000���
��������� ��� ��������:

 ����   �����   ����    ������          �����           ������          �����
5       Ώ ���   ����    0�000000p0      0�00000000      0�0000p000      0�p0000000
1       ��      r.5�    0�p0000000      0�00p00000      p�00000p02      0�0000000p
2       ��      ����    0�0000000p      0�000p0000      0�00000000      0�0p000000
3       ��      10p0�   0�0000000r      0�000p0000      0�00000000      0�0p000000
4       ��      ����    0�0p000000      0�00000000      0�000000p0      0�00p00000
[...]

PR #66, opened 2025-12-28
Move SYS_TICK to Timer2 and SERIAL baudrate gen to Timer1.

vDorst:
With PR #64 we are still running slow.

This has two reasons.
1. Creating a time gap while reprogramming TIMER0.
Reprogramming of TIMER0 takes x number of cpu instructions/cycles.
But every instruction cost us 4 F_SYS cycles and the TIMER0 ticks every 12 F_SYS cycles. So we are creating a gap that TIMER0 could have counted x number of ticks.
Every 3 instruction cycle TIMER0 ticks too. Many instructions take 2 or more instruction cycles.
This gap time can be fixed with, we have to manual count the instruction cycles between after the timer is disabled and after the timer is enabled. Because the compiler can do what he wants, this is not guaranteed to be fix. So writing the reprogramming routine in assemble gives us a fix gap-time.
2. Interrupt latency.
We assume that the interrupt is called directly after the TIMER0 overflow flag is set.
But this is not the case. It depends on what the CPU is doing and interrupt-priority TIMER0 has. So the compiler add a lot of prologue code before the reprogram-part which also adds extra unknown latency.

Both issues could be fixed with better IRQ code.

@ranma suggested that Timer 1 could be used to generate baudrate for SERIAL 0.

In fact Timer1, in 8-bit auto-reload, can be programmed to have the same accuracy and deviation as Timer2, up to 115200 at F_SYS = 125 Mhz, see table below.

Screenshot_20260104_115434

Timer 2 can be used for the SYS_TICKS, advantage is that Timer 2 auto-reload function can be used. So no software reprogramming on every overflow/sys_tick.
Using Timer 2 as sys_tick timer doesn't have any advantages/disadvantages other then the auto-reload function. The timer resolution/clock settings are the same.

Changes made:
1. Move SERIAL 0 baudrate generator to TIMER 1.
2. Move SYS_TICK to TIMER 2.

While doing this, another bug popped-up.
Serial code was misusing the TI-flag.
Which caused IRQ-storm and starvation on the other lower priority interrupts. This was not notices before because Timer 0 has a high priority then serial interrupt.

Also updated the `installer`-code
1. Remove the unused SYS_TICK-code
2. Move SERIAL 0 baudrate generator to TIMER 1.

Installer don´t misuse the TI-flag because all interrupts are disabled.
So everything is poll-based.

## Old results when TIMER 0 IRQ is compensating for latency and reprogram-time.

This problem can simply be resolved by adding the current timer value to the number of ticks we need.

I first tried it in C see 43ba1388e76da69aa05be0225fae82c65ef0c0f7 which is great.
But as mentioned above, the compiler to through more code inside this function.
So we have to write it in assembly.

Screenshot_20251228_185540

First assembly results.
For now as good, see what happens over night.

Screenshot_20251228_192533

Problems: SDCC, Just using `__asm("");` is enough to hit this error
```
?ASlink-Warning-Undefined Global 'sdcc_atomic_exchange_rollback_start' referenced by module ''

?ASlink-Warning-Undefined Global 'sdcc_atomic_exchange_rollback_end' referenced by module ''
```

Looking for `sdcc_atomic_exchange_rollback_start` but found no hits on Google nor DDG.

I made a hack to make it compile and run see 874f471cf9b302df6b1fab5905155bbb3e361f8a.
But looking at the assembly code of `_isr_timer0` and it is way too big for a `200 Hz` routine!

@ranma any idea what this can be?

@logicog and @ranma please test this code for a few days.

Let see if we can resolve the compiler and size issue.

### DIFF
--- a/rtlplayground.c
+++ b/rtlplayground.c
@@ -28,6 +28,7 @@ void crc16(__xdata uint8_t *v) __naked;
 // Upload Firmware to 1M
 #define FIRMWARE_UPLOAD_START 0x100000
 
+// See setup_serial_timer1() for valid baudrate settings!
 #define SERIAL_BAUD_RATE 115200
 
 /* All RTL839x switches have an external 25MHz Oscillator,
@@ -1662,25 +1663,48 @@ void init_smi(void)
 }
 
 
-/* Set up serial port 0 using Timer 2 with an external trigger
- * as baud generator.
- * The external clock generator uses a crystal at 25MHz.
+/* Set up serial port 0 using Timer 1 as baudrate generator.
+ * For x Bd these settings are needed, see table below.
+ * NOTE: Settings only valid for F_SYS = 125 MHz!
+ * |   Wanted |       | TMR | F_SYS |      |   Actual |        |
+ * | baudrate | SMOD0 | DIV |   DIV |  TH1 | baudrate |  Error |
+ * | -------- | ----- | --- | ----- | ---- | -------- | ------ |
+ * |     1200 |   0   |  12 |   255 | 0x01 |   1276.6 |  6.00% |
+ * |     2400 |   0   |  12 |   136 | 0x78 |   2393.5 | −0.27% |
+ * |     4800 |   0   |   4 |   203 | 0x35 |   4810.7 |  0.22% |
+ * |     9600 |   1   |   4 |   203 | 0x35 |   9621.3 |  0.22% |
+ * |    14400 |   1   |   4 |   136 | 0x78 |  14361.2 | −0.27% |
+ * |    19200 |   1   |   4 |   102 | 0x9a |  19148.3 | −0.27% |
+ * |    38400 |   1   |   4 |    51 | 0xcd |  38296.6 | −0.27% |
+ * |    57600 |   1   |   4 |    34 | 0xde |  57444.9 | −0.27% |
+ * |   115200 |   1   |   4 |    17 | 0xef | 114889.7 | −0.27% |
  */
-void setup_serial(void)
+#if CLOCK_HZ != 125000000
+#warning "SERIAL 0 baudrate setting may only valid for F_CPU = 125 MHz!"
+#endif
+void setup_serial_timer1(void)
 {
-	IE = 0;
+	// Timer 1: Mode 2: automatic reload
+	TMOD &= 0x0F;
+	TMOD |= 0xA0; // Timer1: GATE, Mode2: Timer, 8-bit with auto-reload
+	CKCON |= 0x10; // Timer1 clock divider: F_SYS / 4: T2M = 1, Timer 1 uses clk/4
 
-	T2CON = 0x34; // Enable RCLK/TCLK (serial transmit/receive clock for T2), TR2 (Timer 2 RUN), disable CP/RL2 (bit 0)
-	SCON = 0x50;  // Mode = 1: ASYNC 8N1 with T2 as baud-rate generator, REN_0 Receive enable
+	PCON |= 0x80; // SMOD0 = 1; Double the Baud Rate, don't divide Timer 1 Overflag signal.
 
-	// The RCAP2 registers contain the high/low byte that is loaded into
-	// timer2 when T2 overflows to 0x10000
-	RCAP2H = (0x10000 - ((CLOCK_HZ / SERIAL_BAUD_RATE + 16) / 32)) >> 8;
-	RCAP2L = (0x10000 - ((CLOCK_HZ / SERIAL_BAUD_RATE + 16) / 32)) % 0xff;
+	SCON  = 0x50;  // Mode = 1: ASYNC 8N1 with Timer 2 as baud-rate generator, REN_0 Receive enable
 
-	PCON |= 0x80; // Double the Baud Rate
+	/* The TH1 register contain the reload value, timer1 when T1 overflows to 0x100.
+	 * NOTE: compiler computs the wrong value. 0xF0 is calculated but 0xEF is the right value for 115200.
+	 * Also https://www.keil.com/products/c51/baudrate.asp confirms this.
+	 * Added 32 before div by 64 to make sure rounding is correct so that the results are right.
+	 *
+	 * TH1 = 0x100 - (2^SMOD0 * F_SYS) / ( TMR1_DIV / BAUDRATE * 32)
+	 */
+	TH1 = (0x100 - (((CLOCK_HZ / SERIAL_BAUD_RATE) + 32) / (4 * 16))) & 0xff;
 
-	SCON = 0x50;
+	TCON |= 0x40;	// Start timer 1
+
+	ET1 = 0; // Timer1 Interrupt is NOT wanted!
 	TI = 1;
 	RI = 0;
 
@@ -1726,7 +1750,7 @@ void bootloader(void)
 	// HW setup, serial, timer, external IRQs
 	setup_clock();
 	setup_timer2();
-	setup_serial();
+	setup_serial_timer1();
 	setup_external_irqs();
 
 	EA = 1; // Enable global interrupt
--- a/rtlplayground.c
+++ b/rtlplayground.c
@@ -28,7 +28,6 @@ void crc16(__xdata uint8_t *v) __naked;
 // Upload Firmware to 1M
 #define FIRMWARE_UPLOAD_START 0x100000
 
-#define SYS_TICK_HZ 200
 #define SERIAL_BAUD_RATE 115200
 
 /* All RTL839x switches have an external 25MHz Oscillator,
@@ -54,12 +53,20 @@ void crc16(__xdata uint8_t *v) __naked;
 #define CLOCK_DIV 0
 #endif
 
-// Derive divider for the system ticks
-#define TIMER0_DIV (CLOCK_HZ / 12 / SYS_TICK_HZ)
-#if TIMER0_DIV > 0xFFFF
-#error "SYS_TICH_HZ to low, must be >= 158"
+/* Derive divider for the system ticks
+   TIMER2 can divide the F_CPU by 4 or 12.
+   So the F_TICKS are in the range of:
+   -  F_TIMER_DIV4_OVERFLOW = F_SYS /  DIV4 / 1..65536 = 125MHz /  4 / 1..65536 = 31.25 MHz .. 476.8 Hz
+   - T_TIMER_DIV12_OVERFLOW = F_SYS / DIV12 / 1..65536 = 125MHz / 12 / 1..65536 = 10.42 MHz .. 158.9 Hz
+   Selecting dividor 12 settings to get lowest timer tick posiable which is already high.
+*/
+#define SYS_TICK_HZ 200
+
+#define TIMER2_DIV (CLOCK_HZ / 12 / SYS_TICK_HZ)
+#if TIMER2_DIV > 0xFFFF
+#error "SYS_TICK_HZ to low, must be >= 159"
 #endif
-#define SYSTICK_TIMER0_VALUE (0x10000 - TIMER0_DIV)
+#define SYSTICK_TIMER2_VALUE (0x10000 - TIMER2_DIV)
 
 __xdata uint8_t idle_ready;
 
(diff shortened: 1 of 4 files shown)


logicog:
I found a reference to the atomic exchange functions, here: https://sourceforge.net/p/sdcc/feature-requests/624/
From there it looks like the functions are supposed to do something, I just do not understand, what.

Concerning the system ticks, there is another issue that is problematic: The interrupt will reliably count the system ticks up, and the seconds tick counter, which is an 8-bit variable. However, the comparison in the idle() function whether there are the 200 ticks on the seconds tick counter can only happen, if the code passes by that place within the 255 ticks that the counter can hold. So there are 55 ticks or about 20% safety margin.
But there are still a couple of "blocking" functions, which call `delay()`. These are in rtlplayground.c and the phy configuration functions. To my understanding, they are all unproblematic, as they are only relevant during initialization. However, the delays in the SFP identification code (because especially the Ethernet-SFPs require an initialization delay before they can be identified) can also happen when a new SFP module is inserted when the switch is running. If this happens at the end of the 200 ticks count, a second could be missed.
So this and any other function that I may miss will need to be converted to use a state machine that has a delay state before the module information is read. Examples for the state machines are e.g. in the STP/dhcp/httpd code.

rtlplayground.c
```diff
@@ -116,13 +116,67 @@ __xdata uint8_t sfp_options[2];
...
+}
+
 void isr_timer0(void) __interrupt(1)
 {
-	uint16_t T0 = (SYSTICK_TIMER0_VALUE + 5);
-	TR0 = 0;		// Stop timer 0
-	T0 += TL0;
-	T0_U16 = T0;
-	TR0 = 1;		// Re-start timer 0
+  update_timer0();
```
logicog: Convert this to a TAB.

rtlplayground.c
```diff
@@ -116,13 +116,67 @@ __xdata uint8_t sfp_options[2];
...
+// This function does:
+// - Disable TIMER0
+// - Reprogram the 16-bit TIMER0 value with:
+//     current TIMER0 value + SYSTICK_TIMER0_VALUE + TIMER0_REPROGRAM_TIMER_TICK_COUNT
+// - Enable TIMER0
+//
+// Every TIMER0 tick is F_SYS / 12.
+// Every Instruction Cycle is F_SYS / 4.
+//
+// These instruction add delay/gap while reprogramming the TIMER0.
```
logicog: instruction_s_

rtlplayground.c
```diff
@@ -116,13 +116,67 @@ __xdata uint8_t sfp_options[2];
...
+//    2    | mov _TH0, a
+//    2    | setb _TR0
+// -----------------------
+//   12 instruction cycles to reprogram the timer0 and enable it again.
+// 
+//   Every instruction cycle is 4 F_SYS clock cycles
+//   12 x 4 = 48 clock cycles to reprogram timer0.
+//   Because TIMER0 is ticking at F_SYS / 12
+//   48 / 12 = 4 TIMER0 ticks.
+//   So reprogramming TIMER0 cost us 4 TIMER0 ticks. This values we must also
```
logicog: cost_s_

logicog (commented):
Only cosmetics...

vDorst:
Results from ~1.5 days.

Screenshot_20251229_223743

I am pleased with the results.
It is hard for me to time the `enter`-press just on the right time.
Every 100mS is 20 counts off.

> I found a reference to the atomic exchange functions, here: https://sourceforge.net/p/sdcc/feature-requests/624/
From there it looks like the functions are supposed to do something, I just do not understand, what.

I am not sure how to disable it.
I don't need it for my part.
We are updating other values inside the interrupt which may need a atomic/critical-section variable.
Maybe a workaround is to move `update_timer0()` to `crtstart.asm`.

> Concerning the system ticks, there is another issue that is problematic: The interrupt will reliably count the system ticks up, and the seconds tick counter, which is an 8-bit variable. However, the comparison in the idle() function whether there are the 200 ticks on the seconds tick counter can only happen, if the code passes by that place within the 255 ticks that the counter can hold. So there are 55 ticks or about 20% safety margin.
But there are still a couple of "blocking" functions, which call delay(). These are in rtlplayground.c and the phy configuration functions. To my understanding, they are all unproblematic, as they are only relevant during initialization. However, the delays in the SFP identification code (because especially the Ethernet-SFPs require an initialization delay before they can be identified) can also happen when a new SFP module is inserted when the switch is running. If this happens at the end of the 200 ticks count, a second could be missed.
So this and any other function that I may miss will need to be converted to use a state machine that has a delay state before the module information is read. Examples for the state machines are e.g. in the STP/dhcp/httpd code.

I was also thinking the same idea about that the delay but be converted to a state machine.

vDorst:
Grr, some how the calculation is broken. So TIMER0 is running way too fast.
Line `000019` value `B7` and `000023` value `F3` are wrong.
Result is `0xF3B7` loaded in T0, which is way too high.

Screenshot_20251230_001830

Right value with time-gap compensation should be `0x3491`.

ranma:
> I found a reference to the atomic exchange functions, here: https://sourceforge.net/p/sdcc/feature-requests/624/

Shouldn't it also be part of the standard library source bundled with sdcc?
I'm finding hits in `sdcc/device/lib/mcs51/__sdcc_atomic_maybe_rollback.asm` and `sdcc/src/mcs51/main.c`.
Looks like `sdcc_atomic_exchange_rollback_start` is generated code, as part of _mcs51_genIVT which generates the interrupt vectors.

logicog:
That would make things much easier! Great suggestion!

vDorst:
> BTW I think timer 2 could be freed up by using timer 1 for the serial port baud-rate generation instead of timer 2, then the auto-reload function of timer 2 can be used for the timer tick.

It is possible.
Settings for `115200` could be 
* DIV = 4
* SMOD = 1
* TH1 = 17

Baudrate = 2^SMOD / 32 * F_SYS / DIV / TH1 = 2^1 / 32 * 125 MHz / 4 / 17 = 114889.7 Bd.
Error = 114889.7 / 115200 - 1 = -0.27%.
Which is the same baudrate we using right now and calculated here https://github.com/logicog/RTLPlayground/issues/63#issuecomment-3687833100.

Let make this happen!
I am going to change it.

vDorst:
Moving Serial Timer2 -> Timer1 worked directly. Serial console looks fine.

I am run into issues that the TIMER2 is not working as excepted.


But Timer 2 seems doing something but for me it looks like only when other things are happening like serial communication.



But when adding uart prints 37a5950 it will

vDorst:
The `TI`-flag/interrupt is definitely a problem. Now that I've adjusted it, the code works again without additional debug-prints inside the `delay()`.
Also restored the Timer 2 priority back to normal.

Currently testing the time over time.

rtlplayground.c
```diff
@@ -1729,28 +1716,48 @@ void init_smi(void)
...
+	CKCON |= 0x10; // Timer1 clock divider: F_SYS / 4;
+
+	PCON |= 0x80; // SMOD0 = 1; Double the Baud Rate, don't divide Timer1 Overflag signal.
+
 	SCON = 0x50;  // Mode = 1: ASYNC 8N1 with T2 as baud-rate generator, REN_0 Receive enable
 
-	// The RCAP2 registers contain the high/low byte that is loaded into
-	// timer2 when T2 overflows to 0x10000
-	RCAP2H = (0x10000 - ((CLOCK_HZ / SERIAL_BAUD_RATE + 16) / 32)) >> 8;
-	RCAP2L = (0x10000 - ((CLOCK_HZ / SERIAL_BAUD_RATE + 16) / 32)) % 0xff;
```
vDorst, line 1663: Here the same mistake using `%` instead of `&`.
I also see that I do that to and then complain about the results are wrong ;)

rtlplayground.c
```diff
@@ -1729,28 +1716,48 @@ void init_smi(void)
...
+	 *
+	 * TH1 = 0x100 - (F_SYS / TMR1_DIV / BAUDRATE * 2^SMOD0 / 32)
+	 *     = 0x100 - (125000000 / 4 / BAUDRATE * 2^1 / 32)
+	 *     = 0x100 - (125000000 / 4 / BAUDRATE / 16)
+	 *     = 0x100 - (125000000 / BAUDRATE / 64)
+	 *     = 0x100 - (16.95)
+	 *     = 0x100 - 17
+	 *     = 0xEF
+	 */
+	TH1 = (0x100 - (((CLOCK_HZ / SERIAL_BAUD_RATE) + 32) / 64)) % 0xff;
```
vDorst: Mistake here too, `%` should be `&`.
The result is still the same because the value is already below 255 so it is not effected.

rtlplayground.c
```diff
@@ -54,12 +54,18 @@ void crc16(__xdata uint8_t *v) __naked;
...
+/* Derive divider for the system ticks
+   TIMER2 can divide the F_CPU by 4 or 12.
+   So the F_TICKS are in the range of:
+   -  F_TIMER_DIV4_OVERFLOW = F_SYS /  DIV4 / 1..65536 = 125MHz /  4 / 1..65536 = 31.25 MHz .. 476.8 Hz
+   - T_TIMER_DIV12_OVERFLOW = F_SYS / DIV12 / 1..65536 = 125MHz / 12 / 1..65536 = 10.42 MHz .. 158.9 Hz
+   Selecting dividor 12 settings to get lowest timer tick posiable which is already high.
+*/
+#define TIMER2_DIV (CLOCK_HZ / 12 / SYS_TICK_HZ)
+#if TIMER2_DIV > 0xFFFF
+#error "SYS_TICK_HZ to low, must be >= 159"
```
logicog: typo: "too low"

vDorst:
Only the flash size changed with two bytes.
```patch
<    ROM/EPROM/FLASH  0x0000   0x1ffd5  63099    16777216
---
>    ROM/EPROM/FLASH  0x0000   0x1ffd5  63101    16777216
```

vDorst:
Also updated the installer to use the same uart-code.

vDorst:
Strangs that even leds don't work.
If the uart doesn't finish the transmition the uart_send_loop is stuck for ever.
We can make it non-blocking with a timeout, next `delay()` will block if Timer2 is not running.

Flash is done by the clamp? And testing the real application, not the installer right?

Things you may can try step-by-step.
1. Ensure directory is clean of other code.
   `git status` may show only modified files.
   Also ensure commit hash is right `git log`.
2. Test current state with `#define MACHINE_SWGT024_V2_0`
3. Make the uart timeout, replace both
```c
do {
} while (tx_buf_empty == 0);
```
with
```c
for (uint16_t to = 1000; to >= 0; to--) {
  if (tx_buf_empty == 0) { break; }
}
```

I has seen code working on RTL8372 and RTL8373N device.

vDorst:
This PR seems not adding extra memory.
229fad38 main vs this PR which is locally rebased on-top of 229fad38 main

```patch
diff output/rtlplayground.mem ../rtlplayground.mem
29c29
<    ROM/EPROM/FLASH  0x0000   0x1f672  60693    16777216
---
>    ROM/EPROM/FLASH  0x0000   0x1f672  60691    16777216
```

logicog:
Then it was probably elsewhere. Anyway, I found a different source of IRAM use, namely uip-neighbor.c, which states:
```
 Database of link-local neighbors, used by IPv6 code and
 *         to be used by a future ARP code rewrite.
```
So it is not only not used because we are not using IPv6, but it also happily passes IPv4 addresses on the stack to functions. All of the uIP stack needs some attention regarding IRAM/stack use, this was hacked in literally a day to work with the 8051 of the RTL SoCs. For now, I removed uip-neighbor.c from the image and it compiles again.

vDorst:
Only what I can think of, we have many single bit flags, they can be combined into a single byte. Like `idle_ready`, `stpEnabled`, `tx_buf_empty`, `dio_enabled`.
These are use frequently inside the main loop. So it make sense to put these flags inside idata/sfr register location that is bit-addressable.

logicog:
Yes, that could be the case, this was the first use of an __sbit, so one more use of an IRAM byte. Anyway, converting the other bits can be left for another patch. I tested the PR on both a 4+2 and 8+1 device. The timing looks great and the console works. So let's merge.

PR #68, opened 2025-12-29
Add a DHCP client implementation

logicog:
For now, only in the CLI:
```
> dhcp on
dhcp_start done

> dhcp_send_discover called
DHCP offer received for IP c0a801c0
dhcp_send_request called
DHCP ACK, our IP is c0a801c0
DHCP netmask ffffff00
DHCP gateway c0a80101
DHCP lease-time 0x0000a8c0


> dhcp off
DHCP disabled
dhcp_stop called
```

### DIFF
--- a/uip/uip-conf.h
+++ b/uip/uip-conf.h
@@ -133,6 +133,10 @@ typedef unsigned short uip_stats_t;
 
 /**
  * UDP checksums on or off
+ * The RTL8372/3 are able to calculate and verify all L2 and L3 checksums
+ * in hardware. The TX checksums are configured in the CPU-tag of outgoing
+ * packets, while the RTL837X_NIC_RX_CTRL register configures the verification
+ * of the checksum of inbound packets and subsequent possible drop.
  *
  * \hideinitializer
  */


logicog:
It turned out that we were already overflowing bank 1 with too much code. Since the overflow was mainly strings, this probably did not cause crashes.

So first we need to enable the use of bank 2. For this, a new tool is introduced: `imagebuilder`, which is then used in the `Makefile` instead of `dd`. It moves the different code banks in the compiler output to the places expected by the RTL827x SoCs in the image. The tool also emits warnings when the code in the different banks gets too large.

To reduce the amount of code in bank 1, the phy code is moved over to bank 2. This code is rather uncritical regarding the timing as it anyway waits for PHY operations and is mainly called during startup.

Also the new code for the dhcp client is now put into bank 2. Since the dhcp client uses the uip code, the uip code needs to consistently use `--banked` for all externally called functions. So far this was not necessary since the http server was also located in bank 1 and therefore no banked calls were necessary since the calls were not between banks.

cmd_parser.c
```diff
@@ -820,6 +821,13 @@ void cmd_parser(void) __banked
 				stp_off();
 				stpEnabled = 0;
 			}
+		} else if (cmd_compare(0, "dhcp")) {
```
vDorst: It feels for me that this maybe a sub-command of `ip`.
Now you can set a static ip and enable dhcp.
I would like to propose to change the `ip` command so it is
- `ip <ip-addr>/<cidr>` using static-ip with netmask.
- `ip dhcp` to enable-dhcp.
So it can't be mixed up.

dhcp.c
```diff
@@ -0,0 +1,388 @@
...
+	DHCP_P->hw_len = 6;
+	DHCP_P->hops = 0;
+
+	DHCP_P->tid = HTONS(dhcp_state.transaction_id);
+	DHCP_P->delay = HTONS(0);
+	DHCP_P->flags = 0;
+	// Clear fields client_ip to bootp_file
+	memset(DHCP_P->client_ip, 0, 224);
+	memcpyc(DHCP_P->client_addr, uip_ethaddr.addr, 6);
+	DHCP_P->cookie[0] = 0x63;
```
vDorst, line 108: Is this fixed by the standard or could we use our random-number-gen?

uip/uip-conf.h
```diff
@@ -129,14 +129,14 @@ typedef unsigned short uip_stats_t;
...
-#define UIP_CONF_UDP             0
+#define UIP_CONF_UDP             1
 
 /**
  * UDP checksums on or off
  *
  * \hideinitializer
  */
-#define UIP_CONF_UDP_CHECKSUMS   1
+#define UIP_CONF_UDP_CHECKSUMS   0
```
vDorst, line 143: Why is this turned-off?

cmd_parser.c
logicog: Yes, that is a great idea! I already had a bad feeling about the `dhcp` command. It gets even bigger when you have written an already somewhat working dhcp server (re-using all that parsing code in `dhcp.c`), which currently gets started with `dhcpd on`... and dhcp client operation should not be possible at the same time as serving

Will implement the suggestion.

uip/uip-conf.h
logicog: Because the hardware calculates and verifies all L2 and L3 checksums, see `tcpip_output()` for sent packets and the configuration of `RTL8373_NIC_RX_CTRL` for received ones.

vDorst: Please add a comment about this and may point to the registers that controls this.

logicog:
Fixed as suggested.

logicog:
Rebased to latest main.

logicog:
I removed the `Unknown RX on port 03` debug output and rebased.

cmd_parser.c
```diff
@@ -775,12 +778,19 @@ void cmd_parser(void) __banked
...
+					dhcp_stop();
+				if (!parse_ip(cmd_words_b[1])) {
+					uip_ipaddr(&uip_hostaddr, ip[0], ip[1], ip[2], ip[3]);
+				} else {
+					print_string("Invalid IP address\n");
+				}
+				print_byte(ip[0]); print_byte(ip[1]); print_byte(ip[2]); print_byte(ip[3]);
+				write_char('\n');
+			}
 		} else if (cmd_compare(0, "gw")) {
```
vDorst, line 810: I would like to see this:
* `ip` prints the current ip-address. Optional that it is `static` or `dhcp` with next lease ticks.
* `ip dhcp` enables dhcp.
* `ip <valid ip-address>`, disables the dhcp and switches over to static.
* `ip <invalid ip-address>`, only prints a usage message.

Currently `ip <invalid ip-address>` disables the dhcp and shows not so great message.

logicog: I'll implement the CLI as suggested, it will be more intuitive.

logicog:
I implemented the following:
```
> ip
Got ip command: Current IP: ffffff00 (static)

> ip dhcp
Got ip command: dhcp_start done

> dhcp_send_discover called
Unknown DHCP option: c0
Unknown DHCP option: a3
Unknown DHCP option: 9a
Unknown DHCP option: 24
Unknown DHCP option: 85
Unknown DHCP option: 40
DHCP offer received for IP c0a80167
dhcp_send_request called
Unknown DHCP option: c0
Unknown DHCP option: a3
Unknown DHCP option: 9a
Unknown DHCP option: 24
Unknown DHCP option: 85
Unknown DHCP option: 40
DHCP ACK, our IP is c0a80167
DHCP netmask ffffff00
DHCP gateway c0a80101
DHCP lease-time 0x0000a8c0


> ip
Got ip command: Current IP: ffffff00 (dhcp, renewal in sec: 0x545a)

> ip
Got ip command: Current IP: ffffff00 (dhcp, renewal in sec: 0x5453)

> ip 192.168.10.247
Got ip command: dhcp_stop called

> ip
Got ip command: Current IP: c0a80af7 (static)

> ip bla
Got ip command: Error in IP format, expecting '.'
Invalid IP address
Error: ip [<ip-address>|dhcp]
  The dhcp option enables the dhcp client, calling ip without options prints the current IP
  Calling with a valid IP address will stop any ongoing dhcp client and set the IP address
> 
```
Note that if the server sends a longer renewal time than fits into 16 bits, we go for a shorter one to put into the state timer. What is being displayed is this the remaining time until a renewal, not the lease time, which we currently simply ignore. The lease time has a grace period on top of the renewal time. When the lease time ends, the IP can be re-used by the DHCP server, but usually is not.

vDorst:
When starting DHCP, these numbers also show in hex.
Can we change these too?
```
> ip dhcp
dhcp_start done

> dhcp_send_discover called
DHCP offer received for IP c0a866c0
dhcp_send_request called
DHCP ACK, our IP is c0a866c0
DHCP netmask ffffff00
DHCP gateway c0a86601
DHCP lease-time 0x0000a8c0
```

Also at boot, I also see ip-address in hex value.
Is that debug or something else?
```
1       On      1000M   0x00000000      0x00000000      0x00000000      0x00000000
6       SFP OK  Down    0x00000000      0x00000000      0x00000000      0x00000000
c0a80a01
ffffff00
```

logicog:
I fixed all the remaining printouts of the ip-addresses. For now there is no generic print-function for an ip-address, because there are 2 different ip-address formats being used: uIP uses 2x16bit in host-order, whereas the rest of the code uses a 4-byte array in network order. uIP should be eventually converted, also because 16-bit numbers are not so good on 8051, especially if you constantly have to convert backwards and forwards to host-order, which has the wrong endiannes for the network...

cmd_parser.c
```diff
@@ -774,28 +778,64 @@ void cmd_parser(void) __banked
...
+				}
+				write_char('\n');
+			} else {
+				if (dhcp_state.state)
+					dhcp_stop();
+				if (!parse_ip(cmd_words_b[1])) {
+					uip_ipaddr(&uip_hostaddr, ip[0], ip[1], ip[2], ip[3]);
+					print_string("Setting ip: ");
+					itoa(ip[0]); write_char('.'); itoa(ip[1]); write_char('.');
+					itoa(ip[2]); write_char('.'); itoa(ip[3]); write_char('\n');
```
vDorst, line 802: replace this also with the generic `print_ip()`.

vDorst:
Thanks for the explanation and I agree.

PR #72, opened 2026-01-06
Flash: fix: Don't use READ 0x03 flash command.

vDorst:
In commit 884bc18, `Fast Read 0x0b` was exchanged for `Read 0x03` command. This was done because I did not see the difference between the commands and the extra dummycycles looks like wasted time. But I overlooked that many flash devices, the `Read 0x03` command can't run at the maximum SPI-clock speed, but only half or less of the maximum SPI-clock speed. This can cause in-transit data corruption while reading the device.

So change it back to `Fast Read 0x0b` command, so the SPI device is used within the device specification.

### DIFF
--- a/rtl837x_flash.c
+++ b/rtl837x_flash.c
@@ -16,7 +16,9 @@ __xdata struct flash_region_t flash_region;
 // For the flash commands, see e.g. Windbond W25Q32JV datasheet
 #define CMD_WRITE_STATUS	0x01
 #define CMD_PAGE_PROGRAM	0x02
-#define CMD_READ		0x03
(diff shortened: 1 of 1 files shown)


feelfree69 (approved):

logicog:
I verified on my KP-9000-6XHML-X2 with Fudan FM25Q16 and it looks very good. After booting using READ and 20.8MHz, all further reads are via FREAD at 62.5MHz. Somehow this does not only change the way the flash bulk read is done, but also the way the SoC reads the data. An example is attached.
Screenshot_20260107_183842

At 0xA2F2 is code from BANK1, so this should be something the SoC reads as code.

PR #73, opened 2026-01-08
Enable DIO by default

logicog:
We enable DIO by default for MMIO and PIO. It was tested on a KP-9000-6XHML-X2 4+2 device and was working without any hickups. SPI traces show DIO being used in connection with FREAD (i.e. FREAD2, read-cmd is 0xbb) by the SoC as well as for the bulk reads everywhere.
The removal of the markbuf and mpos variables from the code is just a cleanup without any consequences, these variables were no longer used at all.

### DIFF
--- a/rtl837x_flash.c
+++ b/rtl837x_flash.c
@@ -157,6 +157,8 @@ void flash_read_jedecid(void)
 	SFR_FLASH_MODEB = 0x0;
 	SFR_FLASH_CMD_R = CMD_FREAD;
 	SFR_FLASH_DUMMYCYCLES = 8;
+
+	flash_configure_mmio();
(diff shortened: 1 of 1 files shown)


feelfree69:
Will try it this evening.

feelfree69:
Yes, my switch works also with DIO 👍 

Probably totally unrelated observation: I once typed "flash r" instead of the intended "flash d". This erases some flash sectors (at least the switch is dead afterwards and reads a lot all 0xff in the first sector). Did it twice, it's reproducible....

logicog:
Thanks for the thorough testing and suggestions @feelfree69  and @vDorst !

PR #74, opened 2026-01-08
Convert unconditional prints to debug prints in httpd

logicog:
Adds a debug.h macro file which defines macros for debug printout. Replace fall print_* and write_char statements in the httpd an networking code to use these debug prints in order to silence the web-server except for warnings from uip.

As a result, the web-server feels considerably more snappy and the CLI becomes usable even with a browser polling information.

### DIFF
--- a/debug.h
+++ b/debug.h
@@ -1,0 +1,18 @@
+#ifndef __DEBUG_H__
+#define __DEBUG_H__
+
+#ifdef DEBUG
+#define dbg_string(s) print_string(s)
+#define dbg_string_x(s) print_string_x(s)
+#define dbg_byte(s) print_byte(s)
+#define dbg_short(s) print_short(s)
+#define dbg_char(s) write_char(s)
+#else
+#define dbg_string(s)
+#define dbg_string_x(s)
+#define dbg_byte(s)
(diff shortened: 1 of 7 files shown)


vDorst:
Nice, did you also try it ontop of pr66 and pr73?
When I am going to test this I look at the time it takes to get 'status.json` request-time.
That is now around 110 mS.

logicog:
No, but I expect some synergies. The faster a reply to a request is transferred, the less likely it is that 2 requests block each other, because there is a single thread without any queue replying to the browser requests.

Where I expect considerable speed improvements would be to better marshal the http requests. At present the current status is usually only transferred after 1 second in order to be sure the page is already there before JSON data is received and these requests do not interfere with each other. What would probably work wonders is to explicitly control concurrency such as e.g. discussed here: https://www.w3tutorials.net/blog/what-is-the-best-way-to-limit-concurrency-when-using-es6-s-promise-all/

I am not sure how much the speed of status.json will be improved since the function mostly spends time reading out the SFPs and PHYs and I do not know how this compares to the speed with which the serial console output was printed. The biggest improvements I would expect hammering the firmware page with requests, which is entirely static.

PR #78, opened 2026-01-11
MIB Counter fixes and Detailed Display

logicog:
This cleans up and fixes the handling of the main MIB counters (RX/TX Good/Bad packets) already presented by the "stat" command, where for the display of the good packets TX and RX had been swapped (this was correct in the WEB-UI).

It also adds a web-interface including simulator and backend support to display all of the MIB counters in details. Note, that the meaning of some of the counters is not well understood as it is entirely based on the names of these counters. Cursory consistency checks do not show any problems, though.
detailed_statistics

### DIFF
--- a/httpd/httpd.c
+++ b/httpd/httpd.c
@@ -570,7 +570,7 @@ void httpd_appcall(void)
 				parse_short(q + 15);
 				send_vlan(short_parsed);
 			} else if (is_word(q, "/counters.json")) {
-				send_counters(q[19]-'0');
+				send_counters(q[20]-'0');
 			} else if (is_word(q, "/eee.json")) {
 				send_eee();
 			} else if (is_word(q, "/mirror.json")) {
--- a/httpd/page_impl.c
+++ b/httpd/page_impl.c
@@ -143,6 +143,16 @@ void reg_to_html(register uint16_t reg)
 }
 
 
+void reg_to_html_long(register uint16_t reg)
+{
+	reg_read_m(reg);
+	byte_to_html(sfr_data[0]);
+	byte_to_html(sfr_data[1]);
+	byte_to_html(sfr_data[2]);
+	byte_to_html(sfr_data[3]);
+}
+
+
 void send_sfp_info(uint8_t sfp)
 {
 	// This loops over the Vendor-name, Vendor OUI, Vendor PN and Vendor rev ASCII fields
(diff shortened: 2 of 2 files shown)


vDorst:
Issues I found so far:
1. The stats window is too big, even on a QuadHD screen.
   Is it possible create a scrollbar?
2. On my device I need to click on port 5 `SHOW`-button to show port 6 stats.
   So it seems it doesn't make use of port-mapper-table.
3. `TX Good Counter` and `RX Good Counter` seems to be swapped.
4. I see a lot or `TX ...` and `RX ...` on both sides of the table.
   Can `TX`-values on the left-side and `RX`-values on the right?

Screenshot_20260112_195014

logicog:
Thanks for testing this! Concerning your points:
1. I'll try to see how to add a scroll-bar.
2. This is strange, for me it was working correctly. Will need to check.
3. I believe they are correct if TX packets > RX packets. In general, the switch receives many more packets than it sends if it is directly connected to a linux machine, as it does not react to IPv6 nor some IPv4 protocols.
4. I thought about this, but this is difficult to do, a lot of work to fix, plus not very future friendly: there are some counters which do not have a counterpart, such as the Egress queues (only TX), or errors that can only happen for received packets such as the length (only RX). Basically, one would need to say for every field whether it should go into the left or the right column (left is e.g. RX, right TX), and there would be empty fields if there is no counterpart, making the list even longer. To achieve this, one would need to have a complete "grid" of fields in the table and place every counter/description pair into one of the x/y positions in the grid. For this, everything would need to be hand-positioned, which means not doing this algorithmically, but by putting the complete layout into .html and reference it in the .js. In short: lots of layout, little code and considerably larger files. I would prefer to leave this like it is for now, until we have a more final layout.

logicog:
1. I added a scrollbar and rebased.
2. I verified this, and it is working correctly for me. I can see in the debug console that if I press show on port 1, then I see a `GET http://192.168.10.247/counters.json?port=0` In httpd_appcall, this is decoded to port = 0, which is used as a parameter to `send_counters(port=0)' . This function does:
```
        uint8_t i = machine.phys_to_log_port[port]; << physical starting with 0 used, according to machine.h
	slen += strtox(outbuf + slen, "[");
	for (uint8_t counter = 0; counter < 0x37; counter++) {
		STAT_GET(counter, i);  <<< logical port, is 4 according to machine.c
```
I repeated this for other ports and the results are consistent.
3. For me the numbers between the statistics table and the popup are consistent regarding RX/TX of total packets. And comparing with wireshark in the statistics menu I also see a consistent picture regarding packets sent/received between host and switch. Can you check again?
4. I would prefer not to fix. I played through multiple ways for presentation, and I do not see a way this can be done in a way that looks consistent without a lot of manual configuration of the tabular display.

vDorst:
1. Scrollbar is way nicer now, but now I also have a horizontal bar. Both in Firefox and Brave. Can this be changes easily?
2. I think I found the problem. `.phys_to_log_port = {4, 5, 6, 7, 3, 8, 0, 0, 0},` is also wrong on my device.
On my device, SFP ports are swapped. So I think it should be`.phys_to_log_port = {4, 5, 6, 7, 8, 3, 0, 0, 0},`.
3. I am still thinking the counter are wrong.
 My expectation is that the sum of all the `TX Ether ... Octect`-counter should equal to the `TX  Ether Flexible Octets CRC Set 0` or `TX Ether Flexible Octets CRC Set 1`. Same for the RX side. But is seems the opposite. See picture below
 `mib_counters` in `stat.js`, the TX/RX-counters are interleaved, so just missing/adding a counter string could shift the values.
4. That is fine.

 Screenshot_20260113_214049

logicog:
I found the issue with RX and TX being swapped for the "TX/RX Flexible Octet Sets" (BTW: I have no idea what this means). The problem is that one 64 bit counter sometimes contains 2 32bit counters, in particular for packets, whereas the 64 bit counters usually are for octets. There was a bug in the alignment of the descriptions with the counters because sometimes 32 bit counters are not used and only the other half of the 64 bit counter contains data.
I updated the PR.

Regarding:
1. I assume you will make a separate PR?
2. This is unrelated to this PR: The timer itself has stopped, read_reg_timer() always return 0. This seems to have to do with either timer/counter changes or the fact that this register always reads 0 (this happened in the past due to SDS changes(!)), but I cannot track it down to a particular commit. Overall, this means that the web-site currently never times out. Maybe investigate with the I2C slave interface?
3. The "tcp: found no unused connections." comes from uIP. This warning is printed when the user clicks around in the web-page and 2 HTTP requests cross each other and http.c can only answer the already ongoing request, as there can only be one request at a time due to limitations of the TCP stack. So this warning is highly dependent on when and where you follow links in the web-interface. It does not shop up for me when the statistics page just sits there.

vDorst:
> I assume you will make a separate PR? 

I'll do that, I was looking for conformation if it was right.

PR #83, opened 2026-01-16
Add support for L2 entry display and deletion in the web interface

logicog:
l2_screenshot

This PR adds support for displaying the currently learned or static L2 forwarding entries for Unicast traffic in the forwarding table of the RTL-SoC.

The forwarding table of the SoC contains up to 4096 entries which map a combination of MAC-address and VLAN-ID via a hash function to ports to which an Ethernet frame is sent to. The PR adds backend code in the simulator as well as in httpd to send these L2 forwarding entries to the web-frontend. The web-frontend will therefore provide an overview of the MAC addresses of the devices on a LAN and their location behind a particular Ethernet-interface of the switch. The code supports chunking these requests in order to limit the size of the returned packets as these could become very large if all entries were to be returned in one go.

The PR also provides a "Delete" button in the web-frontend as well as the relevant backend support to delete an L2 entry. Note that this is either only temporary, until the next packet is received from that MAC address (at which point the entry is again learned), or to prematurely remove an L2 entry for a device that is no longer part of the LAN, but for which an entry is still present. The switch will, however, also delete all entries for a port for which the cable has been unplugged or the link is otherwise lost. L2 entries are therefore not configuration options but dynamic and the deletion is not done through the firmware's command interface, but is merely a "GET" request that is not persistent.

### DIFF
--- a/rtl837x_port.c
+++ b/rtl837x_port.c
@@ -263,14 +263,23 @@ void port_l2_learned(void) __banked
 	__xdata uint16_t first_entry = 0xffff; // Table does not have that many entries
 
 	while (1) {
-		uint8_t port = 0, other = 0;
+		uint8_t port = 0;
 		reg_read_m(RTL837x_TBL_DATA_0);
 		REG_WRITE(RTL837x_TBL_DATA_0, sfr_data[0], sfr_data[1],sfr_data[2] | 0xc0, sfr_data[3]);
 
-		REG_WRITE(RTL837X_TBL_CTRL, entry >> 8, entry, TBL_L2_UNICAST, 0x1);
+		REG_WRITE(RTL837X_TBL_CTRL, (entry >> 8) & 0xf, entry, TBL_L2_UNICAST, TBL_EXECUTE);
 		do {
 			reg_read_m(RTL837X_TBL_CTRL);
-		} while (sfr_data[3] & 0x1);
+		} while (sfr_data[3] & TBL_EXECUTE);
+
+		reg_read_m(RTL837x_TBL_DATA_0);
+		entry = (((uint16_t)sfr_data[2] & 0x0f) << 8) | sfr_data[3];
+		if (first_entry == 0xffff) {
+			first_entry = entry;
+		} else {
+			if (first_entry == entry)
+				break;
+		}
 
 		// MAC
 		reg_read_m(RTL837x_L2_DATA_OUT_B);
@@ -278,7 +287,6 @@ void port_l2_learned(void) __banked
 			print_byte(sfr_data[2]); write_char(':');
 			print_byte(sfr_data[3]); write_char(':');
 			port = (sfr_data[0] >> 6) & 0x3;
-			other = sfr_data[0];
 			reg_read_m(RTL837x_L2_DATA_OUT_A);
 			print_byte(sfr_data[0]); write_char(':');
 			print_byte(sfr_data[1]); write_char(':');
@@ -298,22 +306,12 @@ void port_l2_learned(void) __banked
 
 			port |= (sfr_data[3] & 0x3) << 2;
 			if (port < 9)
-				write_char('1' + port);
+				write_char(machine.log_to_phys_port[port] + '0');
 			else
-				print_string("10");
+				print_string("CPU");
 		}
-		reg_read_m(RTL837x_TBL_DATA_0);
-		entry = (((uint16_t)sfr_data[2] & 0x0f) << 8) | sfr_data[3] + 1;
-		if (first_entry == 0xffff) {
-			first_entry = entry;
-		} else {
-			if (first_entry == entry)
-				break;
-		}
-#ifdef DEBUG
-		write_char(' '); print_sfr_data();
-		write_char(' '); print_byte(other);
-#endif
+
+		entry++;
 		print_string("\n");
 	}
 }


httpd/page_impl.c
```diff
@@ -283,6 +285,149 @@ void send_counters(char port)
...
+			byte_to_html(sfr_data[1]);
+
+			// type
+			reg_read_m(RTL837x_L2_DATA_OUT_C);
+			if (sfr_data[2] & 0x1)
+				slen += strtox(outbuf + slen, "\",\"type\":\"s\", \"port\":");
+			else
+				slen += strtox(outbuf + slen, "\",\"type\":\"l\", \"port\":");
+
+			port |= (sfr_data[3] & 0x3) << 2;
```
vDorst: I think this is always zero. Because port value 0-9 so only 4-bit.
Maybe just comment is out and add a comment that it always zero on this SOC.

httpd/page_impl.c
```diff
@@ -283,6 +285,149 @@ void send_counters(char port)
...
+			byte_to_html(sfr_data[3]); char_to_html(':');
+			port = (sfr_data[0] >> 6) & 0x3;
+			reg_read_m(RTL837x_L2_DATA_OUT_A);
+			byte_to_html(sfr_data[0]); char_to_html(':');
+			byte_to_html(sfr_data[1]); char_to_html(':');
+			byte_to_html(sfr_data[2]); char_to_html(':');
+			byte_to_html(sfr_data[3]);
+
+			// VLAN
+			slen += strtox(outbuf + slen, "\", \"vlan\":\"");
```
vDorst: remove the space
~~~suggestion
			slen += strtox(outbuf + slen, "\",\"vlan\":\"");
~~~

httpd/page_impl.c
```diff
@@ -283,6 +285,149 @@ void send_counters(char port)
...
+	 * i.e. for a given hash value, 4 entries with that same hash can be stored
+	 * (i.e. the hash points to a bucket with up to 4 entries).
+	 * When the table or a hash bucket is full, further entries will lead to
+	 * L2 flooding.
+	 * To find all entries, we start with entry-index 0 and iteratively search for
+	 * the next entry (with the next higher index), until we arrive again at the first
+	 * entry. The indices are sorted, so if an entry has a smaller index than
+	 * the previous one, we know that we have wrapped around the entire table.
+	 */
+	__xdata uint16_t entry = idx;
```
vDorst: Add bound check on the idx? 0..0x1000 is value.
Return error if invalid?

httpd/page_impl.c
vDorst, line 351: This can also be written as:

LINE 331:
```c
port = sfr_data[0] & 0xB0;
```

~~~suggestion
port |= (sfr_data[3] & 0x3);
port <<= 2;
~~~
So you need to shift only once.


//  sfr_data[0] 
port = 0b2100_0000
// sfr_data[3] 
port |= 0b000_0043 => 0b2100_0043
port <<= 2;
port = 0b0000_4321

httpd/page_impl.c
```diff
@@ -283,6 +285,149 @@ void send_counters(char port)
...
+				break;
+			} else {
+				char_to_html(',');
+			}
+		}
+	}
+}
+
+
+void l2_delete(uint16_t idx)
```
vDorst, line 382: Add boundcheck on idx.

httpd/page_impl.c
```diff
@@ -283,6 +285,149 @@ void send_counters(char port)
...
+			reg_read_m(RTL837x_L2_DATA_OUT_B);
+			charhex_to_html(sfr_data[0] & 0x0f);
+			byte_to_html(sfr_data[1]);
+
+			// type
+			reg_read_m(RTL837x_L2_DATA_OUT_C);
+			if (sfr_data[2] & 0x1)
+				slen += strtox(outbuf + slen, "\",\"type\":\"s\", \"port\":");
+			else
+				slen += strtox(outbuf + slen, "\",\"type\":\"l\", \"port\":");
```
vDorst: Remove the spaces in the output.
~~~suggestion
				slen += strtox(outbuf + slen, "\",\"type\":\"s\",\"port\":");
			else
				slen += strtox(outbuf + slen, "\",\"type\":\"l\",\"port\":");
~~~

httpd/page_impl.c
logicog: Will fix.

httpd/page_impl.c
logicog: I suggest to just do just `entry = idx & 0xfff`. This makes sure that the search in the hash-table will work. The function is supposed to return the "next" table entries. Because the hash algorithm is opaque, this can be anything and the table anyway wraps around, there is no beginning and end. Whatever the first index, the entries returned would be the "first" ones and then the next batch of entries will then be retrieved by the client based on the index of the last entry in the returned batch.

httpd/page_impl.c
logicog: Are you sure? Admittedly, this gives me a headache, but I believe in line 331 it would need to be `& 0xc0` in order to get bits 6 and 7. But these need to becom bits 0 and 1 of the final port number, and if the last that is done is `port <<= 2` these bits will be always 0 in the resulting port number.

vDorst: I was not sure. 
```c
port = data[0] & 0xc0;
port |= data[1] & 0x03;
port = (port << 2) | (port >> 6);
```
This works in my pc test.
Annoying that you have to write it so verbose in C, just to get `RL`-instruction.
In Rust it is just `port = port.rotate_left(2);`.

httpd/page_impl.c
logicog: I suggest I'll again just make it fit into the range, whether the delete then "hits" an entry is anyway returned in the result variable. So the change would be:
```
REG_WRITE(RTL837X_TBL_CTRL, (idx >> 8) & 0xf, idx, TBL_L2_UNICAST, TBL_EXECUTE);
```
with the execute bit also being de-magiced. I will also see whether 
```
REG_WRITE(RTL837x_TBL_DATA_0, sfr_data[0], sfr_data[1],sfr_data[2] | 0xc0, sfr_data[3]);
```
can be better de-magiced, but at present I am a bit puzzled, because both RTL837X_TBL_CTRL and RTL837x_TBL_DATA_0 have a table type (as in ACL, vs IGMP vs L2-UC/MC tables etc.) field and the GPL dump actively avoids to define table types in half of the cases...

httpd/page_impl.c
logicog: That is because in Rust you use rotate, which rotates bits 7 and 8 back into bits 0 and 1, but in C there is no explicit left-shift/rotation behavior, it just uses the Assembler opcode that it finds for the architecture, and for the 8051 that is a shift, so bits 7 and 8 are just shifted out. Rust would probably implement the rotate using the carry bit on an 8051. Time to write a Rust compiler for 8051??? :smiley:

httpd/page_impl.c
logicog: Will do.

logicog:
> I see in Firefox that the site makes about 10 request/sec. 1 req/sec should be enough.
OOps, that was for testing only: I was using very small batches of entries to be returned for testing and that is why I set the timeout to 0.3 seconds, which makes nearly 4 requests per second.
Will change.

> 
> Console output shows this. But the port doesn't match the real ports. Also it shows double entries. In my case the first and last are the same.
> 
> ```
>         MAC             VLAN    type    port
> <Snip Macaddr  >        0x0001  learned 9
> <Snip Macaddr  >        0x0001  learned 5
> 1c:2a:a3:23:00:02       0x0001  learned 10
> <Snip Macaddr  >        0x0001  learned 9
> ```
That code was not actually touched by this PR, but I know now more about how the reading of the table works and will fix the output.

httpd/page_impl.c
vDorst: Before
```asm
2005 ;	page_impl.c:331: port = (sfr_data[0] >> 6) & 0x3;
      000963 90r00r00         [24] 2006 	mov	dptr,#_sfr_data
      000966 E0               [24] 2007 	movx	a,@dptr
      000967 23               [12] 2008 	rl	a
      000968 23               [12] 2009 	rl	a
      000969 54 03            [12] 2010 	anl	a,#0x03
      00096B FE               [12] 2011 	mov	r6,a
      00096C 53 06 03         [24] 2012 	anl	ar6,#0x03

  2181 ;   page_impl.c:351: port |= (sfr_data[3] & 0x3) << 2;                                                      
  2182        000A85 90r00r03         [24] 2182     mov dptr,#(_sfr_data + 0x0003)                                                                          
  2183        000A88 E0               [24] 2183     movx    a,@dptr                                                                                         
  2184        000A89 54 03            [12] 2184     anl a,#0x03                                                                                             
  2185        000A8B 25 E0            [12] 2185     add a,acc                                                                                               
  2186        000A8D 25 E0            [12] 2186     add a,acc                                                                                               
  2187        000A8F 4E               [12] 2187     orl a,r6                                                                                                
  2188        000A90 F5 82            [12] 2188     mov dpl,a   
```
After
```asm
  2005 ;   page_impl.c:331: port = sfr_data[0] & 0xc0;                                                             
  2006        000963 90r00r00         [24] 2006     mov dptr,#_sfr_data                                                                                     
  2007        000966 E0               [24] 2007     movx    a,@dptr                                                                                         
  2008        000967 FE               [12] 2008     mov r6,a                                                                                                
  2009        000968 53 06 C0         [24] 2009     anl ar6,#0xc0                                         
  
      000A81 90r00r03         [24] 2179     mov dptr,#(_sfr_data + 0x0003)                                                                          
  2180        000A84 E0               [24] 2180     movx    a,@dptr                                                                                         
  2181        000A85 FD               [12] 2181     mov r5,a                                                                                                
  2182        000A86 74 03            [12] 2182     mov a,#0x03                                                                                             
  2183        000A88 5D               [12] 2183     anl a,r5                                                                                                
  2184        000A89 42 06            [12] 2184     orl ar6,a                                                                                               
  2185                                     2185 ;   page_impl.c:352: port = (port >> 6) | (port << 2);                                                      
  2186        000A8B EE               [12] 2186     mov a,r6                                                                                                
  2187        000A8C 23               [12] 2187     rl  a                                                                                                   
  2188        000A8D 23               [12] 2188     rl  a                                                                                                   
  2189        000A8E F5 82            [12] 2189     mov dpl,a                                                                                               
  2190                                     2190 ;   page_impl.c:354: itoa_html(port); 
``` 

`port = (port >> 6) | (port << 2);` will result in two `rl`-instructions, what I wanted.
But the compiler is messing up the loading of the value to `AND`.  
instead of using `anl a, #0x03` so it is not wasting instructions and registers.

vDorst:
Thanks!

logicog:
> Edit: Crap, it was my programmer. Not the best day.

I just wanted to reply that the patch actually does not change anything in the L2 configuration, it only adds the possibility to look into it, and that code is only called when you use the web-page or the l2 CLI command :smile:

But one hint when the web-interface crashes (like infinite loops, strange pages) is, that one needs to do `make clean` and then `make`, because the build system is not able to pick up changes in the javascript and html files.

diijkstra:
> But one hint when the web-interface crashes (like infinite loops, strange pages) is, that one needs to do `make clean` and then `make`, because the build system is not able to pick up changes in the javascript and html files.

I know, I know ;) this time it was loose cable on clamp. On my defense, even git bisect was confirming js code is culprit!

PR #87, opened 2026-01-19
Move SW version print after flash initialization

feelfree69:
The code of print_sw_version() is in BANK1 and needs a proper configured flash controller.

### DIFF
--- a/rtlplayground.c
+++ b/rtlplayground.c
@@ -1791,6 +1791,11 @@ void bootloader(void)
 
 	EA = 1; // Enable global interrupt
 
+	// Flash controller should be initialized before any code in other banks is being fetched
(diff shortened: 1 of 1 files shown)


vDorst:
Duo lack of documentation of the internals of this device, I can´t tell if this a really needed or not. Also logicog and me didn´t hit this issue on multiple devices and you may have faulty flash or hardware.

But I think it can´t hurt to do it anyway.

But I think it is better to initialize the flash earlier. So moving `flash_init()` up in the code right below `EA = 1; // Enable global interrupt`.

And add a comment about this issue. That this is a precaution because we are unsure if this is really a problem that using the flash above HOME-code-segment can't be use before initializing the flash.

May make this one comment.
```c
print_string("\nStarting up...\n");
print_string("  Flash controller\n");
```

PR #88, opened 2026-01-21
Add Hisource HI-K0402WS pictures

TylerDurden-23:


TylerDurden-23:
Removed.
I will try to add the documentation - I try to get at least the SFP ports working first.

config.txt
```diff
@@ -1,3 +1,3 @@
-ip 192.168.10.247
-gw 192.168.10.1
+ip 192.168.4.247
+gw 192.168.4.10
```
logicog: Please take this out again, as it is not specific to this device.

debug.h
```diff
@@ -1,6 +1,9 @@
+#define DEBUG
```
logicog: Please take this out again, as it is not specific to this device.

machine.h
```diff
@@ -6,14 +6,15 @@
 /*
  * Select your machine type below
  */
-#define MACHINE_KP_9000_6XHML_X2
+// #define MACHINE_KP_9000_6XHML_X2
```
logicog: Could you change machine.h to merely add the define for the K0402WS, commented out? Otherwise it will be affecting the default. We will need a different solution than having the machine selected defined in machine.h, that is now clear. Possibly a define on the make command line. Maybe you have an idea?

PR #90, opened 2026-01-22
make fw upload working; nicer debug outputs for console

feelfree69:
Only functional change is to disable flash DIO operation by using `flash_init(0)`.
With DIO operation, I get stuck after first flashing the data from the first TCP packet (~1.5kB).
A similar problem has already been observed somebody else; note the comment in function`ìnstaller(void)`

I tried also switching dynamically to `flash_init(0)` before FW upload starts. This also worked out, but then I had troubles to get the operation back to flash_init(1) after all successful flash operations.

For me, the benefit of having an online FW update is greater than the disadvantage for slower flash access.

Other changes are only cosmetic for the serial console.

### DIFF
--- a/httpd/httpd.c
+++ b/httpd/httpd.c
@@ -436,6 +436,9 @@ void handle_post(void)
 			flash_region.addr = CONFIG_START;
 			flash_sector_erase();
 		}
+		flash_init(0); // Re-initialize flash for non-DIO operation, otherwise flashing fails
+		set_sys_led_state(SYS_LED_FAST);
+
 		crc_value = 0;
 		bindex = 0;
 		write_len = 0;
--- a/rtl837x_common.h
+++ b/rtl837x_common.h
@@ -124,5 +124,6 @@ void get_random_32(void);
 void read_reg_timer(uint32_t * tmr);
 void sfp_print_info(uint8_t sfp);
 bool gpio_pin_test(uint8_t pin);
+void set_sys_led_state(uint8_t state);
 
 #endif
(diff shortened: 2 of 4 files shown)


feelfree69:
Forgot to mention: Flashing the whole image as before would overwrite the config area @0x70000. That's why I hardcoded to flash only 0x700000 Bytes.

TylerDurden-23:
@feelfree69 I would ne nice if the http server would sent a 200 after the upload process, before doing the reset.
With firefox I have the problem, that the upload restarts after the switch resets creating an infinite loop if you do not terminate the upload manually.

feelfree69:
> @feelfree69 I would ne nice if the http server would sent a 200 after the upload process, before doing the reset. With firefox I have the problem, that the upload restarts after the switch resets creating an infinite loop if you do not terminate the upload manually.

Feel free to add this to the PR (or  make an additional PR).

logicog:
This is a really great cleanup of the Update code! Very nice and thorough work!

> I tried also switching dynamically to `flash_init(0)` before FW upload starts. This also worked out, but then I had troubles to get the operation back to flash_init(1) after all successful flash operations.

Disabling DIO entirely is a bit drastic, as it provides a 100% speed increase for the device executing commands in anything but BANK0, which is buffered in RAM by the RTL837x SoC. If you use an SPI sniffer on the access to the Flash-Chip, you will see that the SoC only does a prefetch of exactly 3 bytes (which is the longest an 8051 opcode can be) for every command executed from the Flash in BANKs 1, 2... . So it is not just access speed, it is also the speed with which the device runs, which translates into web-pages that are double as responsive.

Why do you not disable DIO before the flash upload and then just don't bother to switch it on again, as the device should anyway reboot? Would that work?

It would also be interesting to understand why the flash write operations fail when DIO is enabled, and why it is not possible to enable it, first on boot, then disable it for writes, and then enable it again. Do you know what hangs exactly among the flash functions?

PR #93, opened 2026-01-23
Add support for command line editing

logicog:
The current code for entering commands via the serial CLI is completely refactored and moved out of rtlplayground.c into a separate file cmd_editor.c putting code into BANK2.
The serial ring buffer sbuf[] is now exclusively used for receiving keys, including escape sequences (DEL will generate a 4-byte escape sequence). The command is now held in cmd_buffer. This reduces XMEM use, because the serial buffer can be much smaller.

Supported is only editing the current command line via backspace, del, cursor left and right.

Because the code cannot distinguish escape sequences which are not yet complete (because the serial isr has not yet put all characters into the serial buffer) from the dozens of unsupported ones (F-keys/home, Page up/down, cursor up/down...) they are ignored and remain in the serial buffer until the ring-pointer overwrites them. This is standard behaviour for consoles, which will print out garbage for unsupported characters. In this implementation, the command line buffer and the visible command line in the terminal are always synced, so garbage can be removed again by ediing the line.

The code makes several redundant checks in order to prevent race-conditions between the serial ISR and the comand-editing code.

### DIFF
--- a/cmd_editor.c
+++ b/cmd_editor.c
@@ -9,7 +9,12 @@ __xdata uint8_t l;
 // Properties of currently edited command line in cmd_buffer[CMD_BUF_SIZE]
 __xdata uint8_t cursor;
 __xdata uint8_t cmd_line_len;
+__xdata uint8_t current_cmdline[CMD_BUF_SIZE];
+__xdata uint16_t history_editptr;
 
+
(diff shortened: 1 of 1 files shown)


logicog:
Rebased to latest main.

PR #95, opened 2026-01-24
Add additional inits for n-devices.

TylerDurden-23:
Additional init statements for n-devices to enable SFP functionality.

### DIFF
--- a/rtlplayground.c
+++ b/rtlplayground.c
@@ -1416,6 +1416,9 @@ void led_config_9xh(void)
 	sfr_mask_data(0, 0xe0, 0xa0);
 	reg_write_m(RTL837X_REG_LED_MODE);
 
+	// Set LED blink rate to slow during booting
+	set_sys_led_state(SYS_LED_SLOW);
+
 	// Disable RLDP (Realtek Loop Detection Protocol) LEDs on loop detection
 	reg_read_m(RTL837X_REG_LED_RLDP_1);
 	sfr_mask_data(0, 0, 0x3);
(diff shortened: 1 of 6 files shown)


rtlplayground.c
```diff
@@ -1515,12 +1515,56 @@ void rtl8373_revision(void)
...
+	sds_read(0, 6, 2);
+	pval = SFR_DATA_U16;
+	sds_write_v(0, 6, 2, pval | 0x2000);
+
+	sds_read(1, 6, 2);
+	pval = SFR_DATA_U16;
+	sds_write_v(1, 6, 2, pval | 0x2000);
+
+	// FOR N-Version: #TX_POLARITY_SWAP
+	reg_read_m(0xa94);
```
logicog: Could you introduce a define for this register, something like RTL837X_REG_CFG_PHY_TX_POLARITY_SWAP to stay in the not-always-consistently-adhered-to nomenclature?
The init routines were all written before the GPL dump was made available, so the registers initially did not have names, unless their function was guessable.

rtlplayground.c
```diff
@@ -1515,12 +1515,56 @@ void rtl8373_revision(void)
...
+        sfr_data[3] = 0x6a;
+	reg_write_m(0xa94);
+}
+
+/* Check if this is an N-series device using the Chip ID */
+bool is_n_device(void)
+{
+	reg_read_m(RTL837X_REG_CHIP_INFO);
+	// Check for Chip ID ending in 0x7000 (0x83727000 or 0x83737000)
+	if ((sfr_data[2]) == 0x70 && (sfr_data[3]) == 0x00) {
```
logicog: It may be nice to add this also to the CPU-detection output at the start of the bootloader: when new devices are being added, it would be nice for us to see what type of CPU it is. There is a chance we never get to the init routine, learn that it is an N-Device, and then we cannot help the user.

Issue #86, opened 2026-01-19
HiSource K0402WS-V3.0


TylerDurden-23:
The K0402WS-V3.0 is a device with an RTL8371b according to the easily available serial output.

How it the chance that it will run with this firmware?

feelfree69:
Is this an unmanaged switch with a hardware switch for 2 different configurations (eg vlan seperation oder link aggregation).
Then I would guess that the hw switch just selects between the different flash chips.

TylerDurden-23:
Very preliminary: It seems to work. The LEDs are wrong of course ... but I can reach the Web interface, it is slow but working.
Data is being forwarded ... 

It even seems to consume less power - maybe because of EEE.

TylerDurden-23:
@logicog I'll have a look at it. Currently I have an issue with building though:

rtl837x_igmp.c:229: error 78: incompatible types
from type 'struct ipmc_table_entry xdata* xdata'
  to type 'unsigned-char xdata* fixed'
make: *** [Makefile:55: output/rtl837x_igmp.rel] Error 1

vDorst:
@TylerDurden-23 you have an old SDCC compiler. When using latest Ubuntu, they still use an old compiler version. See also #58.
You can use docker/podman to start a debian 13 container to build this firmware.

Machine: `MACHINE_KP_9000_6XH_X` is fine.

When the firmware is up and running, can you run `regget 4`, this registers shows the CPU type. Output looks like this: `REGGET: 0x0004: VAL: 0x83720000`

TylerDurden-23:
@feelfree69 I checked it - and yes, it looks like the switch toggles the flashes. Very practical for testing/development. :)

TylerDurden-23:
@vDorst With machine definition MACHINE_KP_9000_6XH_X definition the switch was not booting - MACHINE_SWGT024_V2_0 works.

It looks like web flashing does not work - maybe because it is only a 2 MByte flash chip and not the usual 4MB. 
I'll have to check more thoroughly to be sure.

TylerDurden-23:
@vDorst  Looks like it ...

![Image](https://github.com/user-attachments/assets/e7e27081-a3ba-4386-8b7e-26cef90f17b5)

![Image](https://github.com/user-attachments/assets/827c26f2-48d7-4cf4-b5fe-4538183e0b93)

logicog:
[...]

PR #44, opened 2025-12-08
Led config for Horaco ZX-SG4T2 and Optfocus OFS-XF2GT4

daniel-dona:
Register values taken from the original firmware, tested on the hardware and work as expected. 

What's the plan for managing different hardware versions? Some board definition?

PR #97, opened 2026-01-24
Add support for HiSource K0402WS-V3.0.

TylerDurden-23:
As discussed in #86.

The magic number LED approach is taken from #44  and can be reused there.

### DIFF
--- a/doc/devices/HI-K0402WS.md
+++ b/doc/devices/HI-K0402WS.md
@@ -1,6 +1,6 @@
-# Hisource Hi-KO4022WS
+# Hisource Hi-K0402WS
 
-Following is documentation for unmanaged switch marked as `Hi-KO4022WS`.
+Following is documentation for unmanaged switch marked as `Hi-K0402WS`.
 
 Original software is running UART on 9600 baud rate. 
 


logicog:
> @logicog A "combined" approach seems to work - I can strip down the magic number function to:
> 
> ```
> 
> ```

There is now a custom machine init function in the General LED setup PR. If you still need that, it should get you there.

machine.c
```diff
@@ -231,4 +232,53 @@ __code const struct machine machine = {
...
+#elif defined MACHINE_HI_K0402WS
+__code const struct machine machine = {
+	.machine_name = "HiSource HI-K0402WS",
+	.isRTL8373 = 0,
+	.min_port = 3,
+	.max_port = 8,
+	.n_sfp = 2,
+	.log_to_phys_port = {0, 0, 0, 6, 1, 2, 3, 4, 5},
+	.phys_to_log_port = {4, 5, 6, 7, 8, 3, 0, 0, 0},
+	.is_sfp = {0, 0, 0, 2, 0, 0, 0, 0, 1},
```
vDorst: 2 is strange. This should only mark if the port is an SFP port or not.
So only '0' or '1'.

logicog:
I just see 
```
	.is_sfp = {0, 0, 0, 1, 0, 0, 0, 0, 1},
```
this not correct, but I guess I prompted that due to me incorrectly naming that variable. The numbers may be 0, 1, 2, where 0 is not an SFP, 1 is the left SFP port and 2 is the right port. This associates a port with the respective SFP cage.

logicog:
Sorry, I did not pay enough attention, there was a misunderstanding about the role of the variable. See also PR #122 . I don not think it is possible to just remove a commit. And probably it would not be good practice, as it is no longer traceable. Can you make a PR for this? Sorry for the additional work :-(

PR #98, opened 2026-01-25
Fix soc init

vDorst:
Fix the N-chips serdes swap issues.
Detect and init the chip based on detected type instead of machine profile.

Added helper functions to read/write to/from rtl8224.
`rtl8224_write_reg_u16()` and `rtl8224_read_reg_u16` copy of `phy_read()`/`phy_write()`.
Without the copy, the compiler error-out because we are out-of-ram.
Now I don't use any additional ram.

Todos
- [ ] 1G-bit link not working on SDS 1 on a RTL8372N, see https://github.com/logicog/RTLPlayground/issues/86#issuecomment-3796988335
- [x] Demagic some more register values

### DIFF
--- a/rtl837x_phy.c
+++ b/rtl837x_phy.c
@@ -106,7 +106,7 @@ void phy_config(uint8_t phy) __banked
 
 	//	p0307.003e:0000 P000008.0700003e:0001
 	// EEE avertisment 2 register MMMD 7.0x003e, set bit 0: 2.5G has EEE capability
-	phy_modify(phy, 0x7, 0x3e, 0x0000, 0x0001);
+	phy_modify(phy, PHY_MMD_AN, PHY_EEE_ADV2, 0x0000, 0x0001);
 	delay(20);
 
 	//	p031f.a442:043c P000008.1f00a442:0430


rtl837x_phy.c
```diff
@@ -8,21 +8,25 @@
 
 // Phy ID of the external RTL8224 PHY.
 #define RTL8224_PHY_ID 0x00
+#define RTL8224_DEV_ID 0x1e
```
logicog: In phy.h is already a definition:
```
#define PHY_SDS_CTRL	30
```
This is also the name of that page in the RTL8221 datasheet, the RTL8224 is the Quad version of that.

rtl837x_phy.c
```diff
@@ -8,21 +8,25 @@
...
 #pragma codeseg BANK2
 #pragma constseg BANK2
 
 extern __code uint16_t bit_mask[16];
-
+extern __code const struct machine machine;
+extern __xdata struct machine_runtime machine_detected;
 
 __code uint16_t rtl8224_ca[42] = {
+	// SDS_DATA, SDS_CMD
```
logicog, line 35: Can you maybe rename that array to reflect its purpose, something with SDS config, or so? `ca` was just something from the time where this was magic because of the lack of documentation.
I believe cb is no longer used. Could you remove that?

rtl837x_phy.c
```diff
@@ -43,6 +47,9 @@ __code uint16_t rtl8224_ca[42] = {
 	0xabb0, 0xcedc,
 	0x5078, 0xc90c,
 	0xc45c, 0xc18c,
+	// Note Swapping the RX for N-device here, don't work
```
logicog: Could you say more about how the values in the array work? You seem to understand what they mean. For me this is still just magic.

rtlplayground.c
```diff
@@ -1810,17 +1853,27 @@ void bootloader(void)
...
-			print_string("INCORRECT MACHINE!");
+		write_char('2');
+	}
+	// Detect non-N/N chip, 0xxxxx70xx
+	if (sfr_data[2] == 0x70) {
+		machine_detected.isN = 1;
+		write_char('N');
+	}
+	write_char('\n');
+	if (machine.isRTL8373 != machine_detected.isRTL8373) {
```
logicog, line 1873: Do we still need `machine.isRTL8373` if we anyway detect it? What about an RTL8371B?

rtl837x_phy.c
vDorst: I would like to rename it, because it is Vendor specific register.
In the linux kernel they call it https://elixir.bootlin.com/linux/v6.18.6/source/include/uapi/linux/mdio.h#L28

```c
#define MDIO_MMD_VEND1		30	/* Vendor specific 1 */
#define MDIO_MMD_VEND2		31	/* Vendor specific 2 */
```
We can use the same naming convention.

What do you think?

vDorst: I see that we are already using a simulair name
```
phy.h
14:#define PHY_MMD_PMAPMD       1
15:#define PHY_MMD_AN   7
17:#define PHY_MMD_CTRL 31
```

rtl837x_phy.c
vDorst: Just an array of SDS-settings (sds_data and SDS_CMD). SDS_CMD has already encoded the SDS_INDEX, page and reg.
I'll add some documentation. In the RTL-SDK they have multiple version of SDS-settings, depending on the chip-revision.
We could implement that too in a seperate PR.

rtl837x_phy.c
vDorst: Ok

rtlplayground.c
vDorst: This provides an extra check to see if you have at least the right profile select that can fit on the SOC.

rtl837x_phy.c
logicog: The kernel calls it `MDIO_MMD_VEND1` because it supports a dozen different PHY vendors who support MMD, so it is not clear what this page does for an arbitrary vendor. But the ones we will encounter will always be Realtek, they use always their own ecosystem, because they have their own variant of MDIO that supports up to 63 PHYs. The `PHY_MMD_CTRL` is based on what they call that page in their datasheets. What about `PHY_MMD_SDS` ?

rtlplayground.c
vDorst: But still how does the 8371 relates to the other chips?
Do we need add support for it of can it just use the RTL8373-settings?
Maybe we need at some extra documentation about this chip.

rtl837x_phy.c
vDorst: It is a vendor specific register so the vendor can do that he want, and also use a different layout for every chip.
`PHY_MMD_SDS` is valid for `rtl8221` but not for `rtl8224`.
Looking into `rtl8221`-datasheet and it doesn't have the same layout as `rtl8224`.

So I am think some other options
- `PHY_MMD31` How they also call these registers `MMD 31`
- `PHY_MMD31_VEND2` Make it more distinctive.
- `PHY_MMD_1E`, `PHY_MMD1E` In hex

Issue #76, opened 2026-01-10
TRENDnet TEG-S562 - Failing at I2C SPF


diijkstra:
Thank you for this glorious project! I'm trying to bring up the TRENDnet TEG-S562 4x2.5+2xSPF+ unmanaged switch. 2.5G side is working flawlessly, but I do struggle with SPF modules.

I managed to identify GPIO pins for module detection and LOS (see WIP [doc TEG-S562.md](https://github.com/diijkstra/RTLPlayground/blob/trendnet/doc/devices/TEG-S562.md#gpio)). I do see expected changes when inserting module and adding fiber cable. But... module identification does not work. Certainly I2C data is garbage:

```
> 
<new link: 2000000000, was 0000000000>

<MODULE INSERTED>  Slot: 1  Rate: a0  Encoding: a0  Module: ������������������������������������

sds_config sds: 00, mode: ff

RTL837X_REG_SDS_MODES: 0x000003ff
Error in SDS Mode

<MODULE INSERTED>  Slot: 2  Rate: a0  Encoding: a0  Module: ������������������������������������

sds_config sds: 01, mode: ff

RTL837X_REG_SDS_MODES: 0x000003ff
Error in SDS Mode
```
The same modules (10GTek 10GBase-SR) are working on stock firmware flawlessly, so this is some software misconfiguration.

I did play around with `SCL_PIN` and `SDA_PIN` as described in #35. Did a loop over SCL/SDA values from 0 to 7 trying to read any SPF registry successfully but always got one of `0x00`, `0xff`, `0xa0` in response. This happened for each registry I tried (mostly rate - 12 & module identification - 0, encoding - 11). My fork is based on latest master with DIO/debug changes. 

I'm kinda out of ideas what else could be wrong. Would appreciate any tips how to debug it further... Would tracing the paths from modules to SOC help in any way? Is there a chance that they may use other I2C bus on SOC?

I have access to serial, spi programmer, some dumb oscilloscope, and bunch of micro controllers if needed. Here is a copy of stock firmware in case that might help: [trendnet_teg_S562_stock.bin.gz](https://github.com/user-attachments/files/24541558/trendnet_teg_S562_stock.bin.gz).

Some peculiarities when bringing this toy up which probably are not related:

1. Stock serial was running on 57600 baud rate. Was very limited, allowed to change MAC and show IP (which was not responding).
[...]

PR #104, opened 2026-01-26
Advanced I2C SFP pins definition

diijkstra:
RTL has up to 3 SCL pins and up to 4 SDA pins. Lets allow configuring I2C with individual BUS numbers instead of 0/1 I2C. This enables support for devices where SCL line is not shared between SFP modules. MUX registry is now initialized depending on needed pin function.

Some SFP pins require special MUX settings, lets initialize those depending on SFP configuration. Also lets configure optional TX Disable pin, which right now is set to low at startup.

Minor change in rtl_port, `SFP OK` is replaced with `SFP_IN` depending on detection pin. 

This is ennoblement work to support `Trendnet TEG-S562`, from #76 Subsequent PR will handle LED configuration. 

Caveats:
 - We may still override MUX registry later, do we need some centralized PIN setup?
 - Not sure how to handle invalid bus definitions - right now they are skipped with log.
 - Without SFP is it fine to *not* initialize anything?
 - I've struggled with RAM usage for GPIO setup functions, couldn't add another parameter for intial
   output value. I'm not proud of this mess but ran out of ideas.
 - I2C setup function would be easier with some helpers for setting multiple bits on byte border…

**Requires testing on other devices with common SCL line**. Did not encounter any issues in short testing but would like to gather feedback during further testing.

### DIFF
--- a/machine.c
+++ b/machine.c
@@ -21,7 +21,7 @@ __code const struct machine machine = {
 	.sfp_port[1].pin_tx_disable = GPIO_NA,
 	.sfp_port[1].sds = 0,
 	.sfp_port[1].i2c = { .sda = GPIO39_I2C_SDA4, .scl = GPIO40_I2C_SCL3_MDC1 },
-	.reset_pin = 46,
+	.reset_pin = GPIO46_I2C_SCL0,
 };
 #elif defined MACHINE_KP_9000_6XH_X
 __code const struct machine machine = {
@@ -38,6 +38,7 @@ __code const struct machine machine = {
 	.sfp_port[0].pin_tx_disable = GPIO_NA,
 	.sfp_port[0].sds = 1,
 	.sfp_port[0].i2c = { .sda = GPIO39_I2C_SDA4, .scl = GPIO40_I2C_SCL3_MDC1 },
+	.reset_pin = GPIO_NA,
 };
 #elif defined MACHINE_KP_9000_9XH_X_EU
 __code const struct machine machine = {
@@ -54,6 +55,7 @@ __code const struct machine machine = {
 	.sfp_port[0].pin_tx_disable = GPIO_NA,
 	.sfp_port[0].sds = 1,
 	.sfp_port[0].i2c = { .sda = GPIO39_I2C_SDA4, .scl = GPIO40_I2C_SCL3_MDC1 },
+	.reset_pin = GPIO_NA,
 };
 
 #elif defined MACHINE_SWGT024_V2_0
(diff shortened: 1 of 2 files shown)


rtlplayground.c
```diff
@@ -1765,11 +1817,50 @@ void setup_i2c(void)
...
+	// HW Control register, enable I2C depending on PIN configuration
 	reg_read_m(RTL837X_PIN_MUX_1);
-	sfr_mask_data(3, 0x20, 0x00); // Clear bit 29
-	sfr_mask_data(0, 0x60, 0x40); // Set bits 5-6 to 0b10
-	reg_write_m(RTL837X_PIN_MUX_1);
+	for (uint8_t sfp = 0; sfp < machine.n_sfp; sfp++) {
+		const uint8_t scl_bus = machine.sfp_port[sfp].i2c_bus.scl;
+		const uint8_t sda_bus = machine.sfp_port[sfp].i2c_bus.sda;
+		print_string("Configuring I2C for SFP idx="); print_byte(sfp); print_string(" SCL="); print_byte(scl_bus); print_string(", SDA="); print_byte(sda_bus); write_char('\n');
+		if (scl_bus == 3) {
```
logicog: Maybe use a switch() statement, here?

rtlplayground.c
```diff
@@ -1765,11 +1817,50 @@ void setup_i2c(void)
...
+			sfr_mask_data(1, 0x18, 0x08);
+		} else if (scl_bus == 0) {
+			// Bit 7-8 0b01 -> SCL
+			sfr_mask_data(0, 0x80, 0x80);
+			sfr_mask_data(1, 0x01, 0x00);
+		} else {
+			print_string("Invalid SCL bus number: "); print_byte(scl_bus); write_char('\n');
+		}
+
+		if (sda_bus == 4) {
```
logicog: Maybe use a switch statement, here, too?

logicog:
> not sure why it refuses to use OSEG...

I think sdcc only does an analysis about variable lifetime within a function and function parameters live also outside of the function, so they all end up in DSEG, which is pretty nasty.

But you can use __xdata as a qualifier also for parameters, then they will be handed over as XMEM. They will still be statically allocated in XMEM and cannot be used for anything else, plus the function is no longer re-entrant (but most of the code isn't either), but at least there is lots of XMEM. So, if it makes the code considerably more readable, this is the way to go. 

BTW: The SDK which is compiled using Keil, uses basically only XMEM for call parameters, but is also able to re-use the memory for different functions.

logicog (approved):
Very nice code including taking all the specs apart, and an even further improvement on the previous version! Great work!

diijkstra:
You welcome, thanks for patience!

PR #112, opened 2026-02-04
Enable SFP GPIO pins setup

diijkstra:
Previous changes addded initialization rotine, which was never called.

Looks like, I've lost crucial change somewhere in re-base process... Had it on testing branch, but never added to PR.


logicog:
> Previous changes addded initialization rotine, which was never called.
> 
> Looks like, I've lost crucial change somewhere in re-base process... Had it on testing branch, but never added to PR.

That can happen when tackling git/code/github-PR all the same-time...

PR #107, opened 2026-02-01
Generic Port LED configuration based on machine-configuration

logicog:
This uses machine configuration parameters in `machine.c` to configure the port leds making use of individual LED settings in LED-ids and LED-sets as found in section 8.17 of the datasheet.

In order to implement the LED settings, we first add support for the HG0402XG V1.1, which is the first supported machine making use of 2 LED-sets. For the machine configuration to be meaningful, we base this PR is on top of PR #104, which should be merged, first.

This PR introduces 2 new functions `leds_dump()` and `leds_setup()`, the first gives a human-readable printout of the current port-led configuration, the second uses the definitions in the machine structure to set up the leds.

The LED configuration in the machine-structure works by assigning led-IDs, which are groups of LED-conditions as given in the datasheet to a LED-set and each LED-set can then be assigned to a port.

Additionally, LED-pins 27-29, which are not by default configured by the switch on reset, can be enabled/disabled and selected/deselected as LED-driving pins.

All current machine templates are updated with a LED-configuration section.

The rather complicated configuration of the LEDs in  `leds_setup()` making use of a pointer into code-memory and then dong LE/BE conversion of the 32-bit numbers is entirely due to avoiding DSEG usage. The initial and certainly nicer-to-read implementation using array access used 8(!) DSEG-ram bytes and would not fit.

PR #113, opened 2026-02-04
Partial TRENDNET TEG-S562 support

diijkstra:
Mostly working, lacking LED initialization.
LEDs will be added later once #107 is merged.

Based on #112.


logicog:
Nice work and thanks for investigating all the details!

PR #114, opened 2026-02-05
Add EEE speed option and enable EEE with 2g5 as default at boot

TylerDurden-23:
- Handle speed options for eee_enable function
- Command parser extended to handle speed option for EEE
- Activate EEE with 2g5 speed at boot (instead with default of only up to 1G)


TylerDurden-23:
@logicog Yes, it parameterizes eee_enable to not do a phy_reset, so that the same function can be used during the initialization&boot phase without wasting time to do another phy_reset, which will be done anyhow as part of the confiugration.

Using a third parameter would have been more elegant, but doing it that way was more memory conservative. And not doing a phy reset is a special case of the boot process.

PR #116, opened 2026-02-07
Fix Speed Setting using Auton-Neg

logicog:
Fix Auto-negotiation
Commit https://github.com/logicog/RTLPlayground/commit/7f4c276285bc31ff93a0c6f7c666c42b9cbcd41a used the wrong MMD device
for use with the auto-negotiation page:
```
- phy_write(port, PHY_MMD_AN, 0x00, 0x3200);	// Restart AN
+ phy_write(port, PHY_MMD31, PHY_ANEG_CTRL, 0x3200);	// Restart AN
```
Fix this. We also add more output as feedback and use __xdata to
reduce DSEG use for function parameters, which were using 3 bytes:
```
--------------------------------------------------------
; internal ram data
	.area DSEG    (DATA)
_phy_set_speed_PARM_2:
	.ds 1
_phy_set_speed_PARM_3:
	.ds 1
_phy_set_duplex_PARM_2:
	.ds 1
```
This also fixes the use of 5 bytes for parameters in DSEG in rtl838x_port.c and a bug in the SDS configuration of the MACHINE_KP_9000_6XHML_X2

### DIFF
--- a/rtl837x_phy.c
+++ b/rtl837x_phy.c
@@ -191,6 +191,43 @@ void phy_config_8224(void) __banked
 void phy_set_speed(void) __banked
 {
 	uint16_t v;
+
+	print_string("Setting port "); write_char(machine.log_to_phys_port[phy_settings.port] + '0');
+	if (phy_settings.speed == PHY_OFF) {
+		print_string(" to disabled");
+	} else {
+		print_string(" to speed ");
+		switch(phy_settings.speed) {
+		case PHY_SPEED_AUTO:
+			print_string("auto");
+			break;
+		case PHY_SPEED_10M:
+			print_string("10M");
+			if (phy_settings.duplex)
+				print_string(" full duplex");
+			else
+				print_string(" half duplex");
+			break;
+		case PHY_SPEED_100M:
+			print_string("100M");
+			if (phy_settings.duplex)
+				print_string(" full duplex");
+			else
+				print_string(" half duplex");
+			break;
+		case PHY_SPEED_1G:
+			print_string("1G");
+			break;
+		case PHY_SPEED_2G5:
+			print_string("2G5");
+			break;
+		default:
+			print_string("UNKNOWN");
+			break;
+		}
+	}
+	write_char('\n');
+
 	phy_read(phy_settings.port, PHY_MMD31, 0xa610);
 	v = SFR_DATA_U16;
 	if (phy_settings.speed == PHY_OFF) {
@@ -258,6 +295,14 @@ void phy_set_speed(void) __banked
 void phy_set_duplex(void) __banked
 {
 	uint16_t v;
+
+	print_string("Setting port "); write_char(machine.log_to_phys_port[phy_settings.port] + '0');
+	if (phy_settings.duplex)
+		print_string(" to full duplex");
+	else
+		print_string(" to half duplex");
+	write_char('\n');
+
 	phy_read(phy_settings.port, PHY_MMD_AN, PHY_ANEG_CTRL);
 	v = SFR_DATA_U16;	
 	if (!(v & 0x1000)) { // AN disabled, we are in forced mode


rtl837x_phy.c
```diff
@@ -186,9 +186,46 @@ void phy_config_8224(void) __banked
  * See e.g. RTL8221B datasheet
  * duplex: 0: half, 1: full, 2: both
  */
-void phy_set_speed(uint8_t port, uint8_t speed, uint8_t duplex) __banked
+void phy_set_speed(__xdata uint8_t port, __xdata uint8_t speed, __xdata uint8_t duplex) __banked
```
vDorst: Can we make this just a struct?
Like this, which takes only 16-bit instead of 24.
Also now it is just a single pointer instead of 3.
```c
struct phy_setting {
 uint8_t pad : 3;
 uint8_t duplex : 1;
 uint8_t port : 4; 
 uint8_t speed;
}
```

I think you can go a step further, like sfr_data.
Always use this data to pass around phy_settings.

rtl837x_port.c
```diff
@@ -123,7 +123,7 @@ __xdata uint16_t vlan_name(register uint16_t vlan) __banked
 /*
  * A member that is not tagged, is untagged
  */
-void vlan_create(register uint16_t vlan, register uint16_t members, register uint16_t tagged) __banked
+void vlan_create(__xdata uint16_t vlan, __xdata uint16_t members, __xdata uint16_t tagged) __banked
```
vDorst: Same here make a single struct. So only one pointer is passed around instead of three.

rtl837x_port.c
```diff
@@ -46,7 +46,7 @@ void port_mirror_del(void) __banked
 }
 
 
-void port_ingress_filter(register uint8_t port, uint8_t type) __banked
+void port_ingress_filter(__xdata uint8_t port, __xdata uint8_t type) __banked
```
vDorst, line 51: I think this makes no sense.
I think the compiler now makes it a two 8 bits values in to two 16-bit pointers.
But I can be wrong.

rtl837x_phy.c
```diff
@@ -208,10 +245,10 @@ void phy_set_speed(uint8_t port, uint8_t speed, uint8_t duplex) __banked
 			phy_write(port, PHY_MMD_AN, PHY_ANEG_MGBASE_CTRL, 0x6081);
 			// GBCR (1000Base-T Control Register, MMD 31.0xA412)
 			phy_modify(port, PHY_MMD31, PHY_MMD31_GBCR, 0x0000, 0x0200); // Loop timing enabled
-			phy_write(port, PHY_MMD31, PHY_ANEG_CTRL, 0x3200);	// Restart AN
+			phy_write(port, PHY_MMD_AN,  PHY_ANEG_CTRL, 0x3200);	// Restart AN
```
vDorst: good catch and I am happy that I make a commit for every change so it can be spotted more easily.

vDorst (commented):
I think we need to convert many arguments into a struct in XDATA memory.
Maybe see if it helps if the data always lives in XDATA like we use `sfr_data`.

rtl837x_phy.c
logicog: Very good idea. Will do.

rtl837x_port.c
logicog: Will do.

rtl837x_port.c
logicog: Will check. For me the most important bit is that no DSEG byte is used. Those are forever gone.

rtl837x_phy.c
logicog: I get the hint ;-) My problem was when I overlooked it, that I usually mostly look at the changed files overall. I also don't want to make it a science to craft the perfect individual commits.

rtl837x_port.c
logicog: No, __xdata variables are not passed by pointer. The compiler creates a global variable to pass the parameter in and then fills this variable in before calling. In the function, the compiler directly loads dptr with the right variable address to use in movx. The location of the variables is not passed as a parameter since the compiler knows where it allocated the variable.

vDorst:
Portmapping seems to go wrong.

On my device `SWGT024 V2.0`, I was just changing port `4` but that translate into 3rd port from the right.
```
Initializing Flash controller
Detecting CPU: RTL8372
Software version: v0.1.0-g4da92a3
Build: Feb  8 2026 12:16:23
Hardware: SWGT024 V2.0
NIC reset
```

```
> port 4 1g

PORT 06 1G
Setting port 3 to speed 1G

> 
<new link: 2200000000, was 2201000000>

<new link: 2202000000, was 2200000000>
stat

6       NO SFP  Down    0x00000000      0x00000000      0x00000013      0x00000000
1       On      Down    0x00000000      0x00000000      0x00000002      0x00000000
2       On      Down    0x00000000      0x00000000      0x00000004      0x00000000
3       On      1000M   0x00000000      0x00000000      0x00000004      0x00000000
4       On      Down    0x00000000      0x00000000      0x00000003      0x00000000
5       SFP IN  1000M   0x0000000d      0x00000000      0x00000020      0x00000000

> port 4 show

PORT 06
Link speed: 1000M full duplex
AN enabled, advertising: 1000Base-Full
Link Partner advertises: 10Base-Half 10Base-Full 100Base-Half 100Base-Full 1000Base-Full

> port 4 100m

PORT 06 100M
Setting port 3 to speed 100M half duplex

> 
<new link: 2200000000, was 2202000000>

<new link: 2201000000, was 2200000000>
stat

6       NO SFP  Down    0x00000000      0x00000000      0x00000013      0x00000000
1       On      Down    0x00000000      0x00000000      0x00000002      0x00000000
2       On      Down    0x00000000      0x00000000      0x00000004      0x00000000
3       On      100M    0x00000000      0x00000000      0x00000006      0x00000000
4       On      Down    0x00000000      0x00000000      0x00000003      0x00000000
5       SFP IN  1000M   0x0000000f      0x00000000      0x00000020      0x00000000
```

rtl837x_phy.c
vDorst: No no I don't want to go that route.
It is just I am happy that I did that well when refactor these registers.

rtl837x_port.c
vDorst: ok great.

feelfree69:
> On my device `SWGT024 V2.0`, I was just changing port `4` but that translate into 3rd port from the right.

I trapped into the same pitfall. In WebInterface Ports are numbered 1..8, in serial log 0..7

Edit: Notice to myself: Just read carefully, think, understand, *then* post. Forget about my comment....

logicog:
I pushed the fix.

Issue #115, opened 2026-02-06
Setting individualized MAC addresses


TylerDurden-23:
It seems that the MAC address for the management interface is assigned a static MAC address and once two or more switches with the firmware are on the same segment, there is a collision.

How is this supposed to work?

vDorst:
My idea was using siphash for hashing the MAC address. So it is random-stable by using the UUID + LOT number of the chip.
I was thinking about Chacha20+Poly1305 for TLS encryption because this is also suitable for smaller low-powered devices.

This may take a while to implement plus other stuff in the project is pushing further to the background.

Still making infrastructure to select MAC-address option would be great.

Some MAC-address options that I can think off that would be usefull.
- Manual
- Total random and then saved as "manual"
- Using the UUID and XOR all the bytes to a single 16-bits and use it partially which an other prefix. So it is random-and-stable
- Siphash to hash UUID + LOT for a random-and-stable address and then saved as "manual".

PR #117, opened 2026-02-07
Use a MAC address based on the SoC's UUID

logicog:
We use the SoC's UUID to generate a fixed MAC for a particular switch device. The MAC generated uses a Realtek prefix and then is followed by 3 bytes genertaed from the first 3 bytes of the UUID xored with the last byte in order to prevent being able to deduce the UUID from the public MAC.

Since the MAC is now in XMEM and no longer in CODE space, it is now also possible to change it using a mac-setting command or a different algorithm, which are not provided in the PR.

This intends to fix Issue #115


TylerDurden-23:
@logicog  Looks almost as my code :) - 
I would suggest using a locally administered MAC - then we are not intruding on the address space of somebody else:

```
void ethaddr_init(void)
{
	// Set MAC address of CPU port
	reg_read_m(RTL8373_CHIP_UUID_REG_ADDR);
	uip_ethaddr.addr[0] = 0x13; // Set Locally Administered bit, to avoid conflicts with real MAC addresses, and set bit 1 to indicate locally administered unicast address
	uip_ethaddr.addr[1] = 0xA2; // Random value
	uip_ethaddr.addr[2] = sfr_data[1] ^ sfr_data[0];
	uip_ethaddr.addr[3] = sfr_data[2] ^ sfr_data[0];
	uip_ethaddr.addr[4] = sfr_data[3] ^ sfr_data[0]; 
	uip_ethaddr.addr[5] = sfr_data[3] ^ sfr_data[0];
}
```

logicog:
> Looks almost as my code :) -

I knew I had to move quickly on this one ;-) But jokes aside: After the discussion on the related issue, it was clear that this could be done already now and should not be made too complicated. As it requires some changes in the uIP-code, I thought I would do that, as that stuff is quite particular.

> I would suggest using a locally administered MAC - then we are not intruding on the address space of somebody else:

I had thought about that. My argument to choose Realtek's (old) namespace is that if we use something private, but distinct, then the type of device is easily spotted, if the address gets public, everyone knows that it is one of these devices. Collisions should be extremely rare. There are 2^24 addresses we can use and chances that you have a collision with another Realtek device depends on how many you have in your network (most of the Realtek-based devices use the namespace of the vendor, not from Realtek). If you have 2 other devices from Realtek in your LAN, you end up with a chance of 1/2^47, which is pretty good.
Also, if we use Realtek, you will also be able to spot that device on your LAN, because if you see Realtek, you may think about a Realtek device. One of my RTL8372 switches actually has a Realtek MAC in the flash.

logicog:
Fixed as suggested. The MAC-address now starts with 0x06, which is a Unicast LAA, bytes 1 and 2 are xored from the LOT and bytes 3, 4, 5 are as before.
I tested with an RTL8372 and an RTL8373, and the LOT number does not change too much (the RTL8372 has 0x50504e54, the RTL8373 has 0x4b424e54), so it should be relatively easy to reconstruct from the MAC that it is an RTL837x plus probably the LOT number, but so would be setting it to a particular vendor-id.
The UUID seems to have some randomness, however.

vDorst:
This really looks great.
Good to see that we have a way out not using someone else there MAC-vendor-id.

Only nit, I don't like that the UUID and LOT id is printed in the log because people will that these logs in a issue and these are unique identifies.

PR #119, opened 2026-02-11
Add MACHINE_SWTGW218AS Managed Switch

feelfree69:
This PR adds support for the Managed Switch with Label SWTGW218AS. Sold under Horaco and Mokerlink label.

PCB itself is labeled SWTG118AS-V2.0, see [here](https://github.com/up-n-atom/SWTG118AS/blob/main/photos/SWTG118AS-v2.0/SWTG118AS-v2.0-pcb-top.JPG).

LED config is done as OEM firmware does and fronplate suggests; 
- Right green LED for 10/100/1000M Link/Act, 
- Left amber LED for 2G5 Link/Act.

Additional changes:
- do not set a default machine in machine.h, but throw an error when anybody forgets to change the value to the own machine.
- Set the SYS_LED to slow blinking directly after the LED settings have been configured


machine.c
```diff
@@ -179,6 +179,40 @@ __code const struct machine machine = {
...
+	.sfp_port[0].i2c = { .sda = GPIO39_I2C_SDA4, .scl = GPIO40_I2C_SCL3_MDC1 },
+	.reset_pin = GPIO_NA,
+	.high_leds = { .mux = LED_27 | LED_28 | LED_29, .enable = LED_28 | LED_29 },
+	.port_led_set = { 0, 0, 0, 0, 0, 0, 0, 0, 1},
+	.led_sets = { { LEDS_2G5 | LEDS_LINK | LEDS_ACT, // Green LED (right)
+					0, // unused
+					LEDS_1G | LEDS_100M | LEDS_10M | LEDS_LINK | LEDS_ACT, // Amber LED (left)
+					0
+				  }, // unused
+				  { LEDS_10G | LEDS_LINK | LEDS_ACT, // SFP LED
```
vDorst: Does this led still lid-up on a 1G and 2.5G link?

vDorst (commented):
I am also in favor of choosing the machine type, can this also be selected via a `make` argument?
I personally prefer that this is a separate PR.

Can you also make documentation about your device, like a PCB photo and a simple markdown file?

machine.c
feelfree69: Actually, I can't test it. I thought the SFP port is 10G only. 
Adding other speeds shouldn't hurt.

vDorst: I am going to test it out. I wonder what will happen.

logicog: > Actually, I can't test it. I thought the SFP port is 10G only.

The port supports anything between 1000MBit, 2.5GBit, 5GBit and 10GBit. This is a feature of the Chip, not the SFP port. I believe even 100MBit are supported as part of a mode where 10 clock cycles are used to transfer 1 bit to the module and the clock is actually 1GBit (sorry, to be exact 1.25GBit, as there is always 25% overhead for the error correction).

vDorst: I tested with `LEDS_10G` removed on my device.
The SFP led does nothing anymore on a 10GBit link, 1GBit is works fine, so speed flag is important.

feelfree69:
> For that we could put
> 
> ```
> IMAGESIZE = 524288
> CONFIG_LOCATION = 458752
> HTML_LOCATION = 262144
> ```
> 
> into that file

I know it's unrelated to this PR, but I ask anyhow: Why do we build by default 1MB images? The above settings would also work for 1MB flash size, including updates. And I do not see we're running out of flash space....

feelfree69:
Glad to hear that it works for you too!

My switch with OEM Firmware was also dead some day. The problem ist the flash access, which uses a.mode which is hard at the limit, and a lot of switches die after one or two years.
This firmware uses a flash access mode which has more margin.

vDorst:
Thanks!

PR #122, opened 2026-02-13
Fix MACHINE_KP_9000_6XHML_X2: swap SFP-ports remove unnecessary LED-MUX

logicog:
The left and right SFP ports were swapped in the machine structure. They are now correctly identified a comment is added. The custo LED-mux is removed again, as this was accidentally added for demonstration purposes. It is not necessary, as the device uses the default.

### DIFF
--- a/machine.c
+++ b/machine.c
@@ -13,12 +13,14 @@ __code const struct machine machine = {
 	.n_sfp = 2,
 	.log_to_phys_port = {0, 0, 0, 5, 1, 2, 3, 4, 6},
 	.phys_to_log_port = {4, 5, 6, 7, 3, 8, 0, 0, 0},
-	.is_sfp = {0, 0, 0, 2, 0, 0, 0, 0, 1},
+	.is_sfp = {0, 0, 0, 1, 0, 0, 0, 0, 2},
+	// Left SFP port (5)
 	.sfp_port[0].pin_detect = GPIO50_I2C_SCL2_UART1_TX,
 	.sfp_port[0].pin_los = GPIO10_LED10,
 	.sfp_port[0].pin_tx_disable = GPIO_NA,
 	.sfp_port[0].sds = 0,
 	.sfp_port[0].i2c = { .sda = GPIO41_I2C_SDA3_MDIO1, .scl = GPIO40_I2C_SCL3_MDC1 },
+	// Right SFP port (6)
 	.sfp_port[1].pin_detect = GPIO30_ACL_BIT3_EN,
 	.sfp_port[1].pin_los = GPIO37,
 	.sfp_port[1].pin_tx_disable = GPIO_NA,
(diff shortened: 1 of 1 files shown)


machine.c
```diff
@@ -13,12 +13,14 @@ __code const struct machine machine = {
 	.n_sfp = 2,
 	.log_to_phys_port = {0, 0, 0, 5, 1, 2, 3, 4, 6},
 	.phys_to_log_port = {4, 5, 6, 7, 3, 8, 0, 0, 0},
-	.is_sfp = {0, 0, 0, 2, 0, 0, 0, 0, 1},
+	.is_sfp = {0, 0, 0, 1, 0, 0, 0, 0, 2},
```
vDorst, line 16: I am confused about `is_sfp` variable.
Based on the name it should be a array of bools.

`machine.c` tells me it also defines the I2C port, which is not the case anymore.
```
machine.h
56:     uint8_t is_sfp[9];  // 0 for non-SFP ports 1 or 2 for the I2C port number
```

So what should it represent? Other than the MAC/port is connected to an SFP-CAGE.

logicog: Well, that was the initial problem. The variable is more than it suggests. Namely it is the association between a port number and the SFP cage. 1 is left, 2 is right. Or if you have just 1 cage it is just "1". So the array not only states whether the port is an SFP port (it is true, i.e. not 0) and then also which of the ports based on the geometry of the device.

The sfp_port[] information associates GPIO pins, I2C bus and SDS with the SFP cage. But you still need to know which port it relates to. If you get the 1 and 2 mixed up, you will see in `stat` that a port has a module, but it will not be up, while you are actually seeing traffic on the port (if you have a 1GBit module that is by default configured for a SDS).

Now, another point one may argue about is, whether there should not be an association between port and SDS and not between SFP-cage ans SDS as is currently done. It could even be that there is a fixed association between port and SDS, i.e. port 3 is always SDS 0, and port 8 is always SDS 1, which could be derived from the datasheet and appears to be true for all devices in `machine.c`.

logicog:
> I am not surprised that they update the SOC to an N-version but keep the productnumber the same.

I think the product number is based on what the device does, not what it is. The managed versions are considerably more expensive (usually the price is about double), while they are likely only a small fraction of the devices sold. So probably they squeeze every penny out of the non-managed versions, which are much easier to compare between brands and make the bulk of what is sold. The lower target price includes using the -N versions, which reduces costs for the SoC and board, often a smaller Flash and they remove the reset switch.

I need to find time for taking the photos, I like using natural light...

machine.c
vDorst: Shall we move the discussion to another issue? So I can merge this.
Maybe we also can think about how to support PHY or DIRECT connection.

logicog: Yes. Let's move the discussion. I think you opened up #129 , that is a good place to continue this.

PR #127, opened 2026-02-14
Add button handling with the possibility to revert to default settings

feelfree69:
- added handle_button() function
  - 3 sec press resets the switch
  - 10 sec press restores default settings
- there's a new section @0x6f0000 for default config; same contents as config.txt for now
- firmware upload image is now being flashed to 0x800000 instead of 0x1000000 => devices with 1MB flash also support upgrading firmware
- minor cosmetics:
  - BUILD_DATE also in Web-interface
  - functionality for update moved to new function check_and_flash_update_image()

### DIFF
--- a/rtlplayground.c
+++ b/rtlplayground.c
@@ -1161,6 +1161,9 @@ void idle(void)
 		reg_write_m(RTL837X_REG_SEC_COUNTER);
 		reg_read_m(RTL837X_REG_SEC_COUNTER);
 
+		// Check for button presses once a second
+		handle_button();
+
 #ifdef DEBUG
 		print_sfr_data();
 		write_char('\n');
@@ -1199,9 +1202,6 @@ void idle(void)
 
 	// Check for changes with SFP modules
 	handle_sfp();
-
-	// Check for button presses
-	handle_button();
 
 	// Check new Packets RX
 	handle_rx();


machine.h
vDorst, line 1: Please remove this file, is not parted of the PR.

rtlplayground.c
```diff
@@ -1064,6 +1062,66 @@ void handle_sfp(void)
...
+	__xdata uint32_t source = DEFAULT_CONFIG_START;
+	__xdata uint32_t dest = CONFIG_START;
+
+	flash_region.addr = CONFIG_START;
+	flash_sector_erase();
+
+	for (uint8_t i = 0; i < 8; i++) // 8 * 512 Byte = 4 kByte (1 sector)
+	{
+		flash_region.addr = source;
+		flash_region.len = 0x200;
```
vDorst: Please demagic all the `0x200` numbers.
Also the many `__xdata uint8_t flash_buf[512];` scattered over the codebase.

rtlplayground.c
```diff
@@ -1064,6 +1062,66 @@ void handle_sfp(void)
...
+		dest += 0x200;
+		source += 0x200;
+	}
+
+	print_string("Written default config to flash\n");
+}
+
+void handle_button(void)
+{
+	if (machine.reset_pin == GPIO_NA) {
```
vDorst, line 1090: The compiler complains about
```
rtlplayground.c:1090: warning 94: comparison is always false due to limited range of data type
rtlplayground.c:1091: warning 126: unreachable code
```

this is because `reset_pin` is `int8_t` other types are `uint8_t`.
```c
typedef struct machine {
  ...
  int8_t reset_pin;
}
```
So change `reset_pin` to `uint8_t` is fine.

rtlplayground.c
```diff
@@ -1800,6 +1856,82 @@ void setup_i2c(void)
...
+		__xdata uint16_t i = 0;
+		__xdata uint16_t j = 0;
+		__xdata uint8_t * __xdata bptr;
+		print_string("Identified update image. Checking integrity...");
+		flash_init(0); // Re-initialize flash for non-DIO operation, otherwise flashing will fail
+		set_sys_led_state(SYS_LED_FAST);
+		crc_value = 0x0000;
+		for (i = 0; i < 1024; i++) {
+			flash_region.addr = source;
+			flash_region.len = 0x200;
```
vDorst: Also here demagic value.

rtlplayground.c
```diff
@@ -1064,6 +1062,66 @@ void handle_sfp(void)
...
+void handle_button(void)
+{
+	if (machine.reset_pin == GPIO_NA) {
+		return;
+	}
+
+	bool button_pressed = !gpio_pin_test(machine.reset_pin);
+	if (button_last != button_pressed)
+	{
+		print_string(button_pressed ? "Button pressed\n" : "Button released\n");
```
vDorst, line 1097: Idea to change the SYS LED? start blinking while button is pressed.

rtlplayground.c
```diff
@@ -1135,11 +1193,9 @@ void idle(void)
 	// Check for changes with SFP modules
 	handle_sfp();
 
-	/* Button pressed on KL-8xhm-x2:
-	reg_read(RTL837X_REG_GPIO_32_63_INPUT);
-	if (!(sfr_data[2] & 0x40))
-		print_string("Button pressed\n");
-	*/
+	// Check for button presses
+	handle_button();
```
vDorst: Can we move this into `if (sec_counter >= SYS_TICK_HZ) {}` statement?
Reduces the poll time from 200 Hz to 1 Hz, I think which is fine.
Also the `sfr_data` already has the time-data loaded so that can also be reduced.

vDorst (changes requested):
Some comments.

rtlplayground.c
feelfree69: polling with 1Hz is too slow: a user could make a short press where only the press is detected; not the release. and after several seconds/minutes, he does again a short press, this time only the release is detected.
Resulting in an unwanted "Revert to default action".

vDorst: If this is the case the design is broken. But it is not, because every poll the button status is checked.
Did you tested inside this if-statement?

feelfree69: Yes, I did a test and something went wrong. But you're right, it shouldn't happen.
The only problem will be that button presses/releases shorter than 1 second won't be recognised at all.

I'll give it another try.

vDorst: > The only problem will be that button presses/releases shorter than 1 second won't be recognised at all.

For me this is not a issue at all. Great that it is snappy but we hope that this button is almost never used and you have to press it +3 secs anyways.
We should describe this in the documentation so people know what they can expect.

feelfree69: You were right. The initial bug was that I did not load the `RTL837X_REG_SEC_COUNTER` again in `handle_button` - but `sfr_data` was already overwritten by `gpio_pin_test` ....

UAb5eSMn:
> firmware upload image is now being flashed to 0x800000 instead of 0x1000000 => devices with 1MB flash also support upgrading firmware

Would this complicate things if you wanted to use the update mechanism to revert to the OEM firmware, which is larger?

feelfree69:
> > firmware upload image is now being flashed to 0x800000 instead of 0x1000000 => devices with 1MB flash also support upgrading firmware
> 
> Would this complicate things if you wanted to use the update mechanism to revert to the OEM firmware, which is larger?

I don't know if reverting to OEM firmware is supported or foreseen at all. 

You can use an RTLplayground image built in the `installer` subdir to flash from OEM to RTLPlayground. 
But to my knowledge, there's no way back other than with a clamp and an external programmer.

PR #139, opened 2026-02-25
Fix reset button GPIO-Number for MACHINE_KP_9000_6XHML_X2

logicog:
The GPIO of the reset button was wrong for the KeepLink KP_9000_6XHML_X2 as identified while testing PR #127


vDorst:
thanks

PR #140, opened 2026-02-25
Check for sufficient flashsize before applying an update

feelfree69:
- Flashsize is read via jedec-command from flash chip
  - if flashsize is not sufficient for flashing an update image:
    - the update is rejected
    - the update check at boottime is skipped (otherwise, due to address wrap-around, a device with 512k flash would copy its running image to itself and finally crash)
- cmdline "flash" commands:
  - corrected output
  - removed obsolete commands which did not do useful things other than bricking the device...
- minor cosmetics on console output

### DIFF
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -803,47 +803,27 @@ void cmd_parser(void) __banked
 			}
 		} else if (cmd_compare(0, "stat")) {
 			port_stats_print();
-		} else if (cmd_compare(0, "flash") && cmd_words_b[1] > 0 && cmd_buffer[cmd_words_b[1]] == 'r') {
-			print_string("\nPRINT SECURITY REGISTERS\n");
+		} else if (cmd_compare(0, "flash") && cmd_words_b[1] > 0 && cmd_buffer[cmd_words_b[1]] == 's') {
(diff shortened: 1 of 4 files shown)


logicog:
> So I tried some quick changes, but didn't resolve it.
> Then I noticed that the calculation for the OEM firmware security algorithm at https://github.com/up-n-atom/SWTG118AS/tree/main?tab=readme-ov-file (Firmware versions prior and up to 1.9.x) only uses the first four bytes of the uid. I assumed that the OEM also wasn't able to read the full uid, so didn't look further.

I also came across this. I still believe that the original behavior probably falls under "Well, 32 bits security are also not bad, so why fix this?". It is very likely that it would be possible to read the UUID, provided one knew how to properly use the SPI controller. The problem is that we do not really need the UUID and figuring out the SPI controller would take a lot of experimenting, plus equipment to see what is going on on the SPI lines at these speeds. While I have that equipment, it is really cumbersome and time-consuming to use...

feelfree69:
> So what about displaying the flash size in the Web-interface in MByte. Plus maybe prevent people from uploading an image that they later will not be able to flash?

Excellent ideas, where this PR can be used as a base. 
Several small steps with fast merges are better than waiting for the optimum solution.

PR #143, opened 2026-02-28
Add flash_size to web interface

feelfree69:
Human readable flash sizes in console and Web-Interface

### DIFF
--- a/rtl837x_flash.c
+++ b/rtl837x_flash.c
@@ -12,8 +12,6 @@ __xdata struct flash_region_t flash_region;
 
 __xdata uint32_t flash_size;
 __xdata uint8_t flash_capacity_code;
-__code char * __code flash_size_text[] = { "256 KB", "512 KB", "1 MB", "2 MB", "4 MB", "8 MB", "16 MB" };
-
 
 // For the flash commands, see e.g. Windbond W25Q32JV datasheet
 #define CMD_WRITE_STATUS	0x01
@@ -140,10 +138,16 @@ void flash_read_uid(void)
 
 __code char* get_flash_size_str(void)
 {
-	if (flash_capacity_code >= 0x12 && flash_capacity_code <= 0x18)
-		return flash_size_text[flash_capacity_code - 0x12];
-	else
-		return "unknown";
+	switch (flash_capacity_code) {
+		case 0x12: return "256 KB";
+		case 0x13: return "512 KB";
+		case 0x14: return "1 MB";
+		case 0x15: return "2 MB";
+		case 0x16: return "4 MB";
+		case 0x17: return "8 MB";
+		case 0x18: return "16 MB";
+		default: return "unknown";
+	}
 }
 
 void flash_read_jedecid(void)


httpd/page_impl.c
```diff
@@ -17,6 +17,8 @@
 // #define DEBUG
 #include "debug.h"
 
+extern __xdata char flash_size_str[10];
```
UAb5eSMn: In some locations 10 is used an in others 16

httpd/page_impl.c
```diff
@@ -234,6 +240,9 @@ void send_basic_info(void)
 	slen += strtox(outbuf + slen, BUILD_DATE);
 	slen += strtox(outbuf + slen, "\",\"hw_ver\":\"");
 	slen += strtox(outbuf + slen, machine.machine_name);
+	slen += strtox(outbuf + slen, "\",\"flash_size\":\"");
+	string_to_html(flash_size_str);
```
UAb5eSMn: Or maybe use the numeric flash size in bytes (decimal number, or hexadecimal string) and let the javascript code handle the formatting?

rtlplayground.c
```diff
@@ -240,6 +241,15 @@ uint16_t strtox(register __xdata uint8_t *dst, register __code const char *s)
 	return dst - b;
 }
 
+uint16_t strcpy(register __xdata uint8_t *dst, register const char *s)
```
UAb5eSMn: This function isn't used?
The return value is different than the official strcpy definition. Might be confusing?

rtl837x_flash.c
```diff
@@ -159,8 +160,17 @@ void flash_read_jedecid(void)
 	flash_size = 1UL << cap;
 	print_byte(cap);
 	print_string(" = ");
-	print_long(flash_size);
-	print_string(" Bytes\n");
+	switch(cap) {
+		case 0x12: memcpyc(flash_size_str, "256 KB", 7); break;
```
vDorst: It feels over complicated to create an extra buffer to just store the mem size as a string. Can't we just store the pointer to the code memory string?
And use `print_string` to show it.

rtlplayground.c
```diff
@@ -25,6 +25,7 @@
 
 extern __code const struct machine machine;
 extern __xdata uint32_t flash_size;
+extern __xdata char flash_size_str[10];
```
vDorst: Original size is 16.

rtlplayground.c
vDorst: Do we need the `register`?

httpd/page_impl.c
vDorst: Original size is 16

rtl837x_flash.c
feelfree69: will try

rtlplayground.c
feelfree69: I have no idea - guess I got the "template" from memcpyc()

feelfree69: Indeed, very confusing. Probably I used it intermediately. Will remove.

PR #147, opened 2026-03-02
Bandwidth control

logicog:
Add a bandwidth control mechanism for ingress and egress at a given port.

The documentation contains an example making use of this feature using loopback on a Linux device through an external RTL-switch. Applications of the feature include configuring bandwidth shares of e.g. an uplink connection and the simulation of low bandwidth connections with or without packet drops.

### DIFF
--- a/doc/bandwidth.md
+++ b/doc/bandwidth.md
@@ -1,0 +1,238 @@
+# Egress and Ingress Bandwidth Control
+The RTL8372/3 allows to control the bandwidth of data transmitted (egress) and/or
+admitted (ingress) at any given port. Once admitted, packets are internally switched
+at wire-speed, since the backplane of the devices has a bandwidth of 60GBit/s.
+
+The devices schedules transmission of packets by assigning packets to 8 queues
+implemented in hardware per port, which share a total of 8Mbit of memory internal
+to the switching part of the SoCs. Packets are assigned to the respective queues
+based on the priority assigned to a packet, which can be based on various
+properties of a packet such as  IEEE 802.1P priority, DSCP value, physical port
+number, destination or source MAC, Ether-Type-based, CVID, SVID, IPv4 source or
+destination IP, IPv4/IPv6 TOS field, IPv6 Flow Label and even TCP/UDP
+source/destination port. Once in a queue, packets are scheduled for egress
+based on differnent configurable algorithms.
+
+RTLPlayground currently allows only to control the bandwidth at ingress at a port
+or just before packets leave a port. There is no control of the priority assignment
+or queue scheduling mechanisms. The bandwidth can be controlled in steps of 16Kbit/s
+from 16Kbit/s to 10Gbp/s.
+
+The bandwidth control as currently implemented allows e.g. to assign a certain
+share of bandwidth to an attached device (e.g. to share an uplink), or simulate
+connections with low bandwidth and even bad connectivity with packet drops when
+ingress is not controlled by Flow Control but by simply droping packets.
+
+## Ingress/Egress control
+The relevant registers for controlling Ingress and Egress at a port are:
+```
+#define RTL837X_IGBW_CTRL		0x4c10
+#define IGBW_INC_BYPASS_PKT		0x100
+#define IGBW_INC_IFG			0x80
+#define IGBW_ADM_DHCP			0x20
+#define IGBW_ADM_ARPREQ			0x10
+#define IGBW_ADM_RMA			0x08
+#define IGBW_ADM_BPDU			0x04
+#define IGBW_ADM_RTKPKT			0x02
+#define IGBW_ADM_IGMP			0x01
+#define RTL837X_IGBW_PORT_CTRL		0x4C18
+#define RTL837X_IGBW_PORT_FC_CTRL	0x4C8C
+#define RTL837X_EGBW_PORT_CTRL		0x1c34
+#define RTL837X_EGBW_CTRL		0x447c
+#define EGBW_INC_IFG			0x02
+#define EGBW_CPUMODE			0x01
+```
+`RTL837X_IGBW_CTRL/RTL837X_EGBW_CTRL` control the behaviour of the bandwidth control
+at ingress and egress. The flags such as `IGBW_ADM_DHCP`control whether certain types
+of packets such as DHCP are exempt from being ingress controlled. The
+`IGBW_INC_IFG/EGBW_INC_IFG` flags control whether the Inter Frame Gaps are part of
+the bandwidth being controlled. `EGBW_CPUMODE` controls whether packets generated
+by the internal CPU are subject to egress control.
+
+`RTL837X_IGBW_PORT_CTRL/RTL837X_EGBW_PORT_CTRL` configure the bandwidth for ingress
+and egress at a port.
+
+`RTL837X_IGBW_PORT_FC_CTRL` configures whether packets are bandwidth-controlled using
+Flow Control (port-bit set), or simply dropped (port-bit clear).
+
+## Ingress/Egress bandwidth API
+The code currently provides the following functions:
+```
+void bandwidth_setup(void) __banked;
+void bandwidth_ingress_set(uint8_t port, __xdata uint32_t bw) __banked;
+void bandwidth_ingress_disable(uint8_t port) __banked;
+void bandwidth_ingress_drop(uint8_t port) __banked;
+void bandwidth_egress_set(uint8_t port, __xdata uint32_t bw) __banked;
+void bandwidth_egress_disable(uint8_t port) __banked;
+void bandwidth_status(uint8_t port) __banked;
+```c
+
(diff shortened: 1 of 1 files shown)


rtl837x_bandwidth.c
```diff
@@ -0,0 +1,108 @@
...
+	__xdata uint8_t * __xdata bwptr = &bw;
+
+	print_string("bandwidth_ingress_set called, port "); print_byte(port); write_char('\n');
+	sfr_data[0] = 0;
+	sfr_data[1] = 0x10 | (*(bwptr + 2) >> 4);  // Set bit 20 to enable ingress bandwidth control
+	sfr_data[2] = (*(bwptr + 2) << 4) | (*(bwptr + 1) >> 4);
+	sfr_data[3] = (*(bwptr) >> 4) | (*(bwptr + 1) << 4);
+	reg_write_m(RTL837X_IGBW_PORT_CTRL + port * 4);
+
+	// We enable Flow Control instead of just dropping packets
```
feelfree69, line 45: I would make this as a configuration option. Usually, enabling Flow Control is a bad idea.

logicog: It is configurable, you can change it to dropping the packets instead via `bandwidth_ingress_drop()`. While indeed FC is not great and depends on the support by the linkpartner, it is much better than the alternative, namely to drop the packets, see the examples in the documentation: bandwidth.md. As can be seen, the traffic is no longer deterministic because of packets being dropped arbitrarily in the middle of the configuration causing TCP back-off to trigger. Using Flow Control, the data stream shows reliably the desired bitrate without any dropped packets. So I prefer making FC the default in order to not surprise the user with bitrates that are erratic.

doc/bandwidth.md
```diff
@@ -0,0 +1,238 @@
...
+
+To enable bandwidth control of ingress for physical port 2 to be set to 256 Kbit/s
+do:
+```
+> bw in 2 0100 
+```
+
+To drop packets when the bandwidth is exceeeded at port 2 do:
+```
+> bw in 2 drop
```
feelfree69, line 106: What is the way back, i.e. to enable FLowControl again?

doc/bandwidth.md
```diff
@@ -0,0 +1,238 @@
...
+  Configures or shows the status of bandwidth control
+```
+The bandwidth is given as the `<hexvalue>` in Kbit/s. Note that the control is only
+possible at a granularity of 16 Kbit/s and the minimum value is also 16 Kbit/s. The
+hexadecimal numbers must be given in full bytes, i.e. have an even number of digits.
+
+To enable bandwidth control of ingress for physical port 2 to be set to 256 Kbit/s
+do:
+```
+> bw in 2 0100
```
feelfree69, line 101: This implictly enables flow control, which does work only if the link partner enables it too at its RX side.

I would make the FlowControl feature configurable explicitly.

doc/bandwidth.md
```diff
@@ -0,0 +1,238 @@
...
+`bandwidth_status(port)` shows the current bandwidth control status for a given port
+
+## Bandwidth control configuration on the Serial Console
+The following commands are provided on the serial console:
+```
+> bw [in|out|status] <port> [<hexvalue>|off|drop]
+  Configures or shows the status of bandwidth control
+```
+The bandwidth is given as the `<hexvalue>` in Kbit/s. Note that the control is only
+possible at a granularity of 16 Kbit/s and the minimum value is also 16 Kbit/s. The
```
feelfree69, line 95: I think it's ok at the console to have this granularity of what the chip supports and the unintuitive way of using hex values.
When adding this feature to the Web-Interface, I would expect eg a list of dedicated values, eg. from 16 kbit/s to 10 Gbit/s in steps of factor 2 or 4.

doc/bandwidth.md
logicog: This is maybe a misunderstanding. The bandwidth control does not enable or disable FlowControl. FC is a PHY feature, here we are talking about scheduling even before the MAC layer. The bandwidth control merely _uses_ FC or not. It is not enabling or disabling it. This would require to configure the PHY. I verified it using `sudo ip netns exec client ethtool -a `: if you disable the use of FC, it still is configured and enabled on the link, even if no longer used. It is even used during transmission, it is just not used when the bandwidth is exceeded to lower throughput.

So drop explicitly stops usage of FC. To enable it again, one can again use `bw in 1 1000`.

Do you want an explicit command/function to enabled/disable _use_ of FC?

doc/bandwidth.md
logicog: No, this is a misunderstanding, see above. FC is neither enabled nor disabled. This would not be possible at this level. FC stays on all the time even if not used. BTW: by default it is enabled and used if the link partner enables it. There is no way currently to configure it at the link level. The idea is that like EEE if it works it is good, if not, there is nothing much you can do. So we keep the default for all involved PHYs, which enables it by default.

doc/bandwidth.md
logicog: Good idea! In the web-interface configuring and verifying the settings is much easier. I have already a skeleton for the web-page  but wanted the basics to be done, first.

doc/bandwidth.md
feelfree69: > Do you want an explicit command/function to enabled/disable use of FC?

Yes, see my comment on `bandwidth_ingress_set`. There you implicitly switch FC on
And you have a command to switch FC off (ie to drop packets), but no command to switch FC back on, other than setting the bandwidth limit again.
While the config in SoC in regards of ingress bandwidth and FC is perfectly separated, the console commands are not. That looks as a bad design to me.

cmd_parser.c
```diff
@@ -697,6 +698,62 @@ void parse_eee(void)
...
+	if (cmd_words_b[3] < 0)
+		goto err;
+
+	port = cmd_buffer[cmd_words_b[2]] - '1';
+	port = machine.phys_to_log_port[port];
+
+	if (cmd_words_b[1] > 0 && cmd_compare(1, "status")) {
+		bandwidth_status(port);
+		return;
+	}
```
vDorst: I think this should be above line 706 that checks that we have 3 parameters?

vDorst: Yes it should
```
> bw status
ingress: disabled
egress: disabled
```

vDorst (changes requested):
Not all bw commands are working.

cmd_parser.c
```diff
@@ -697,6 +698,62 @@ void parse_eee(void)
...
+
+	if (cmd_words_b[3] > 0 && cmd_compare(3, "drop")) {
```
vDorst: "off"/`bandwidth_ingress_disable()/bandwidth_egress_disable()` seems to be missing

cmd_parser.c
```diff
@@ -697,6 +698,62 @@ void parse_eee(void)
...
 
+void parse_bw(void)
+{
+	__xdata uint8_t port;
+	__xdata uint32_t bw = 0;
+
+	if (cmd_words_b[3] < 0)
+		goto err;
+
+	port = cmd_buffer[cmd_words_b[2]] - '1';
```
vDorst, line 709: Missing a port range check?

cmd_parser.c
logicog: No, because `bw status` should not work, it should require the port number so the command have should 3 parts. The bw commands always work on a particular port, the bandwidth control is not a global property.

cmd_parser.c
logicog: > "off"/`bandwidth_ingress_disable()/bandwidth_egress_disable()` seems to be missing

Yes, somehow I lost code during a rebase, will fix.

cmd_parser.c
logicog: I'll add a check. Otherwise one could probably kill the switch, because one could get random writes into memory as the port number is used as a multiplier in the memory address calculation.

cmd_parser.c
vDorst: Ah yes, I missed that.

cmd_parser.c
```diff
@@ -697,6 +698,76 @@ void parse_eee(void)
...
+	if (port < 0 || port > 9)
+		goto err;
+
+	port = machine.phys_to_log_port[port];
+
+	if (cmd_words_b[1] > 0 && cmd_compare(1, "status")) {
```
vDorst: Because we change `(cmd_words_b[3] < 0)` on L706. 
Then `cmd_words_b[1] > 0` is always true?
So need we still need to check this?

Same as below `(cmd_words_b[3] > 0)` is that also always true?

logicog:
0x800000 is in kbit now. That would be 8.3 Gbit.

logicog:
You are using 10 Gbit modules, so there is no flow control. That works only with Ethernet. Lack of FC will account for a 15 to 20% loss in performance when you start throttling.

vDorst:
ah that explains it.
Thanks for the work!

PR #150, opened 2026-03-04
Bandwidth web

logicog:
Screenshot_20260304_202356

This PR is on top of PR #147 and provides the Web Interface for controlling the bandwidth

### DIFF
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -150,6 +150,14 @@ uint8_t atoi_hex(uint8_t idx)
 		h_idx++;
 	}
 
+	if (h_idx & 1) {
+		hexvalue[h_idx >> 1] <<= 4;
+		hexvalue[3] = hexvalue[3] >> 4 | (hexvalue[2] << 4);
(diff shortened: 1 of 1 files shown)


vDorst:
I see that after refresh using 'F5', changing the bandwidth issue is still there.
Is that easy to fix?

logicog:
I rebased to main and cleaned up the commits.

logicog:
> strange that `20000` is just fine.

There is a bug in the parser for hex-numbers. It works only when the hexadecimal number has an even number of digits. 30d40 has 5 digits and does not work. I noticed this at the beginning when testing and wanted to fix this either in the web-interface or in the parser, but forgot. Setting 030d40 would work. Will look into it tomorrow.

logicog:
I decided to fix the bug at the root and corrected `atoi_hex` so that it now correctly parses hex numbers with an odd number of digits. BTW: I am sure this took about 10 times longer than fixing the javascript to add a leading 0 when there was an odd number of digits in the conversion result of the number to hex.

PR #151, opened 2026-03-06
Web request concurrency management

logicog:
The web-server in RTLPlayground is single-treaded and can only handle 1 concurrent connection. Web-Browsers such as Firefox use up to 6 concurrent connections to a given domain. This will lead to the `tcp: found no unused connections` showing up in the CLI and timeouts happening in the connections from the browser. While it does not seem possible to control this concurrency for connections for .html and .js files, it is possible to control this behaviour for the xhttp requests done to transfer the status of the ports and the information on the web-pages of the device.

Implement a queueing system for the xhttp requests and control the sequence of .json calls when a page is created.

Screenshot_20260306_185542

As can be seen, the calls for `status.json` and `bandwidth.json` are now coordinated making them finish in < 40ms. The coordination of the sequence of .json requests at the load of a page makes it possible to reload a page in < 300ms, which appears as if the page loads instantaneously.

Of course, the better solution would be to add further concurrent connections to uIP, making use of a queueing system and possibly the use of interrupt-driven RX, however, in terms of snappiness, the present solution already goes a long way.

This is currently on top of the bandwidth PR, but will become independent once those are merged. This is mostly to start testing and discussion.

### DIFF
--- a/html/bandwidth.js
+++ b/html/bandwidth.js
@@ -3,7 +3,6 @@ function createBW() {
   var tbl = document.getElementById('bwtable');
   const limit = '<input type="checkbox" id="limit_port" onchange="exec();">'
   if (tbl.rows.length <= 2  && numPorts) {
-     clearInterval(createBWInterval);
      console.log("CREATING TABLE ", tbl.rows.length);
      for (let i = 2; i < 2 + numPorts; i++) {
        const tr = tbl.insertRow();
@@ -94,6 +93,7 @@ async function applyBandwidth(i) {
 function getBW() {
   var xhttp = new XMLHttpRequest();
   xhttp.onreadystatechange = function() {
+    console.log("IN getBW ");
     if (this.readyState == 4 && this.status == 200) {
       const s = JSON.parse(xhttp.responseText);
       console.log("BW: ", JSON.stringify(s));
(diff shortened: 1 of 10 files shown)


logicog:
Rebased.

logicog:
Unfortunately, the interesting bits in the error message in the screenshot are missing. Could you make a screenshot with the full javascript error, please?

Also, is this Chrome or Firefox?

Could you say, when this appears: does it appear on all web-pages or only on the port statistics page? When you say not reliable: is this reproducible, or random?

logicog:
> It is working with Firefox - except the port status at the top that never worked with Firefox.

Can you explain what did not work with Firefox regarding the port status, exactly? Because I was testing with Firefox and for me it was working.

I have Chromium installed and will start testing with that, too.

logicog:
> Before I allowed loading the port SVG without authentication, it would also redirect to the login page there and not load the SVG at all. Showing that Firefox thinks of the embedded SVG as a different origin.

Then I do not understand why it works for me, but also for others who use Firefox. I am using the ESR version of Firefox. Do you use a different one?

logicog:
> PS: It is working with the ESR version - seems they have changed something after 140.

Could you investigate? Maybe it is the same issue as for Chrome. And sorry, I keep  asking questions, instead of providing answers, I am working in parallel on the management VLAN issue, and I do not make progress...

PR #152, opened 2026-03-07
Fix "Get Configuration" on vlan.html

mmmspatz:
Hi, first off thank you for this project. I was ready to toss my bricked Mokerlink 2G08110GSM when I found this.

I noticed that entering some vlan config, clicking "Update/Create", then clicking "Get Configuration" resulted in Tagged/Untagged/PVID displaying completely wrong data. It looks like the parisng in fetchVLAN() in vlan.js was just completely disconnected from the layout of the registers read by vlan_get() in rtl837x_port.c. Who knows how that happened.

1. Fix fetchVLAN() member/untag parsing: the old code read bits [9:0] as untagged and bits [10:19] as tagged, but the VLAN table register layout is members in [9:0] and untag in [19:10]. Now correctly derives tagged (member && !untag) and untagged (member && untag).

2. Add PVID to vlan.json response: PVID is stored per-port in separate PVID registers (RTL837x_PVID_BASE_REG), not in the VLAN table entry. The old code nonsensically tried to read it from bits [20:29] of the VLAN table. Add port_pvid_get() and build a pvid bitmask in send_vlan() so the JS can parse it correctly.

3. Remove auto-PVID logic from setC(): setC() is called by fetchVLAN() while it's loading existing config, and so this logic mangled the display of the existing config. Also, it doesn't make sense to auto-set the PVID for tagged members (PVID relates to untagged ingress.)

### DIFF
--- a/doc/vlan.md
+++ b/doc/vlan.md
@@ -33,8 +33,8 @@ An entry is deleted by adding an invalid entry (00 instead of 0x02 in
 RTL837x_TBL_DATA_IN_A).
 
 A port is assigned a PVID by setting the PVID-bits of the corresponding
-register of the port. 2 ports share a register. One port uses the higher
-16 bits, the other (even ports) use the lower. The base register is
+register of the port. 2 ports share a register. An odd port uses bits [23:12],
+an even port uses bits [11:0].  The base register is
 RTL837x_PVID_BASE_REG (0x4e1c) and the registers go to 0x4e2c so that also
 the CPU-Port may have a PVID.
 
@@ -51,7 +51,9 @@ registers 0x1238, 0x1338, ...
 The code currently provides the following functions:
 ```
 void port_pvid_set(uint8_t port, __xdata uint16_t pvid) __banked;
-void vlan_create(uint16_t vlan, uint16_t members, uint16_t tagged) __banked;
+uint16_t port_pvid_get(uint8_t port) __banked;
+void vlan_create(void) __banked;   // reads from global vlan_settings
+int8_t vlan_get(register uint16_t vlan) __banked;  // returns data in sfr_data
 void vlan_delete(uint16_t vlan) __banked;
 
 ```


UAb5eSMn:
Thanks for looking into this.
I still get random members from vlan.json for vlans > 1022.
The max value of the vid  input on the web page has a value of 2047. IEEE 802.1Q allows 1 .. 4094 and unfortunately I set all vlans on my LAN in the 3000 and 4000 range.

UAb5eSMn:
Based on the comment by logicog, i tried the following and that seems to work correctly for reading vlan 1 .. 4094:
```diff
diff --git a/html/vlan.html b/html/vlan.html
index 133b84c..7962776 100644
--- a/html/vlan.html
+++ b/html/vlan.html
@@ -12,7 +12,7 @@
     <form id="vform" action="/vlan.html">
       <div>
         <label for="vid">VLAN ID:</label>
-        <input type="number" min="1" max="2047" id="vid" name="vid">
+        <input type="number" min="1" max="4094" id="vid" name="vid">
         <button type="button" style="margin: 0 0 0 24px" onclick="fetchVLAN();">Get Configuration</button>
       </div>
       <br/><br/>
diff --git a/rtl837x_port.c b/rtl837x_port.c
index 1a28c69..79b396e 100644
--- a/rtl837x_port.c
+++ b/rtl837x_port.c
@@ -104,7 +104,7 @@ void vlan_delete(uint16_t vlan) __banked
  */
 int8_t vlan_get(register uint16_t vlan) __banked
 {
-	if (vlan >= 0x3ff) // VLAN 4095 is special
+	if (vlan >= 0xfff) // VLAN 4095 is special
 		return -1;
 
 	REG_WRITE(RTL837X_TBL_CTRL, vlan >> 8, vlan, TBL_VLAN, TBL_EXECUTE);
``` 

It took a bit longer than expected, because the firmware update upload is broken (upload succeeds, serial console shows success, but no new firmware found after reboot), so I had to go find my programmer  + clip.

Some other things I noticed while testing this (out of scope for this pull request):
- Some input validation could be added, so no out of bounds vlans can be read/created/deleted/...
- Most vlan settings were not saved when performing 'Save Settings to Flash' in the web interface (and the sys led kept flashing until the device was rebooted)
- vlan updates aren't atomic. If the browser loses the connection because of the tagged/untagged vlan update, the pvid updates are not executed
- Settings of just one vlan on the page and no way to discover which vlans exist. Displaying all known vlans at once on one page would be nice (a bit like the openwrt luci vlan configuration page)

rtl837x_port.c
```diff
@@ -78,6 +78,17 @@ void port_pvid_set(uint8_t port, __xdata uint16_t pvid) __banked
 	}
 }
 
+uint16_t port_pvid_get(uint8_t port) __banked
+{
+	uint16_t reg = RTL837x_PVID_BASE_REG + ((port >> 1) << 2);
```
feelfree69, line 83: Difficult to understand, at least a comment would be helpful

doc/vlan.md
```diff
@@ -33,8 +33,8 @@ An entry is deleted by adding an invalid entry (00 instead of 0x02 in
 RTL837x_TBL_DATA_IN_A).
 
 A port is assigned a PVID by setting the PVID-bits of the corresponding
-register of the port. 2 ports share a register. One port uses the higher
-16 bits, the other (even ports) use the lower. The base register is
+register of the port. 2 ports share a register. An odd port uses bits [23:12],
+an even port uses bits [11:0].  The base register is
```
feelfree69, line 37: @logicog told in his comment that there are up to 10 ports, 9 physical and CPU port. His code also used bits [19:10] and [9:0]. What is now correct?

logicog: There are two things being mixed up, here. One is the membership (tagged / untagged) of a port to a VLAN. This is done using a VLAN "register" stored in the VLAN table with 10 bits each representing a port map. The index here is the VLAN, the data is a port map. And then there is a per-port PVID, where each port is represented by part of a register-half, into which a PVID is filled. This time the index is a port, and the data is a PVID.

diijkstra: I confirm what @logicog is saying - indexing for PVID is done per port number (0-9). The layout in this description matches what port_pvid_set does and definitely is an improvement as previous version was wrong (12 bits are used not 16).

rtl837x_port.c
diijkstra: How about adding a separate function to return the register address? It could be re-used in port_pvid_set. The logic for the 'registry offset' is `port number/2` and second shift comes from the registry size (4 bytes per register - same as `offset * 4`)

PR #159, opened 2026-03-13
Web: Add SFP state/pin status

diijkstra:
The state of the SFP module is being already send in status.json, but was not parsed via Web UI. When debugging SFP LOS/Signal detection it is useful to see what module is reporting back.

Extended the SFP mouse-over information with:
 - RX LOS as reported in 0x02 bit of register 238
 - External TX Disabled from 0x80 bit of register 238
 - TX fault from 0x04 of the same register
 - The last state of RX LOS pin (available also when there is no 0x40 option on the module)

This should help identify missing/incorrect setup of TX disabled pin. I'm not super sure how 'standard' those locations are - tested on H!Fiber 10G modules as seen below:

Zrzut ekranu_20260313_234334
Zrzut ekranu_20260313_235800

I'm also considering to push only one of the LOS signals i.e. from the pin. On the other hand if pin is misconfigured it may be misleading... Any opinions? Not super critical, but I had to decode status register while hunting for tx-disabled signal.

I'm hoping to add ability to disable SFP port if TX-Disabled pin is present, hence the proposed changes.

### DIFF
--- a/html/main.js
+++ b/html/main.js
@@ -90,11 +90,12 @@ function update(callback) {
 	  var iHTML = "<table border=\"0\" class=\"tt_table\">";
 	  iHTML += "<tr><td align=\"left\">Link speed</td><td>:</td><td>" + linkS[p.link + 1] + "</td></tr>";
 	  if (p.isSFP) {
-            pAdvertised[n] = 0;
+	    pAdvertised[n] = 0;
+	    const hasExtendedStatus = p.sfp_options & 0x40;
 	    iHTML += "<tr><td>Vendor</td><td>:</td><td>" + p.sfp_vendor + "</td></tr>";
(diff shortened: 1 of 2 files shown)


logicog:
Really sweet idea!

> I'm also considering to push only one of the LOS signals i.e. from the pin. On the other hand if pin is misconfigured it may be misleading... Any opinions?

In my opinion, the web-interface is not for debugging the device, but for less experienced users who do not know the difference between the pin and the state. So what about showing this only when there is a discrepancy? And by default show "RX-LOS" only? Then we will get people informing us that there is a problem...

logicog:
I think we were using 2 spaces and then a tab if we had 8 spaces together. This used to be an issue because Javascript does far more indentation compared with C, so using tabs only was indenting too much. But we also had issues with the size of the page in bytes, which was the reason for putting 8 spaces together into a tab. The size no longer matters really, so all spaces and no tabs is also OK.

PR #163, opened 2026-03-14
fix crash

### LINKED ISSUE #162: Console: "stat#" crashes the firmware?
Just made a typo in the console, entered `stat#`. 
The switch then probably enters an endless loop? No console or web activity anymore.

I just had a very quick look at cmd_parser.c, but did not find anything obvious.

BTW: Entering unknown commands should IMHO not be silently ignored, but an error message should be printed.

feelfree69:
This fixes #162.


logicog:
I need to test this and see debug with my own eyes, will do tomorrow. I am not able to think this through this evening.

Issue #157, opened 2026-03-10
Setting a management vlan


UAb5eSMn:
While performing some vlan tests, I've also tried to configure a management vlan, which was added in pull request #105
I'm only able to manage the switch from vlan 1 and only by disabling the management vlan (with 'vlan 0 mgmt').
Setting 'vlan 1 mgmt' (or anything else) does not appear to work.

logicog:
Can you check whether you see broadcast packets on the CPU port in a different VLAN than 1 (just move the printouts higher in the function)? This could be an L2 issue, namely that the MAC of the CPU port is not learned and the packets do not get forwarded to that port. The forwarding logic learns destination MACs on a per-VLAN level. The CPU port currently will only transmit in VLAN 1, this probably prevents learning this VLAN/port combination. Maybe we need to tag outgoing packets  from the CPU port with the management VLAN...

logicog:
So that means, we must make sure that the switch knows the CPU port will take Ethernet frames with the device-MAC and the management VLAN. There are multiple possible solutions:
- Just put a an L2 entry with the MAC and the management VLAN into the L2 table. The problem is that the code for adding entries to the L2 table is not yet written. `port_l2_learned()` will print entries, but we cannot add them. We know how to add e.g. L3 entries for IGMP and VLANs into the same table engine, but not L2 entries. I am also not 100% sure just learning the entry will work. But the SDK would tell us how to add such entries.
- We make uIP VLAN-aware and just add a 802.1Q VLAN tag into a standard Ethernet frame. The problem is that UIP is very carefully built to not have to deal with dynamically detecting and handling packets with different header lengths and layouts. Everything is pre-compiled into structures, so that the 8051 does not throw a tantrum.
- We use the RTL frame descriptor for TX packets that comes before the actual Ethernet frame to ask the switch's packet engine to write us a nice 802.1Q VLAN tag. This has the benefit that we handle the management vlan only in `rtlplayground.c`, uIP does not need to know we are adding the tag. The problem is: while I know this is possible, there is no documentation. The SDK does not have code for the 8051, and the TX descriptor is part of the 8051-specific code for sending a packet.
[...]

PR #166, opened 2026-03-15
Cleanup Frame header structures and fix management VLAN

logicog:
The frame header structure has been mostly magic offsets so far. Add a TX frame descriptor structure and use structures to access the frame descriptor and the frames for TX and RX as well as for assembly of packets.

uIP uses the same buffer for receiving packets as well as for transmitting them, and requires that the application data is aligned in memory for TX and RX packets, so that in some cases data is not even copied over from a request packet into a reply packet. Also the sequence of copying of information requires this alignment. The start of the application data for TX and RX packets is therefore always aligned.

The RTL837x SoCs require a frame descriptor at the beginning of any packet to be sent. This descriptor has a size of 8 bytes. When a packet is received, the RX descriptor can be placed in any memory location and is not part of the frame buffer. Instead, the frame starts directly with the Ethernet frame, which however includes at least an RTL tag (8 bytes) plus optionally 4 bytes for an RTL VLAN tag (2 bytes service VLAN, 2 bytes inner VLAN). Since VLAN support is enabled in our case, the VLAN tag is always present. Because the application data start is fixed, this means that for TX packets we do not need the full memory space from the start of the uip_buffer to the application buffer start when sending normal Ethernet frames. Instead there are 4 buffer bytes at the beginning, which we skip when handing the TX descriptor to the ASIC.. When sending management frames, we insert a dot1Q tag (4 bytes) into the Ethernet frame and push the TX frame descriptor with its initial padding over the beginning of the uip_buffer, so that the frame starts with the actual frame descriptor.

Management VLAN is fixed by inserting a dot1q tag with the management frame. If we are using an untagged management port, the tag is removed by the switch engine, the management device does not see that tag.

### DIFF
--- a/rtl837x_common.h
+++ b/rtl837x_common.h
@@ -72,18 +72,9 @@ struct vlan_tag {
 #define RTL_FRAME_TAG_ID	0x8899
 #define RTL_FRAME_TAG_VERSION	0x04
 
-struct rtl_frame_desc {
-	uint8_t padding[VLAN_TAG_SIZE]; // Pad to the size of a RTL-VLAN tag or dot1q-tag
-	uint8_t tx_seq;
-	uint8_t chksum_flags;	// 0x7 enables Checksums for frame header, L2 and L3
-	uint8_t reserved_1 [2];
-	uint16_t len; // Length is Little Endian
-	uint8_t reserved_2 [2];
-};
-
 // For TX, an 8 byte (plus 4 byte padding when when VLAN is enabled)
 // header describing the frame to be moved to the Asic is used
-#define RTL_FRAME_DESC_SIZE	(sizeof (struct rtl_frame_desc))
+#define RTL_FRAME_DESC_SIZE	12
 
 // This is the standard size of an Ethernet frame header
 #define ETHER_HEADER_SIZE	14
--- a/rtl837x_common.h
+++ b/rtl837x_common.h
@@ -70,10 +70,20 @@ struct vlan_tag {
 #define RTL_TAG_SIZE		(sizeof (struct rtl_tag))
 #define VLAN_TAG_SIZE		(sizeof (struct vlan_tag))
 #define RTL_FRAME_TAG_ID	0x8899
+#define RTL_FRAME_TAG_VERSION	0x04
 
-// For RX and TX, an 8 byte header describing the frame to be moved to the Asic
-// and received from the Asic is used
-#define RTL_FRAME_HEADER_SIZE	8
+struct rtl_frame_desc {
+	uint8_t padding[VLAN_TAG_SIZE]; // Pad to the size of a RTL-VLAN tag or dot1q-tag
+	uint8_t tx_seq;
+	uint8_t chksum_flags;	// 0x7 enables Checksums for frame header, L2 and L3
+	uint8_t reserved_1 [2];
+	uint16_t len; // Length is Little Endian
+	uint8_t reserved_2 [2];
+};
+
+// For TX, an 8 byte (plus 4 byte padding when when VLAN is enabled)
+// header describing the frame to be moved to the Asic is used
+#define RTL_FRAME_DESC_SIZE	(sizeof (struct rtl_frame_desc))
 
 // This is the standard size of an Ethernet frame header
 #define ETHER_HEADER_SIZE	14
@@ -152,5 +162,8 @@ void read_reg_timer(__xdata uint32_t * tmr);
 void sfp_print_info(uint8_t sfp);
 bool gpio_pin_test(uint8_t pin);
 void set_sys_led_state(uint8_t state);
-
+void sds_read(uint8_t sds_id, uint8_t page, uint8_t reg);
+void sds_write_v(uint8_t sds_id, uint8_t page, uint8_t reg, uint16_t v);
+void sds_config_mac(uint8_t sds, uint8_t mode);
+void sds_config(uint8_t sds, uint8_t mode);
 #endif
--- a/rtl837x_init.c
+++ b/rtl837x_init.c
@@ -1,0 +1,247 @@
+#include <stdint.h>
+#include "rtl837x_common.h"
+#include "rtl837x_sfr.h"
+#include "rtl837x_regs.h"
+#include "rtl837x_port.h"
+#include "rtl837x_phy.h"
+#include "phy.h"
+#include "machine.h"
+
+extern __xdata uint8_t sfr_data[4];
+extern __code struct machine machine;
+extern __xdata struct machine_runtime machine_detected;
+
+#pragma codeseg BANK2
+#pragma constseg BANK2
+
+/*
+ * Configure the PHY-Side of the SDS-SDS link between SoC and PHY
+ */
+void static sds_init(void)
+{
+/*
+	p001e.000d:9535 R02f8-00009535 R02f4-0000953a P000001.1e00000d:953a
+	p001e.000d:953a p001e.000d:953a R02f8-0000953a R02f4-00009530 P000001.1e00000d:9530
+
+	RTL8373:
+	p001e.000d:0010 R02f8-00000010 R02f4-0000001a P000001.1e00000d:b7fe
+	p001e.000d:0010 p001e.000d:0010	R02f8-00000010 R02f4-00000010 P000001.1e00000d:b7fe
+*/
+	phy_read(0, PHY_MMD30, 0xd);
+	uint16_t pval = SFR_DATA_U16;
+
+	// PHY Initialization:
+	REG_WRITE(0x2f8, 0, 0, pval >> 8, pval);
+	delay(20);
+
+	pval &= 0xfff0;
+	pval |= 0x0a;
+	REG_WRITE(0x2f4, 0, 0, pval >> 8, pval);
+	delay(10);
+
+	phy_write_mask(0x1, PHY_MMD30, 0xd, pval);
+
+	phy_read(0, PHY_MMD30, 0xd);
+	pval = SFR_DATA_U16;
+
+	REG_WRITE(0x2f8, 0, 0, pval >> 8, pval);
+
+	pval &= 0xfff0;
+	REG_WRITE(0x2f4, 0, 0, pval >> 8, pval);
+
+	phy_write_mask(0x1, PHY_MMD30, 0xd, pval);
+
+	if (machine_detected.isN) {
+		uint16_t pval;
+
+		print_string("  N-settings");
+		// Serdes 0 RX PN swap for 64B/66B
+		sds_read(1, 6, 2);
+		pval = SFR_DATA_U16;
+		sds_write_v(1, 6, 2, pval | 0x2000);
+
+		// Serdes 1 RX PN swap for 8B/10B
+		sds_read(1, 0, 0);
+		pval = SFR_DATA_U16;
+		sds_write_v(1, 0, 0, pval | 0x200);
+
+		// Serdes 0 RX PN swap for 64B/66B
+		sds_read(0, 6, 2);
+		pval = SFR_DATA_U16;
+		sds_write_v(0, 6, 2, pval | 0x2000);
+
+		if (machine_detected.isRTL8373) {
+			// RTL8224: Serdes 0 RX PN swap for 64B/66B
+			// We assume that RTL8373N always paired with RTL8224N.
+			// This sds register value is 0x0000 at reset.
+			// So only write to it.
+			RTL8224_SDS_WRITE(0, 6, 2, 0x2000);
+		} else {
+			// Serdes 0 RX PN swap for 8B/10B
+			sds_read(0, 0, 0);
+			pval = SFR_DATA_U16;
+			sds_write_v(0, 0, 0, pval | 0x200);
+		}
+	}
+}
+
+
+void rtl8373_init(void) __banked
+{
+	print_string("\nrtl8373_init called\n");
+
+	// r65d8:3ffbedff R65d8-3ffbedff
+	reg_bit_set(0x65d8, 0x1d);
+
+	sds_init();
+	// Disable all SERDES for configuration
+	REG_SET(RTL837X_REG_SDS_MODES, 0x000037ff);
+
+	// q000601:c800 Q000601:c804 q000601:c804 Q000601:c800
+	sds_read(0, 0x06, 0x01);
+	uint16_t pval = SFR_DATA_U16;
+	sds_write_v(0, 0x06, 0x01, pval | 0x04);
+	delay(50);
+	sds_read(0, 0x06, 0x01);
+	pval = SFR_DATA_U16;
+	sds_write_v(0, 0x06, 0x01, pval & 0xfffb);
+
+	phy_config_8224();
+	sds_config_mac(1, SDS_OFF);    // Off for now until SFP+ port used
+	sds_config_mac(2, SDS_SGMII);  // For RTL8224
+	sds_config(0, SDS_QXGMII);
+
+	// SDS 1 setup
+	// q012100:4902 Q012100:4906 q013605:0000 Q013605:4000 Q011f02:001f q011f15:0086
+	sds_write_v(1, 0x21, 0x00, 0x4906);
+	sds_write_v(1, 0x36, 0x05, 0x4000);
+	sds_write_v(1, 0x1f, 0x02, 0x001f);
+	sds_read(1, 0x1f, 0x15);
+	pval = SFR_DATA_U16;
+
+	// r0a90:000000f3 R0a90-000000fc
+	reg_read_m(RTL837X_CFG_PHY_MDI_REVERSE);
+	sfr_mask_data(0, 0x0f,0x0c);
+	reg_write_m(RTL837X_CFG_PHY_MDI_REVERSE);
+
+	if (machine_detected.isN) {
+		print_string("  TX_POLARITY_SWAP\n");
+		// FOR N-Version: #TX_POLARITY_SWAP
+		reg_read_m(RTL837X_CFG_PHY_TX_POLARITY_SWAP);
+			sfr_data[2] = 0x59;
+			sfr_data[3] = 0x6a;
+		reg_write_m(RTL837X_CFG_PHY_TX_POLARITY_SWAP);
+	}
+
+	rtl8224_phy_enable();
+
+	// Disable PHYs for configuration
+	phy_write_mask(0xff,PHY_MMD31,0xa610,0x2858);
+
+	// Set bits 0x13 and 0x14 of 0x5fd4
+	// r5fd4:0002914a R5fd4-001a914a
+	reg_bit_set(0x5fd4, 0x13);
+	reg_bit_set(0x5fd4, 0x14);
+
+	// Configure ports
+	uint16_t reg = 0x1238; // Port base register for the bits we set
+	for (char i = 0; i < 9; i++) {
+		// Bit 7 (0x40) enables replacement of the RTL-VLAN tag with an 802.1Q VLAN tag
+		REG_SET(reg, 0xe77);
+		reg += 0x100;
+	}
+
+	// r0b7c:000000d8 R0b7c-000000f8 r6040:00000030 R6040-00000031
+	reg_bit_set(0xb7c, 5);
+
+	// R7124-00001050 R7128-00001050 R712c-00001050 R7130-00001050 R7134-00001050 R7138-00001050
+	// R713c-00001050 R7140-00001050 R7144-00001050 R7148-00001050
+	REG_SET(0x7124, 0x1050); REG_SET(0x7128, 0x1050); REG_SET(0x712c, 0x1050);
+	REG_SET(0x7130, 0x1050); REG_SET(0x7134, 0x1050); REG_SET(0x7138, 0x1050);
+	REG_SET(0x713c, 0x1050); REG_SET(0x7140, 0x1050); REG_SET(0x7144, 0x1050);
+	REG_SET(0x7148, 0x1050);
+
+	reg_bit_set(RTL837X_REG_HW_CONF, 0);
+
+	// enable EEE for all ports at 2.5G and 10G, but don't reset the PHYs
+	port_eee_enable_all(EEE_2G5 | EEE_NORESET);
+
+	// TODO: patch the PHYs
+
+	// Re-enable PHY after configuration
+	phy_write_mask(0xff,PHY_MMD31,0xa610,0x2058);
+
+	// Enables MAC access
+	// Set bits 0xc-0x14 of 0x632c to 0x1f8, see rtl8372_init
+	// r632c:00000540 R632c-001f8540 // RTL8373: 001ff540
+	reg_read_m(0x632c);
+	sfr_mask_data(1, 0x70, 0xf0); // The ports of the RTL8824
+	sfr_mask_data(2, 0x10, 0x1f);
+	reg_write_m(0x632c);
+
+	print_string("\nrtl8373_init done\n");
+}
+
+
+void rtl8372_init(void) __banked
+{
+	print_string("\nrtl8372_init called\n");
+
+	sds_init();
+	phy_config(8);	// PHY configuration: External 8221B?
+	phy_config(3);	// PHY configuration: all internal PHYs?
+	// Set the MAC SerDes Modes Bits 0-4: SDS 0 = 0x2 (0x2), Bits 5-9: SDS 1: 1f (off)
+	// r7b20:00000bff R7b20-00000bff r7b20:00000bff R7b20-00000bff r7b20:00000bff R7b20-000003ff r7b20:000003ff R7b20-000003e2 r7b20:000003e2 R7b20-000003e2
+	reg_read_m(RTL837X_REG_SDS_MODES);
+	sfr_mask_data(1, 0, 0x03);
+	sfr_mask_data(0, 0, 0xe2);
+	reg_write_m(RTL837X_REG_SDS_MODES);
+
+	// r0a90:000000f3 R0a90-000000fc
+	reg_read_m(RTL837X_CFG_PHY_MDI_REVERSE);
+	sfr_mask_data(0, 0x0f, 0x0c);
+	reg_write_m(RTL837X_CFG_PHY_MDI_REVERSE);
+
+	// Disable PHYs for configuration
+	phy_write_mask(0xf0,PHY_MMD31,0xa610,0x2858);
+
+	// Set bits 0x13 and 0x14 of 0x5fd4
+	// r5fd4:0002914a R5fd4-001a914a
+	reg_bit_set(0x5fd4, 0x13);
+	reg_bit_set(0x5fd4, 0x14);
+
+	// Configure ports 3-8:
+	//
+	// r1538:00000e33 R1538-00000e37 r1538:00000e37 R1538-00000e37 r1538:00000e37 R1538-00000f37
+	// [...]
+	///
+	uint16_t reg = 0x1238 + 0x300; // Port base register for the bits we set
+	for (char i = machine.min_port; i <= machine.max_port; i++) {
+		// Bit 7 (0x40) enables replacement of the RTL-VLAN tag with an 802.1Q VLAN tag
+		REG_SET(reg, 0xe77);
+		reg += 0x100;
+	}
+
+	// r0b7c:000000d8 R0b7c-000000f8 r6040:00000030 R6040-00000031
+	reg_bit_set(0xb7c, 5);
+
+	reg_bit_set(RTL837X_REG_HW_CONF, 0);
+
+
+	// enable EEE for all ports at 2.5G and 10G, but don't reset the PHYs
+	port_eee_enable_all(EEE_2G5 | EEE_NORESET);
+	
+	// TODO: patch the PHYs
+
+	// Re-enable PHY after configuration
+	phy_write_mask(0xf0,PHY_MMD31,0xa610,0x2058);
+
+	// Enables MAC access
+	// Set bits 0xc-0x14 of 0x632c to 0x1f8, see rtl8372_init
+	// r632c:00000540 R632c-001f8540 // RTL8373: 001ff540
+	reg_read_m(0x632c);
+	sfr_mask_data(1, 0x70, 0x80);
+	sfr_mask_data(2, 0x10, 0x1f);
+	reg_write_m(0x632c);
+	print_string("\nrtl8372_init done\n");
+}
--- a/rtlplayground.c
+++ b/rtlplayground.c
@@ -149,43 +149,40 @@ struct eth_in {
 	u16_t ether_type;
 };
 
-struct rtl_dot1q_frame {
-	union {
-	struct {
-		uint8_t tx_seq;
-		uint8_t chksum_flags;	// 0x7 enables Checksums for frame header, L2 and L3
-		uint8_t reserved_1 [2];
-		uint16_t len; // Length is Little Endian
-		uint8_t reserved_2 [2];
-		struct uip_eth_addr dst;
-		struct uip_eth_addr src;
-		uint16_t tpid;
-		uint16_t tci;
-	} q_frame;
-	struct {
-		uint8_t padding[VLAN_TAG_SIZE]; // Pad to the size of a RTL-VLAN tag or dot1q-tag
-		uint8_t tx_seq;
-		uint8_t chksum_flags;	// 0x7 enables Checksums for frame header, L2 and L3
-		uint8_t reserved_1 [2];
-		uint16_t len; // Length is Little Endian
-		uint8_t reserved_2 [2];
-		struct uip_eth_addr dst;
-		struct uip_eth_addr src;
-	} nonq_frame;
-	};
-};
-
 // Dot 1Q tag size is the size of tpid + tci
 #define DOT_1Q_TAG_SIZE 4
+
+struct q_frame {
+	uint8_t tx_seq;
+	uint8_t chksum_flags;	// 0x7 enables Checksums for frame header, L2 and L3
+	uint8_t reserved_1 [2];
+	uint16_t len; // Length is Little Endian
+	uint8_t reserved_2 [2];
+	struct uip_eth_addr dst;
+	struct uip_eth_addr src;
+	uint16_t tpid;
+	uint16_t tci;
+};
+
+struct nonq_frame {
+	uint8_t padding[DOT_1Q_TAG_SIZE];
+	uint8_t tx_seq;
+	uint8_t chksum_flags;	// 0x7 enables Checksums for frame header, L2 and L3
+	uint8_t reserved_1 [2];
+	uint16_t len; // Length is Little Endian
+	uint8_t reserved_2 [2];
+	struct uip_eth_addr dst;
+	struct uip_eth_addr src;
+};
 
 #define ETH_IN ((__xdata struct eth_in *)&uip_buf[0])
 #define ETHERTYPE_OFFSET (12 + VLAN_TAG_SIZE + RTL_TAG_SIZE)
 
 // The output frame structure with initial frame descriptor including padding
-#define FRAME (((__xdata struct rtl_dot1q_frame *)&uip_buf[0])->nonq_frame)
+#define FRAME ((__xdata struct nonq_frame *)&uip_buf[0])
 
 // The output frame structure with 802.1Q field and the padding moved before the buffer-start
-#define FRAME_Q (((__xdata struct rtl_dot1q_frame *)&uip_buf[-VLAN_TAG_SIZE])->q_frame)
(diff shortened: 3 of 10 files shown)


rtl837x_init.c
```diff
@@ -0,0 +1,247 @@
...
+extern __code struct machine machine;
+extern __xdata struct machine_runtime machine_detected;
+
+#pragma codeseg BANK2
+#pragma constseg BANK2
+
+/*
+ * Configure the PHY-Side of the SDS-SDS link between SoC and PHY
+ */
+void sds_init(void)
```
diijkstra: I think it can be made a static, as it is not used outside of this object file

rtl837x_common.h
```diff
@@ -62,10 +62,20 @@ struct vlan_tag {
 #define RTL_TAG_SIZE		(sizeof (struct rtl_tag))
 #define VLAN_TAG_SIZE		(sizeof (struct vlan_tag))
 #define RTL_FRAME_TAG_ID	0x8899
+#define RTL_FRAME_TAG_VERSION	0x04
+
+struct rtl_frame_desc {
+	uint8_t padding[VLAN_TAG_SIZE]; // Pad to the size of a RTL-VLAN tag or dot1q-tag
+	uint8_t tx_seq;
+	uint8_t chksum_flags;
```
diijkstra: Can we add some hint about flags, i.e. 0x07 => enable all checksums? Easier to spot here than in rtlplayground.c

rtlplayground.c
```diff
@@ -859,26 +906,25 @@ uint8_t sfp_read_reg(uint8_t slot, uint8_t reg)
...
-	uip_buf[VLAN_TAG_SIZE + 6] = uip_buf[VLAN_TAG_SIZE + 7] = 0;
+	FRAME->tx_seq = tx_seq++;
+	FRAME->chksum_flags = 0x07;    // Enable all checksums
+	FRAME->reserved_1[0] = 0x00; FRAME->reserved_1[1] = 0x00;
+	FRAME->len = uip_len;
+	FRAME->reserved_2[0] = 0x00; FRAME->reserved_2[1] = 0x00;
+
+	// For the management VLAN we insert an 802.1Q VLAN tag
+	if (management_vlan) {
+		for (uint8_t i = 0; i < sizeof(struct rtl_dot1q_frame) - 8; i++)
```
diijkstra: Can we somehow avoid magic `-8`?

rtlplayground.c
```diff
@@ -859,26 +906,25 @@ uint8_t sfp_read_reg(uint8_t slot, uint8_t reg)
...
+			uip_buf[i] = uip_buf[i + VLAN_TAG_SIZE];
```
diijkstra: I do not understand this part. What am I missing?

FRAME_Q is moved by VLAN_TAG_SIZE before uip start, while FRAME is pointing to uip_buf, right? 

So if FRAME->tx_seq is pointing to uip_buf + VLAN_TAG_SIZE, than we are effectively copying what we initialized in lines 909-911? I don't understand how this is vlan tag?

rtlplayground.c
```diff
@@ -551,15 +578,35 @@ void nic_rx_packet(register uint16_t buffer, register uint16_t ring_ptr)
  */
 void nic_tx_packet(uint16_t ring_ptr)
 {
-//	uint16_t buffer = (uint16_t) tx_buf;
-	uint16_t buffer = (uint16_t) uip_buf + VLAN_TAG_SIZE;
-	SFR_NIC_DATA_U16LE = buffer;
-	
+	uint16_t len;
+
+	/* If we have a management VLAN, we have inserted a dot1Q-tag into the frame,
```
diijkstra: This contradicts the statement in line 162. I don't understand how does it work, why we are moving pointer here while FRAME_Q is pointing before start of the buffer?

rtlplayground.c
```diff
@@ -133,8 +134,34 @@ __xdata bool button_last;
...
+	struct rtl_tag rtl_tag;
+	struct vlan_tag vlan_tag;
+	u16_t hwtype;
+	u16_t protocol;
+	u8_t hwlen;
+	u8_t protolen;
+};
+
+struct rtl_dot1q_frame {
+	struct rtl_frame_desc desc;
```
diijkstra: Shouldn't we move VLAN padding from rtl_frame_desc to rtl_dot1q_frame? That way FRAME_Q would be a pointer with padding, before uip_buf and FRAME without VLAN padding at the start of uip_buf.

rtlplayground.c
```diff
@@ -133,8 +134,34 @@ __xdata bool button_last;
...
+};
+
+#define ETH_IN ((__xdata struct eth_in *)&uip_buf[0])
 #define ETHERTYPE_OFFSET (12 + VLAN_TAG_SIZE + RTL_TAG_SIZE)
 
+// The output frame structure with initial frame descriptor including padding
+#define FRAME ((__xdata struct rtl_frame_desc *)&uip_buf[0])
+
+// The output frame structure with 802.1Q field and the padding moved before the buffer-start
+#define FRAME_Q ((__xdata struct rtl_dot1q_frame *)&uip_buf[-VLAN_TAG_SIZE])
```
diijkstra: Out of curiosity, shouldn't we define some padding to uip_buf explicitly? I'm  worried that if sdcc will move the location of uip_buf, we will do nasty stuff.

diijkstra (commented):
I'm obviously struggling with understanding the 'before uip_buffer' part. Will have another go tomorrow

rtlplayground.c
vDorst: This feels dangerous and error prone in the future.

Also the buffer definition is missing a big warning that `rx_headers` can't be moved from `uip_buf` because it is used to store the rx-vlan data so that the current `uip_buf` be used without move the data.
```c
// NIC buffers for packet RX/TX
__xdata uint8_t rx_headers[16]; // Packet header(s) on RX
__xdata uint8_t uip_buf[UIP_CONF_BUFFER_SIZE+2];
```

For me, it make sense to make this also a struct so we ensure it always stays together.

vDorst (changes requested):
One point that feels dangerous.

rtl837x_init.c
logicog: Will do!

rtl837x_common.h
logicog: Will do. At one point I even knew, which bit which checksum was. There are 3 bits for the 3 checksums (Ethernet frame, L2-checksum and L3 checksum).

rtlplayground.c
logicog: Will do.

rtlplayground.c
logicog: Yes. We are moving the entire Ethernet frame header start 4 bytes (the size of a dot1Q tag) forward in the buffer. Because we need to insert that tag. Probably the confusion comes because I am not properly separating VLAN_TAG_SIZE and the size of the dot1Q tag. Both of them are 4 bytes long, which is probably not a coincidence, but they mean entirely different things. the VLAN_TAG_SIZE relates to the RTL proprietary VLAN tag, which is 2 bytes outer VLAN-ID and 2 bytes inner VLAN-id, whereas the dot1Q VLAN tag is 2 bytes: 1 a tag id, the other the 4 bit priority and 12 bit VLAN.
Maybe I should introduce a DOT_1Q_SIZE define, which would also be 4 bytes?

rtlplayground.c
logicog: Need to check.

rtlplayground.c
logicog: We could start even later in `uip_buf`. The mess starts befause uIP is not VLAN aware and does not support variable size frame headers. But I do not think that sdcc will pull the rug from under our feet. The bytes before the start of uip_buf are never written to, and uip itself always starts placing its own data well into that buffer. uIP is aware that microcontrollers may need to place additional frame information before the Ethernet frame and supports this. If you look into `uipopt.h` the defines are:
```
#ifdef UIP_CONF_LLH_LEN
#define UIP_LLH_LEN UIP_CONF_LLH_LEN
#else /* UIP_CONF_LLH_LEN */
#define UIP_LLH_LEN     ETHER_HEADER_SIZE + RTL_TAG_SIZE + VLAN_TAG_SIZE
#endif /* UIP_CONF_LLH_LEN */
```
`UIP_LLH_LEN` is where uIP starts putting its application data to. Note that it is not dynamic, i.e. can take a VLAN tag into account. Also, arriving packets do not have a dot1q tag.

logicog: This is a misunderstanding. rx_headers can be anywhere they want. the tx-headers need to start directly before the Ethernet frame, which is the reason they are inside the `uip_buf`. There are 16 rx_headers because I initially hoped to be able to somehow have a packet queue for multiple TCP connections...

The -VLAN_TAG_SIZE does not point to memory ever written to, except if we used all 16  rx_headers, which is not the case.

rtlplayground.c
logicog: I will reformulate to:
```
	/* If we have a management VLAN, we have inserted a dot1Q-tag into the frame and
	 * the frame starts at the beginning of uip_buf with the RTL TX descriptor,
	 * otherwise the frame is a normal Ethernet frame which starts with
	 * an RTL TX descriptor being padded at the beginning, in the second case
	 * we need to skip the padding for the sending of the frame.
	 */
```

rtlplayground.c
logicog: No. The frame descriptor is used by uIP and other places where we create IP frames to calculate the entire size of the Ethernet frame header including anything the SoC needs. uIP is not VLAN aware, so it only needs to know to add 4 bytes to the frame header size, which come with the `rtl_frame_desc`
```
$ rg rtl_frame_desc
rtlplayground.c
152:    struct rtl_frame_desc desc;
166:#define FRAME ((__xdata struct rtl_frame_desc *)&uip_buf[0])
923:            // Shift the ethernet header before the HW type including the rtl_frame_desc to the beginning of uip_buf
925:            for (uint8_t i = 0; i < sizeof(struct rtl_dot1q_frame) - sizeof(struct rtl_frame_desc) + VLAN_TAG_SIZE; i++)

rtl837x_common.h
72:struct rtl_frame_desc {
83:#define RTL_FRAME_DESC_SIZE  (sizeof (struct rtl_frame_desc))
$ rg RTL_FRAME_DESC_SIZE
rtl837x_stp.c
85:#define STP_O ((__xdata struct stp_pkt *)&uip_buf[RTL_FRAME_DESC_SIZE])

uip/uip_arp.c
135:#define BUF_O ((__xdata struct arp_hdr_o *)&uip_buf[RTL_FRAME_DESC_SIZE])
136:#define IPBUF ((__xdata struct ethip_hdr *)&uip_buf[RTL_FRAME_DESC_SIZE])

rtl837x_common.h
83:#define RTL_FRAME_DESC_SIZE  (sizeof (struct rtl_frame_desc))

uip/uipopt.h
450:#define UIP_LLH_LEN     ETHER_HEADER_SIZE + RTL_FRAME_DESC_SIZE

```

rtlplayground.c
diijkstra: OK, that sounds way better!

rtlplayground.c
```diff
@@ -136,8 +137,37 @@ __xdata bool button_last;
 __xdata uint8_t button_sec_counter_last;
 __sbit tx_buf_empty;
 
+struct eth_in {
+	struct uip_eth_addr dst;
+	struct uip_eth_addr src;
+	struct rtl_tag rtl_tag;
+	struct vlan_tag vlan_tag;
+	u16_t hwtype;
```
diijkstra: How about `ether_type`?  I'm also a little bit vary of naming for following fields. They are fine for ARP packages, but will not be valid for IPv4. Since I do not see anywhere us taking the size of eth_in, how about leaving the protocol/hwlen/protolen out?

rtlplayground.c
```diff
@@ -947,12 +996,12 @@ void handle_rx(void)
 			if (uip_len) {
 				tcpip_output();
 			}
-		} else if (uip_buf[ETHERTYPE_OFFSET] == 0x08 && uip_buf[ETHERTYPE_OFFSET + 1] == 0x06) { // ARP?
+		} else if (ETH_IN->hwtype == HTONS(0x0806)) { // ARP?
```
diijkstra: This is definitely an ARP, as per https://en.wikipedia.org/wiki/EtherType

rtlplayground.c
```diff
@@ -947,12 +996,12 @@ void handle_rx(void)
...
 			uip_arp_arpin();
 			if (uip_len) {
 			    tcpip_output();
 			}
-		} else if (uip_buf[ETHERTYPE_OFFSET] == 0x08 && uip_buf[ETHERTYPE_OFFSET + 1] == 0x00) { // TCP?
+		} else if (ETH_IN->hwtype == HTONS(0x0800)) { // TCP?
```
diijkstra: More likely a IPv4

rtlplayground.c
diijkstra: Agree, you are right.

rtlplayground.c
diijkstra: @logicog correct me if I'm wrong. We are **not** writing anything to before uip_buf when sending frame, right? 

FRAME_Q is just used to move our view over FRAME so that the headers before dot1q tag can be moved to the left. The output frame starts exactly at uip_buf[0], right? 

Would it be bad performance wise if we hide FRAME_Q arithmetic inside  tcpip_output and nix_tx_packet? That way, it would not be exposed to accidental use.

rtlplayground.c
```diff
@@ -862,26 +912,27 @@ uint8_t sfp_read_reg(uint8_t slot, uint8_t reg)
...
-	uip_buf[VLAN_TAG_SIZE + 6] = uip_buf[VLAN_TAG_SIZE + 7] = 0;
+	FRAME->tx_seq = tx_seq++;
+	FRAME->chksum_flags = 0x07;    // Enable all checksums
+	FRAME->reserved_1[0] = 0x00; FRAME->reserved_1[1] = 0x00;
+	FRAME->len = uip_len;
+	FRAME->reserved_2[0] = 0x00; FRAME->reserved_2[1] = 0x00;
+
+	// For the management VLAN we insert an 802.1Q VLAN tag
+	if (management_vlan) {
+		// Shift the ethernet header before the HW type including the rtl_frame_desc to the beginning of uip_buf
```
diijkstra:
```suggestion
		// Shift the ethernet header before the HW type including the rtl_frame_desc (excluding initial padding) to the beginning of uip_buf
```
Definitely clearer now!

rtlplayground.c
diijkstra: Agree, you are right.

rtlplayground.c
```diff
@@ -862,26 +912,27 @@ uint8_t sfp_read_reg(uint8_t slot, uint8_t reg)
...
+		// to allow space to insert the dot 1Q tag
+		for (uint8_t i = 0; i < sizeof(struct rtl_dot1q_frame) - sizeof(struct rtl_frame_desc) + VLAN_TAG_SIZE; i++)
```
diijkstra: How about this or some slightly changed comment?
```suggestion
		for (uint8_t i = 0; i < sizeof(struct rtl_dot1q_frame) - (VLAN_TAG_SIZE*2); i++)
```
We are copying the rtl output tag + src/dst addresses, omitting the built-in padding and tpid/tci 'footer' which is not there for non vlan data. 

It took me some time to mentally notice that we have both padding and vlan in the struct... Can we add a comment for tpid/tci that those fields are present only when vlan is enabled? And when set, the padding is not there?

Or even... lets make a `rtl_dot1q_frame` union with two structures:
- with padding but without tpid/tci
- without padding but with tpid/tci

That way, we can treat uip_buf[0] as one or the other, transforming the padded version to un-padded with vlan when needed.

diijkstra (changes requested):
OK, the code is correct. But I really think we should at least hide FRAME_Q in output functions as it is very easy to shoot yourself in the foot. Since we are not using tpid/tcid outside the output, I think we can try playing with unions to provide a type which we can cast to and shift the view over output frame. This would avoid 'virtually' moving before uip_buf just to shift tpid/tci into the picture.

Anyhow, I do confirm that the code looks to be correct

logicog:
@diijkstra Your comments all make a lot of sense. I am traveling over this weekend and will only be able to answer in detail starting on Sunday.

rtlplayground.c
logicog: Will fix as suggested.

rtlplayground.c
logicog: Will remove question-mark.

rtlplayground.c
logicog: Will change to IPv4 and remove question mark.

rtlplayground.c
logicog: I changed this to use a union. I believe this makes the code indeed more understandable. Could you have a look?

logicog:
I believe that I addressed all the concerns. Could someone check this and possibly merge this PR, or provide feedback on what is still missing?

UAb5eSMn:
Something may have regressed for the management vlan. I rebased your branch on main, so could be I messed something up.
I tested this pull request for ticket #157 on 2026-03-15 and it mostly worked as expected then (except for 'vlan 0 mgmt', which still only worked for ports connected to vlan 1).
Now management is only possible with 'vlan 0 mgmt' from ports corrected to vlan 1
After changing from 'vlan mgmt 0'  to 'vlan 1 mgmt', that keeps working for about 30 seconds and then no more (cached L2 information timing out or something)
Setting other management vlan does not work anymore at all

logicog:
Thanks a lot @UAb5eSMn for the feedback! I probably made a last minute mistake, and I know, I only tested whether RX/TX works, because I thought the rest was a given. I will, however, only be able to look into this on Monday, as I will be traveling all day tomorrow.

logicog:
I tried the following while pinging the device via a link on port 1:
```
> vlan 1 mgmt
```
ping works and I see:
```
> l2

06:1e:04:b9:xx:xx       0x0001  learned CPU
00:e0:4c:43:xx:xx       0x0001  learned 1
```

Now I change to management VLAN 2:
```
> vlan 2 mgmt
Management VLAN set to 0x0002
```
(ping no longer works)
I then did:
```
> vlan 2 1u

vlan_create called
vlan: 0x0002, members: 0x0210, tagged: 0x0200
vlan_create done 

> pvid 1 2
port_pvid_set called 
```
(ping works again)
```
> l2

06:1e:04:b9:xx:xx       0x0001  learned CPU
00:e0:4c:43:xx:xx       0x0001  learned 1
00:e0:4c:43:xx:xx       0x0002  learned 1
06:1e:04:b9:xx:xx       0x0002  learned CPU
```
(both VLAN 1 and VLAN 2 are learned).
After about 2 minutes I see:
```
> l2

00:e0:4c:43:xx:xx       0x0002  learned 1
06:1e:04:b9:xx:xx       0x0002  learned CPU
```
i.e. VLAN 1 has been forgotten, but the ping still works. I then checked the web and it also works and shows the correct configuration for VLAN 2 (port 1 is untagged member, PVID of port 1 is 2).

I then did:
```
> vlan 3 mgmt
Management VLAN set to 0x0003
```
and the ping stops. Setting the management VLAN back to 2 re-enables ping.
So I believe this works as intended. Can you send the sequence of steps where it fails for you, @UAb5eSMn ?

UAb5eSMn:
Note that I rebased your branch on main, because my device locks up if pull request #190 is missing.

My startup config:
```
ip 192.168.10.248
gw 192.168.10.1
netmask 255.255.255.0
vlan 1 1 2 8t
vlan 102 3 4 8t
vlan 103 5 6 8t
vlan 104 7 8t 9
vlan 666 8t
pvid 1 1
pvid 2 1
pvid 3 102
pvid 4 102
pvid 5 103
pvid 6 103
pvid 7 104
pvid 8 666
pvid 9 104
passwd admin
vlan 0 mgmt
```

With this setting I can reach the device admin interface / ping on port 1, 2 and 8 (8 from another switch with tagged vlan 1 traffic)
No response on the other ports that are not connected to vlan 1.

After executing `vlan 1 mgmt`, even those 3 ports stop responding

After executing `vlan 1 1 2 3 4 5 6 7 8t 9`, still nothing (still with `vlan 1 mgmt`)
And then after disabling the management vlan with `vlan 0 mgmt`, a response is possible on all ports.

logicog:
After a reboot, I did:
```
> vlan 1 1 2 8t

vlan_create called
vlan: 0x0001, members: 0x0231, tagged: 0x0201
vlan_create done 

> vlan 102 3 4 8t

vlan_create called
vlan: 0x0066, members: 0x02c1, tagged: 0x0201
vlan_create done 

> vlan 103 5 6 8t

vlan_create called
vlan: 0x0067, members: 0x0309, tagged: 0x0201
vlan_create done 

> vlan 104 7 8t 9

vlan_create called
vlan: 0x0068, members: 0x0201, tagged: 0x0201
vlan_create done 

> vlan 666 8t

vlan_create called
vlan: 0x029a, members: 0x0201, tagged: 0x0201
vlan_create done 

> pvid 1 1

port_pvid_set called 

> pvid 2 1

port_pvid_set called 

> pvid 3 102

port_pvid_set called 

> pvid 4 102

port_pvid_set called 

> pvid 5 103

port_pvid_set called 

> pvid 6 103

port_pvid_set called 

> pvid 7 104

port_pvid_set called 

> pvid 8 666

port_pvid_set called 

> pvid 9 104

port_pvid_set called 

> passwd admin

> vlan 0 mgmt
Management VLAN disabled
0x0000

> vlan 1 mgmt
Management VLAN set to 0x0001

```
During the whole time, a ping to port 1 worked. I then re-based on top of current main (for #190 ) and redid the test. Still, I am always able to ping port 1 with management VLANs 0 and 1. Can you try with a simpler setup, so that we can close in on the issue?
I rebased the PR and pushed again.

UAb5eSMn:
Thanks. Sorry it took so long. I rebuilt with the rebased branch and it's working now.
The only small remaining issue is with 'vlan 0 mgmt'. That does not work from ports that are not a member of vlan 1. I can live with that, so merge would be fine with me.

logicog:
> Thanks. Sorry it took so long. I rebuilt with the rebased branch and it's working now.

Thanks a lot for testing!

> The only small remaining issue is with 'vlan 0 mgmt'. That does not work from ports that are not a member of vlan 1. I can live with that, so merge would be fine with me.

Maybe one has to make the CPU port explicitly member of the respective VLANs, although I thought that was already the case. Anyway, I think we can also leave this for another time, as at least the configuration is safe. @vDorst could you take a decision on whether we can merge this?

vDorst:
I am not again that `vlan 0 mgmt` will listen on all vlans.
But my option is that is should not be the default.
Default should be only listen on vlan 1.
Is that the default right now?

UAb5eSMn:
> I am not again that `vlan 0 mgmt` will listen on all vlans. But my option is that is should not be the default. Default should be only listen on vlan 1. Is that the default right now?

No, `vlan 0 mgmt` is the default.
Since `vlan 0 mgmt`  has some strange behaviour, setting `vlan 1 mgmt` as default could make sense.
I wouldn't mind if `vlan 0 mgmt` was removed entirely: https://github.com/logicog/RTLPlayground/issues/157#issuecomment-4063520509

logicog:
I rebased this on top of current main, but it seems that vlan parsing is currently broken:

```
X-VLAN: 0x0001
 RX dst: 061e04b94f1a
 MGMT-VLAN: 0x0000


> vlan 2 mgmt

>  RX-VLAN: 0x0001
 RX dst: 061e04b94f1a
 MGMT-VLAN: 0x0000
 RX-VLAN: 0x0001

```

The `vlan 2 mgmt`command should report something:

```

		if (cmd_compare(2, "mgmt")) {
			management_vlan = vlan_settings.vlan;
			if (!vlan_settings.vlan)
				print_string("Management VLAN disabled\n");
			else
				print_string("Management VLAN set to "); print_short(management_vlan); write_char('\n');
			return;
		}
```

Also, the command history no longer works. Maybe #203 ? I can only investigate further this evening...

vDorst:
I changed `atoi_short()` in #204 but removed the `*vlan = 0` https://github.com/logicog/RTLPlayground/pull/204/files#diff-b292ca554f4948e53f64e55d6372d772ba7283a870c965238dd697c857436788L189

On the `atoi_byte()` is was better to first process the value on a local variable and then store it on in xdata location.
I also tried that on  `atoi_short()` but codegen what not better, but did not fully restore the old code.

I had to change the history copy a bit in #203, https://github.com/logicog/RTLPlayground/pull/203/changes, you may can start here to look for the history issues.

logicog:
I just tested the following:
I ping the switch on port 1 and then reset the switch. I then issue:
```
> pvid 1 2
> vlan 2 1
```
The ping continues to work (because the management vlan by default is 0).
I then do:
```
> vlan 1 mgmt
```
ping no longer works (packets at CPU-port have VLAN-tag 2, and 1 is required). I now do:
```
> vlan 0 mgmt
```
And ping again works, but evidently port 1 is still a member of vlan 1. So now I do:
```
> vlan 1 2
```
And now ping stops to work. While packets still arrive with VLAN 2 at the CPU port, and they get treated (because of management VLAN 0), the return packets are no longer sent out of port 1. That is probably because they do not have a VLAN tag, like at all, because VLAN tags are only added for packets from the CPU-port if a management VLAN is set. Since no VLAN is the same as VLAN 1 for the RTL hardware, but port 1 is not a member of VLAN 1, it gets egress filtered.

From my point of view, this looks like a feature of the RTL hardware. Setting management VLAN to 0 still has the benefit that you can reach the CPU from different VLANs (provided the egress ports are still member of VLAN 1). So in my opinion we should keep things as they are now. Maybe with a bit of documentation on the management VLAN.

logicog:
> Default should be only listen on vlan 1.
> Is that the default right now?

No. And you are right, I will change the behavior for the management VLAN to be by default 1.

vDorst:
`vlan 0 mgmt` will eventually be possible. This is because we want to offer services across multiple VLANs, such as management, DHCP server, etc.

The code may not be ready yet, but it will be soon.

logicog:
> The code may not be ready yet, but it will be soon.

Indeed. For that we need to have vlan-aware services, and currently the code is not able to do that.

vDorst:
In https://github.com/logicog/RTLPlayground/pull/166#discussion_r3020373323, there is a talk about converting/moving the two structs into a union, but it seems that is not happened?
I expected that `eth_in` and `rtl_dot1q_frame` are placed in a union.

UAb5eSMn:
> In [#166 (comment)](https://github.com/logicog/RTLPlayground/pull/166#discussion_r3020373323), there is a talk about converting/moving the two structs into a union, but it seems that is not happened? I expected that `eth_in` and `rtl_dot1q_frame` are placed in a union.

There was a commit for that, but it was removed during the rebase on april 14th:
https://github.com/logicog/RTLPlayground/commit/b48f68ccecd61ba643b5e6b4e6fbfb6fd0f8b0a3

logicog:
Yes, I somehow messed that up, I use different machines for development and I somehow got them out of sync. I am trying to fix this just now. I believe I have the commit back in, but now I am having trouble with the CLI.

logicog:
Ok, so now it works, but by default the management VLAN needs to be turned off, i.e. be 0. Because it only works with management VLAN = 1 if one also properly tags the packets arriving at the switch, and that is not the case in the default configuration.

logicog:
> There was a commit for that, but it was removed during the rebase on april 14th:
> [b48f68c](https://github.com/logicog/RTLPlayground/commit/b48f68ccecd61ba643b5e6b4e6fbfb6fd0f8b0a3)

BTW: How did you manage to find this?

UAb5eSMn:
> > There was a commit for that, but it was removed during the rebase on april 14th:
> > [b48f68c](https://github.com/logicog/RTLPlayground/commit/b48f68ccecd61ba643b5e6b4e6fbfb6fd0f8b0a3)
> 
> BTW: How did you manage to find this?

Github mentions the commit ID's of the heads of both branches (and I still had them in my local repo):
b48f68ccecd61ba643b5e6b4e6fbfb6fd0f8b0a3 (old)
391061ac2324dbb892a08b802a276ce50d08b26a (rebased)

So the git log can be viewed and shows that the number of commits is different.

git range-diff can show the differences. The sixth commit is different (the new one might be bad) and the ninth commit is missing.

```
git range-diff 97e73d95aad31186a6e65db587d655a76eb521db..b48f68ccecd61ba643b5e6b4e6fbfb6fd0f8b0a3 d4d12fe3b4dc1341e8ee64bc4215db73a54f591e..391061ac2324dbb892a08b802a276ce50d08b26a

 1:  7091d61 =  1:  79d59b0 Move initialization routines to BANK2
 2:  ab4f6c3 =  2:  4d703be Cleanup DHCP client code
 3:  66cfd1b =  3:  af1a368 Add frame descriptor structure
 4:  25a2987 =  4:  7eceb7e Cleanup STP, use new frame descriptor definitions
 5:  5cb3e30 =  5:  a23d130 Add NOTHS macro
 6:  8af5f46 !  6:  a254bfc Use header structures of TX and RX, add dot1q for management VLAN TX
    @@ rtlplayground.c
      #include "cmd_editor.h"
     @@ rtlplayground.c: __xdata bool button_last;
      __xdata uint8_t button_sec_counter_last;
    - __sbit tx_buf_empty;
    + volatile __bit tx_buf_empty;

     +struct eth_in {
     +  struct uip_eth_addr dst;
     +  struct uip_eth_addr src;
     +  struct rtl_tag rtl_tag;
     +  struct vlan_tag vlan_tag;
    -+  u16_t ether_type;
    ++  u16_t hwtype;
    ++  u16_t protocol;
    ++  u8_t hwlen;
    ++  u8_t protolen;
     +};
     +
     +struct rtl_dot1q_frame {
    @@ rtlplayground.c: uint8_t sfp_read_reg(uint8_t slot, uint8_t reg)
     +  if (management_vlan) {
     +          // Shift the ethernet header before the HW type including the rtl_frame_desc to the beginning of uip_buf
     +          // to allow space to insert the dot 1Q tag
    -+          for (uint8_t i = 0; i < sizeof(struct rtl_dot1q_frame) - (VLAN_TAG_SIZE * 2); i++)
    ++          for (uint8_t i = 0; i < sizeof(struct rtl_dot1q_frame) - sizeof(struct rtl_frame_desc) + VLAN_TAG_SIZE; i++)
     +                  uip_buf[i] = uip_buf[i + DOT_1Q_TAG_SIZE];
     +          FRAME_Q->desc.len += DOT_1Q_TAG_SIZE;
     +          FRAME_Q->tpid = HTONS(0x8100);
    @@ rtlplayground.c: void handle_rx(void)
                                tcpip_output();
                        }
     -          } else if (uip_buf[ETHERTYPE_OFFSET] == 0x08 && uip_buf[ETHERTYPE_OFFSET + 1] == 0x06) { // ARP?
    -+          } else if (ETH_IN->ether_type == HTONS(0x0806)) { // ARP
    ++          } else if (ETH_IN->hwtype == HTONS(0x0806)) { // ARP?
                        uip_arp_arpin();
                        if (uip_len) {
                            tcpip_output();
                        }
     -          } else if (uip_buf[ETHERTYPE_OFFSET] == 0x08 && uip_buf[ETHERTYPE_OFFSET + 1] == 0x00) { // TCP?
    -+          } else if (ETH_IN->ether_type == HTONS(0x0800)) { // IPv4
    ++          } else if (ETH_IN->hwtype == HTONS(0x0800)) { // TCP?
                        if (!management_vlan || management_vlan == rx_packet_vlan) {
                                uip_arp_ipin(); // Learn MAC addresses in TCP packets
                                uip_input();
 7:  27eef0b =  7:  6cfc699 Use RTL descriptor definition
 8:  9c5a283 =  8:  391061a Correctly use RTL_FRAME_DESC_SIZE for header length
 9:  b48f68c <  -:  ------- Make frame structure a union
```

logicog:
> git range-diff can show the differences. The sixth commit is different (the new one might be bad) and the ninth commit is missing.

Thanks for the explanation! range-diff was something I didn't known anything about.

logicog:
There was still an error in the frame buffer alignment: while it worked, it was still not what we wanted, namely get rid of the padding. I fixed that and re-enabled management vlan to be by default 1. Re-based and pushed.

UAb5eSMn:
I've not looked at the code, but I have re-tested with your rebased branch at https://github.com/logicog/RTLPlayground/commit/5924bfcb5097fdf28925f4aaf0c2375f965365e4 and it all seems fine to me. Management vlans work as expected and vlan 1 is the default management vlan.

logicog:
> I've not looked at the code, but I have re-tested with your rebased branch at [5924bfc](https://github.com/logicog/RTLPlayground/commit/5924bfcb5097fdf28925f4aaf0c2375f965365e4) and it all seems fine to me. Management vlans work as expected and vlan 1 is the default management vlan.

Thanks so much for all the testing work you do!

logicog:
I moved again the structures out of the union, as this was not adding anything, as discussed with @vDorst elsewhere. This is merely semantics and does not change how the code works. I tested it and pushed this.

vDorst:
```c
for (uint8_t i = 0; i < sizeof(struct q_frame) - (VLAN_TAG_SIZE * 2); i++)
```

`VLAN_TAG_SIZE` = 4 bytes. times two = 8 bytes.
But the padding is only 4 bytes. So I think we are coping 4 bytes too short.
Also using `DOT_1Q_TAG_SIZE` and `VLAN_TAG_SIZE` for the same function.
I would like to see that only `DOT_1Q_TAG_SIZE` is used in `struct nonq_frame` and `tcpip_output()`. And use the sizeof(struct nonq_frame) - padding, because that is what we are moving.

This is what I tested.
```patch
index 6ac9a85..6ee8532 100644
--- a/rtlplayground.c
+++ b/rtlplayground.c
@@ -149,6 +149,9 @@ struct eth_in {
        u16_t ether_type;
 };
 
+// Dot 1Q tag size is the size of tpid + tci
+#define DOT_1Q_TAG_SIZE 4
+
 struct q_frame {
        uint8_t tx_seq;
        uint8_t chksum_flags;   // 0x7 enables Checksums for frame header, L2 and L3
@@ -162,7 +165,7 @@ struct q_frame {
 };
 
 struct nonq_frame {
-       uint8_t padding[VLAN_TAG_SIZE];
+       uint8_t padding[DOT_1Q_TAG_SIZE];
        uint8_t tx_seq;
        uint8_t chksum_flags;   // 0x7 enables Checksums for frame header, L2 and L3
        uint8_t reserved_1 [2];
@@ -172,9 +175,6 @@ struct nonq_frame {
        struct uip_eth_addr src;
 };
 
-// Dot 1Q tag size is the size of tpid + tci
-#define DOT_1Q_TAG_SIZE 4
-
 #define ETH_IN ((__xdata struct eth_in *)&uip_buf[0])
 #define ETHERTYPE_OFFSET (12 + VLAN_TAG_SIZE + RTL_TAG_SIZE)
 
@@ -997,9 +997,9 @@ void tcpip_output(void)
 
        // For the management VLAN we insert an 802.1Q VLAN tag
        if (management_vlan) {
-               // Shift the ethernet header before the HW type including the rtl_frame_desc to the beginning of uip_buf
+               // Shift the ethernet header `nonq_frame` without the padding of `DOT_1Q_TAG_SIZE`-bytes before the HW type including the rtl_frame_desc to the beginning of uip_buf
                // to allow space to insert the dot 1Q tag
-               for (uint8_t i = 0; i < sizeof(struct q_frame) - (VLAN_TAG_SIZE * 2); i++)
+               for (uint8_t i = 0; i < sizeof(struct nonq_frame) - DOT_1Q_TAG_SIZE; i++)
                        uip_buf[i] = uip_buf[i + DOT_1Q_TAG_SIZE];
                FRAME_Q->len += DOT_1Q_TAG_SIZE;
                FRAME_Q->tpid = HTONS(0x8100);  // Change ether-type to Dot1Q
```

logicog:
> `VLAN_TAG_SIZE` = 4 bytes. times two = 8 bytes.
> But the padding is only 4 bytes. So I think we are coping 4 bytes too short.

No, because the tpid and tci fields of struct q_frame are not yet filled. These fields, which are the last 4 bytes are (over-)written to a couple of lines below, so they can, but do not need to be copied.  Which is the reason I use sizeof(struct q_frame), to remind the reader we are creating a q_frame structure. Struct q_frame and non_qframe have the same size because of the padding, so it does not matter whether one writes either of them in the sizeof(), but struct q_frame is clearer to me because that is what is being created, and then one also understands the "2 *".

I'll use DOT_1Q_TAG_SIZE as suggested in the output path. VLAN_TAG_SIZE came in because it was used as padding for when the incoming frame has no VLAN-extension to the Realtek (0x8899) tag.

logicog:
PR updated as indicated above.

logicog:
I changed the copy operation to copy 20 bytes over for the Q-frame.

vDorst:
Thanks

PR #167, opened 2026-03-16
Fix browser issues with Chrome and latest Firefox

logicog:
Fix the issue with Chrome that it would not automatically retry xhttp requests. Also fix the issue with Firefox latest versions (> 140) that would block Javascript from changing style elements of SVG graphics.


TylerDurden-23 (approved):
I've tested it with Firefox 148.0.2, Edge and Brave - and it seems to work perfectly fine now.

PR #172, opened 2026-03-21
Document firmware upload via CLI

diijkstra:
While working on ingress console commands, I was missing a quick way to upload firmware via CLI.
Documented the process with curl, maybe will be usefull for someone. 

And another fix for ASIC port mapping on trendnet. Physical mapping was wrong after switching SFP ports around.


logicog:
Great idea, thanks a lot for sharing!

Issue #169, opened 2026-03-17
ZX-SWTGW218AS can't set VLANs


Kabbone:
pcb: SWTG118AS-v2.0
Branch: main

It seems I can't set any vlans, neither on the command line nor through the web interface.
When I try to set it on the command line it just hangs with no output, not recoverable.
Tried different commands
`vlan 100 2t`
`vlan 100 IOT 2t`

To be said, the OEM firmware didn't work anymore (blinking SYS LED), even with flashing it didn't recover, only with RTLPLayground flash.

logicog:
> Indeed it does work if I add back the !isletter check. Without checking I assume the word/command in cmd is perhaps not \0 terminated?

I believe the patch is actually even correct. cmd is 0-terminated, they are the commands, such as "vlan". I spent several hours today trying to understand what is going on. But as soon as I enable some of the debugging in the cmd_compare() fiunction, it actually works also with the patch (you may want to try that yourself, it is quite strange). I am starting to believe it may be a compiler bug, because the most obvious change in the assembly is that a short jump is changed to long one. Possibly the patch without isletter breaks because it generates shorter code, which the compiler fails to generate correctly. I know compiler bugs are extremely rare, but at the moment I do not have any better explanation. I hope that @ranma can have a look at this, he is an expert for the sdcc compiler.

PR #173, opened 2026-03-21
Console: Add Ingress type & VLAN show table

diijkstra:
Added new `vlan show` command to dump current VLAN settings. Supports printing PVID & ingress filtering per port.

Added new `ingress [p]<mode>` command to setup ingress filtering. Added wrappers for enabling/disabling vlan filtering. Currently not configurable, but state in console is read via ASIC registers.

Full VLAN dump in the console & web support of new commands will be added later.

**Notice**:
Marking as draft due to #169 and #157, lets wait for #167 to be merged first. 
When time allows I will extend vlan dump to cover VLAN configuration.

### DIFF
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -382,9 +382,8 @@ void parse_ingress(void)
 	}
 	__xdata uint8_t log_port = 0;
 	__xdata vlan_ingress_mode_t mode = VLAN_INVALID;
-	__xdata uint8_t w = 1;
 
-	if (vlan_ingress_mode_parse(cmd_buffer[cmd_words_b[w]], &mode)) {
+	if (vlan_ingress_mode_parse(cmd_buffer[cmd_words_b[1]], &mode)) {
 		// Setting mode for all ports at once
 		for (log_port = machine.min_port; log_port <= machine.max_port; log_port++) {
 			if (!port_ingress_filter(log_port, mode)) {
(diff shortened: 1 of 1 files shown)


logicog:
#167  was merged, and I believe #157 is also resolved. The remaining issue is #169 , for this there are work-arounds such as enabling debugging in the function, which somehow heals the generated code, but the root-cause is not understood. There is also PR #174 , which should fix the issue of the VLAN display in the browser.

cmd_parser.c
```diff
@@ -342,16 +342,86 @@ void parse_vlan(void)
...
+}
+
+void parse_ingress(void)
+{
+	if (cmd_words_b[1] <= 0) {
+		goto err;
+	}
+	__xdata uint8_t log_port = 0;
+	__xdata vlan_ingress_mode_t mode = VLAN_INVALID;
+	__xdata uint8_t w = 1;
```
logicog: It is not necessary to define this already here. The first time us is for w fixed to 1, so you can just use '1'  when used for the vlan_ingress_mode_parse. Then you can use an IRAM variable for the loop, i.e.:
` for(uint8_t w = 1; cmd_words_b[w] > 0; w++)`. As long as that w ends up in the overlay section OSEG, this should be cleaner code, because w is used only in the loop, reduces XMEM usage (the compiler does not use the fact that w is not visible outside of the function and the byte is lost), should be faster, because there is no XMEM access involved, plus shorter, as the XMEM access requires multiple opcodes.

diijkstra: Definitely you are right, I couldn't decide between for and while ;) Addressed in latest commit.

PR #174, opened 2026-03-21
Fix display of VLANs for ports

logicog:
The display of port-membership of VLANs was not correctly displayed, since the relationship between physical and logical ports was not taken into account. Fix that.


TylerDurden-23 (commented):
I've tested it - and it is working properly now.
@logicog Thanks for the fix.

PR #176, opened 2026-03-24
Improve compatibility with 2.5G SFP modules

janh:
Some modules might report a nominal bitrate of 2500 Mbd or 3200 Mbd instead of 3100 Mbd. Accept the entire range to make them work.

### DIFF
--- a/rtlplayground.c
+++ b/rtlplayground.c
@@ -997,7 +997,7 @@ static inline uint8_t sfp_rate_to_sds_config(register uint8_t rate)
 {
 	if (rate == 0xd)
 		return SDS_1000BX_FIBER;
-	if (rate == 0x1f)  // Ethernet 2.5 GBit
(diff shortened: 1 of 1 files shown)


janh:
I looked at the Linux code again: Your knowledge isn't outdated, it depends on the compliance code reported by the module.

If some Fibre Channel speed bits are set, it ends up in another code path, which doesn't allow 2500 Mbd: https://github.com/torvalds/linux/blob/c369299895a591d96745d6492d4888259b004a9e/drivers/net/phy/sfp-bus.c#L279-L291

Same for active or passive cables: https://github.com/torvalds/linux/blob/c369299895a591d96745d6492d4888259b004a9e/drivers/net/phy/sfp-bus.c#L197-L214

Not sure though if it is an intentional difference or an oversight...

PR #177, opened 2026-03-25
Add support for K0501W boards

janh:
This board appears for example in the Davuaz Da-K6501W switch, and is designed similarly to Hi-K0402WS.

However, it only has one SFP port and another 2.5G copper port using RTL8221B 2.5G PHY instead. Also, some components on the board, like additional LEDs, mode switch, and second flash chip are not populated.

Like earlier revisions of the K0402W(S) board, there is no UART, so debugging is limited.

As this is an unmanaged switch, installation needs a flash programmer.


logicog:
Our boot-log should also say whether it is -N or not

PR #181, opened 2026-03-27
Improve compatibility with 1G SFP modules

janh:
It looks like a similar change as for 2.5G is also needed for 1G modules, as some vendors round the actual 1250 MBd down to 1200 MBd.


TylerDurden-23:
Wouldn't it be better to also use a range comparison here?

vDorst:
Thanks!

@TylerDurden-23 for two values this is fine.

PR #183, opened 2026-03-27
Fix I2C access and SFP diagnostic monitoring in web interface

janh:
The current I2C access code works for what it is actually used for. But it is broken when trying to do larger transfers.

Extend the handling of SFP diagnostic data to include calibration data when needed. Also fix scaling for TX/RX power values and convert them to dBm.

### DIFF
--- a/html/main.js
+++ b/html/main.js
@@ -112,6 +112,10 @@ function decodeSfpRxPower(val, cal) {
   return applyRxPowerCalibration(rxPower, cal) / 10000;
 }
 
+function convertPowerTodBm(val) {
+  return 10 * Math.log10(val);
+}
+
 function update(callback) {
   var xhttp = new XMLHttpRequest();
   xhttp.onreadystatechange = function() {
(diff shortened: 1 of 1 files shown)


vDorst:
Too bad I can't test the calibrated values.
None of the SFP modules I have, provides a calibration value.

But I found one bug in our main code.
One on the modules used `\0` to terminate string-values.

Brave and Firefox don't like a `\0` in a string.

```
VM140:1 Uncaught SyntaxError: Bad control character in string literal in JSON at position 1068 (line 1 column 1069)
    at JSON.parse (<anonymous>)
    at xhttp.onreadystatechange (main.js:126:22)
```

```json
    {
        "portNum": 5,
        "logPort": 8,
        "isSFP": 1,
        "enabled": 1,
        "sfp_options": "0x68",
        "sfp_temp": "0x20d9",
        "sfp_vcc": "0x80e7",
        "sfp_txbias": "0x1c23",
        "sfp_txpower": "0x0000",
        "sfp_rxpower": "0x0000",
        "sfp_state": "0x02",
        "sfp_vendor": "ZYXEL\0\0\0\0\0",
        "sfp_model": "PMG3000-D20B\0\0\0",
}
```
Screenshot_20260327_234126

httpd/page_impl.c
```diff
@@ -196,13 +200,10 @@ void sfp_send_data(uint8_t slot, uint8_t reg, uint8_t len)
 		reg_read_m(RTL837X_REG_I2C_CTRL);
 	} while (sfr_data[3] & 0x1);
```
vDorst, line 201: bit 0 indicated transaction is done.
bit 1 indicated a error when true.

PR #185, opened 2026-03-29
Fix swgt024 v2.0

vDorst:
Fix swgt024 v2.0 managed version.
Also added unmanaged version.


TylerDurden-23 (commented):
Looks good, compiles correctly.
We should limit the number of renames of existing devices though.

PR #189, opened 2026-04-03
Add support for keepLink KP-9000-9XHML-X (2M-PCB23-V2.2 version)

bstreiff:
keepLink has at least two versions of the KP-9000-9XHML-X in circulation. RTLPlayground already had support for the 2M-PCB23-V3.1 version. This change documents and adds support for the 2M-PCB23-V2.2 version.

The V2.2 version does not work with the V3.1's LED configuration, which motivates adding it as a new machine type (`MACHINE_KP_9000_9XHML_X_V2_2`). The existing `MACHINE_KP_9000_9XHML_X` is renamed to `MACHINE_KP_9000_9XHML_X_V3_1` to avoid ambiguity.

Confirmed to work:
- LED configuration matches stock firmware
- Port ordering matches label
- A SFP can be detected and EEPROM read (I don't have a fiber cable to test link)
- Reset button presses are detected


vDorst:
Thanks

PR #192, opened 2026-04-06
Add commands to read/write SerDes and PHY registers

janh:
I wrote this to play around a bit with the SerDes and PHY registers. Maybe it is also useful for others. Not sure though if these commands are really important enough to merge into main branch.

### DIFF
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -164,6 +164,20 @@ uint8_t atoi_hex(uint8_t idx)
 }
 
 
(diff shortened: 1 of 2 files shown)


logicog:
This is a great idea. I am working on support for the 10GBit Ethernet devices and this will come in handy. The code looks good. I will need to test a bit more. BTW: Sorry for the delay, it is Easter holidays...

logicog:
Tested and works great. Thanks a lot for sharing!

PR #148, opened 2026-03-02
Syslog: Duplicates console output to remote syslog server; Web-Interface: Send console commands to device

feelfree69:
Probably far from perfect, but at least it works for me.... helps developing with devices that do not have a serial console attached.

You can enter a console command in the web interface (System Settings, Console Tab), and send it to the device.
All output from the console can be send via UDP to a configurable syslog server.
I did not try out with real syslog servers so far, but just read the messages from wireshark output....

PR #195, opened 2026-04-09
Fix reading configuration over flash block boundaries

logicog:


UAb5eSMn:
I can confirm that a configuration > 256 characters (329 in my case, so not tested with something huge) can be saved and read back from the web interface. The configuration is also applied successfully during bootup (verified the serial console output).

Edit: It appears I was just lucky. There's some issues with another config. I'll report back when I've investigated.

Edit2:
The issue is with a cmd that's a bit too long (more than 127 characters), so it doesn't fit in cmd_buffer. Then the code exits the while loop at: while (cmd_idx < (CMD_BUF_SIZE - 1)) {
Other commands still in flash_buf will also not be processed. The next blocks are read into flash_buf but those are also not processed anymore, because cmd_idx did not get reset to 0.
In summary: After a command that's too long, no further commands are issued. That may be intentional, but maybe print an error message then?

logicog:
I will see what can be done about the long commands, this has always been an issue.

logicog:
I added graceful handling of lines in the configuration file that are too long. With the following file
```
$ cat config.txt
ip 192.168.10.247
gw 192.168.10.1
netmask 255.255.255.0
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
eee status

```
I now get:
```

A minimal prompt to explore the RTL8372:

Clock register: 0x00001101
Register 0x7b20/RTL837X_REG_SDS_MODES: 0x00000bff
Verifying PHY settings:

1       On      1000M   0x00000000      0x00000000      0x0000000e      0x00000000
6       NO SFP  Down    0x00000000      0x00000000      0x00000000      0x00000000
Setting ip: 192.168.10.247
Setting gw: 192.168.10.1
Setting netmask: 255.255.255.0
ERROR: Command too long: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa

```
Note that an error is printed and the device does not hang. However, the final command after the too long line (`eee status`) is not executed, i.e. the command execution is stopped at the first error, similar to what a shell script would do.

vDorst:
I wonder, can we visual show that there is a boot error so the user know there is an issue.

logicog:
We could set a `boot_ok` variable once the boot commands have all executed without an error. This could be shown in the web-interface, e.g. with an exclamation mark somewhere if the boot did not go OK. Currently we however do not return any error codes for commands that fail, merely these too long lines are reported when executing the configuration. Note that you cannot actually enter lines that are too long on the CLI because they would not fit into the line buffer. We could start by setting a global error variable when a command fails (greetings from the non-reentrant beginnings of libc) and `boot_ok` would only be set if there are no errors during command execution of the configuration.

vDorst:
`boot_ok` is I think not usefull when config stops at a weird point that you can't recover from.

So I was thinking what should we do when this happens.
The issue is that not every device has a `SYSTEM`-LED nor a button.
So I was thinking, maybe we should we default to a failsafe-default config.
Which can also be the default config.

Default config does:
- Put all ports in disabled mode, so your network is not leaking all over the place.
- Only one port is active to accept web-requests. (Only pick one port that is handled by the SOC on both 82 and 83 devices.)
- Fixed password, ip and netmask.
- Ideally a error code why we defaulted to the failsafe-config.

logicog:
That would be kind of complicated and confusing since we would need 2 different configuration files and 2 different configuration areas. And the issue is here already in the compiled-in configuration file.

For some things like the default password, we already compile some things in without a configuration file. We could maybe fall back to a default configuration that is compiled in and not in a configuration file. But that would still mean that the failed configuration file execution would need to trigger a re-boot into that failsafe mode in which no configuration file is executed.

feelfree69:
> We could maybe fall back to a default configuration that is compiled in and not in a configuration file.

Note that the compiled in config.txt is written twice to the image, one serves as the Default Config which is never overwritten (Location `DEFAULT_CONFIG_LOCATION`), the other one is the user modifiable part (at `CONFIG_LOCATION`).

Of course that could be 2 different files at compile-time, but at the moment I don't see any benefit.

The fallback to default is already implemented, but triggered by the button only.

logicog:
Thanks for the reminder @feelfree69 , I had forgotten about that!

So, why don't we treat the config.txt configuration as always OK, even if it fails. I mean, if you are able to compile that in, you probably know what you are doing and have some way of reverting. The problem would be the configuration that the user can modify from the web. My understanding is that this gets executed on boot. So what about treating this configuration if it fails to execute like as if the user had pressed the button and the failsafe configuration gets flashed over to the current one. That way we have some consistency what happens and do not complicate things further. It would also be relatively easy to implement: it is basically just: if error detected in boot config -> goto flash over failsafe configuration.

feelfree69:
> it is basically just: if error detected in boot config -> goto flash over failsafe configuration.

Even if we decide to implement that, I would not add this to the current PR.

logicog:
> Even if we decide to implement that, I would not add this to the current PR.

Yes. @vDorst and I have a tendency to feature-creep in PRs. ;-)
I'll add the variable that tracks an error when the initial configuration is executed and we can leave it at that for now, I hope. Will push that in a bit.

cmd_parser.c
```diff
@@ -860,6 +862,7 @@ uint8_t cmd_tokenize(void) __banked
 	print_string_x(&cmd_buffer[0]);
 	write_char('<'); write_char('\n');
 #endif
+	err_status = 0;
```
vDorst: replace `0` with `ERR_OK`.

cmd_parser.c
```diff
@@ -1165,28 +1171,42 @@ void execute_config(void) __banked
...
+				break;
+		}
+		if (cmd_idx >= (CMD_BUF_SIZE - 1)) {
+			cmd_buffer[cmd_idx] = '\0';
+			print_string("ERROR: Command too long: ");
+			print_string_x(cmd_buffer);
+			write_char('\n');
+			err_status = ERR_CMD_TOO_LONG;
+			goto config_done;
+		}
```
vDorst: I think this put in the top of while loop above.
Because it is the same check.
Change the the while loop back to 
```c
do {
} while(cfg_idx)
```

vDorst (commented):
Small thinks

cmd_parser.c
```diff
@@ -1165,28 +1171,42 @@ void execute_config(void) __banked
...
 
 		len_left -= FLASH_READ_BURST_SIZE;
 		pos += FLASH_READ_BURST_SIZE;
```
vDorst: I was hoping by swapping these, the compiler could reuse the already known values. But is doesn't.

cmd_parser.c
logicog: Yes, you are absolutely right. Will fix.

cmd_parser.c
logicog: Yes, there is something strange with how sdcc implements the flash interface, it is using DSEG variables and eats IRAM, see also my comment in PR #148 . We need to fix this, somehow.

vDorst: len_left could be just a 'uint8_t', instead of counting the actual size.
We can count just number of pages.

```c
	__xdata uint8_t len_left = CONFIG_LEN / FLASH_READ_BURST_SIZE;
```
Rename the `len_left`, `pages_left`.

logicog: Fixed as suggested and pushed.

logicog:
All suggestions implemented - amazing how much more one can squeeze out when more eyes look at a PR!

vDorst:
Thanks! Looking great.

One thing what I am thinking about is extra compiletime checks around `CONFIG_LEN` and `FLASH_READ_BURST_SIZE`. To ensure `CONFIG_LEN` is always a multiple of `FLASH_READ_BURST_SIZE`.

It is also fine for an other PR. So I can merge this one.

logicog:
I added that compile-time check:
```
sdcc -MMD -mmcs51 -I. -Ihttpd -Iuip -o output/cmd_parser.rel -c cmd_parser.c
cmd_parser.c:1486:10: error: #error "CONFIG_LEN not a multiple of FLASH_READ_BURST_SIZE"
 1486 |         #error "CONFIG_LEN not a multiple of FLASH_READ_BURST_SIZE"
      |          ^~~~~
```

vDorst:
Thanks!

PR #197, opened 2026-04-10
Port Isolation CLI command

logicog:
This adds support for parsing a `isolate` CLI command to configure port isolation. The syntax is:
```
isolate <port> [show|off] [port]...
```
Show shows the isolation settings for the given port, off allows traffic from the given port to any other given port, including the CPU port, which otherwise has the number 10.

### DIFF
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -1146,7 +1146,10 @@ uint8_t cmd_tokenize(void) __banked
 	line_ptr = 0;
 	is_white = 1;
 	uint8_t word = 0;
-	cmd_words_b[0] = -1;
+
+	for (uint8_t i = 0; i < N_WORDS; i++)
+		cmd_words_b[i] = -1;
+
 	while (cmd_buffer[line_ptr] && line_ptr < CMD_BUF_SIZE - 1) {
 		if (is_white && cmd_buffer[line_ptr] != ' ') {
 			is_white = 0;
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -371,6 +371,74 @@ err:
 	print_string("Error: vlan (<vlan-id>|show) [port][t/u]...\n");
 }
 
+
+void parse_isolate(void)
+{
+	__xdata uint16_t members = 0;
+
+	if (cmd_words_b[3] <= 0)
+		goto err;
+
+	print_string("\nISOLATE ");
+
+	__xdata int8_t port_configured = cmd_buffer[cmd_words_b[1]] - '1';
+	port_configured = machine.phys_to_log_port[port_configured];
+	if (isnumber(cmd_buffer[cmd_words_b[1] + 1]))  // CPU-port, logical port 9
+		port_configured = (port_configured + 1) * 10 + cmd_buffer[cmd_words_b[1] + 1] - '1';
+	if (port_configured < 0 || port_configured > 9)
+		goto err;
+
+	print_byte(port_configured); write_char('\n');
+
+	if (cmd_compare(2, "show")) {
+		members = port_isolation_get(port_configured);
+		for (uint8_t i = 0; i < 10; i++) {
+			if (members & 1) {
+				if (i < 9)
+					write_char(machine.log_to_phys_port[i] + '0');
+				else
+					print_string("CPU");
+				write_char(' ');
+			}
+			members >>= 1;
+		}
+		return;
+        }
+
+	if (cmd_compare(2, "off")) {
+		for (uint8_t i = machine.min_port; i < machine.max_port; i++)
+			members |= ((uint16_t)1) << i;
+		members |= 0x200; // CPU-port
+		port_isolate(port_configured, members);
+		return;
+	}
+
+	uint8_t w = 2;
+	while (cmd_words_b[w] > 0) {
+		__xdata uint8_t port;
+		if (isnumber(cmd_buffer[cmd_words_b[w]])) {
+			port = cmd_buffer[cmd_words_b[w]] - '1';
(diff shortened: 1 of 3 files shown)


TylerDurden-23:
@logicog Was it intended, that this PR did contain the CLI extension and the keepLink KP-9000-9XHML-X V2.2 support?

logicog:
No, that was a mistake. I updated the PR to only contain the port-isolation-related code.

cmd_parser.c
```diff
@@ -371,6 +371,49 @@ void parse_vlan(void)
...
 
+
+void parse_isolate(void)
+{
+	__xdata uint16_t members;
+
+	print_string("\nISOLATE ");
+	__xdata int8_t port_configured = cmd_buffer[cmd_words_b[1]] - '1';
+	port_configured = machine.phys_to_log_port[port_configured];
+	if (port_configured < 0 || port_configured > 9)
```
TylerDurden-23, line 388: The upper bondary check is not working:

> isolate 10 show

ISOLATE 000x03ff
> isolate 11 show

ISOLATE 000x03ff
>

cmd_parser.c
```diff
@@ -371,6 +371,49 @@ void parse_vlan(void)
...
+		goto err;
+
+	print_byte(port_configured);
```
TylerDurden-23: Does it really make sense to output the physical port number? All other commands output the logical port number.

cmd_parser.c
```diff
@@ -371,6 +371,49 @@ void parse_vlan(void)
...
+
+	uint8_t w = 2;
+	while (cmd_words_b[w] > 0) {
+		__xdata uint8_t port;
+		if (isnumber(cmd_buffer[cmd_words_b[w]])) {
+			port = cmd_buffer[cmd_words_b[w]] - '1';
+			if (isnumber(cmd_buffer[cmd_words_b[w] + 1])) {
+				port = (port + 1) * 10 + cmd_buffer[cmd_words_b[w] + 1] - '1';
+				if (cmd_buffer[cmd_words_b[w] + 2] == 't')
+					vlan_settings.tagged |= ((uint16_t)1) << port;
```
TylerDurden-23: Wouldn't the isolated ports also have to be translated from logical to physical?

cmd_parser.c
logicog: Will fix.

cmd_parser.c
logicog: Not all the commands output logical port numbers. E.g. the stat and the EEE commands use the physical port numbers. The VLAN-related commands mostly use logical port numbers, however. Personally, I think we should move to physical ports everywhere, because we are getting more and more "users" to use the CLI and the difference between logical and physical can be rather confusing (it still confuses me every time I write code related to ports...).

cmd_parser.c
logicog: I think I pushed an old version here, the reference to vlan is a copy and paste error. It was supposed to read:
```
		if (isnumber(cmd_buffer[cmd_words_b[w]])) {
			port = cmd_buffer[cmd_words_b[w]] - '1';
			if (isnumber(cmd_buffer[cmd_words_b[w] + 1])) {
				port = (port + 1) * 10 + cmd_buffer[cmd_words_b[w] + 1] - '1'; // logical port
				if (port != 9) // CPU port is logical port 9
					goto err;
			} else {
				port = machine.phys_to_log_port[port];
				if (port < machine.min_port || port > machine.max_port)
					goto err;
			}
			members |= ((uint16_t)1) << port;
		}
```
I will update the PR accordingly. Does this address your concern?

logicog:
Updated pr and rebased.

cmd_parser.c
TylerDurden-23: @logicog With the rebased PR - if I simply enter "isolate" ... it seems to activate a more or less random isolation configuration. Expected would be an error output.

logicog:
I had completely messed up this PR. I somehow "lost" a fix to the commits to the command parser that was already addressing your issues, so I did not understand where they came from. I worked with a different development machine last week and forgot to push the last changes, it seems.
I believe I have everything now as it should be. Could you try again, @TylerDurden-23 ? Sorry for the mess!

TylerDurden-23:
@logicog The show output looks much better now.
There is still an oddity: When restarting and entering "isolate" - it activates isolation for port 1 or 9.
If you enter "isolate 1 show" first and then enter "isolate" - it gives the correct error message.
Looks like a problem with variable initialization.

logicog:
I believe I fixed this behavior by adding a check on the number of arguments present. Can you check again?

TylerDurden-23:
@logicog Still the same behavior:
image

logicog:
That should not be possible because of:
```
void parse_isolate(void)
{
	__xdata uint16_t members = 0;

	if (cmd_words_b[3] <= 0)
		goto err;

	print_string("\nISOLATE ");
```
I was also not able to reproduce the behavior you see, so the check for the number of arguments was just something that is anyway a good idea. Still, I do not understand how what you see is possible.

You seem to have a long list of commands in `config.txt`, could you try with just the bare minimum?

TylerDurden-23:
@logicog It seems indeed to be related to prior commands from the startup - but it still seems to be a real bug:
The command_parser seems only to clean up the token store at position 1: 
	cmd_words_b[0] = -1;
But not the positions further behind.

It should loop over the array and clean everything from 0 to N_WORDS

image

logicog:
Of course! Stupid me...

TylerDurden-23 (approved):
All looks fine now!

Issue #171, opened 2026-03-21
Idea: description of ports.


JanKiel2000:
Would it be possible to add descriptions to individual ports to identify what they are connected to?

feelfree69:
Implementation done in #200 .

PR #200, opened 2026-04-13
Port names

feelfree69:
This implements #171 

The port command has been extended, now also assigning a name is possible:
```
> port 1 name Desktop Computer
Logical Port: 04

Name set to: "Desktop Computer"
```
Removing an assigned name is done leaving the name empty:
```
> port 1 name
Logical Port: 04

Name set to: ""
```

html/.js has been implemented entirely by MS Copilot - no warranties ;-)

grafik

### DIFF
--- a/cmd_parser.h
+++ b/cmd_parser.h
@@ -7,6 +7,7 @@
 
 extern __xdata uint8_t cmd_buffer[CMD_BUF_SIZE];
 extern __xdata uint8_t cmd_available;
+extern __xdata char port_names[9][PORT_NAME_SIZE];
 
 uint8_t cmd_tokenize(void) __banked;
 void cmd_parser(void) __banked;
--- a/cmd_parser.h
+++ b/cmd_parser.h
@@ -7,7 +7,6 @@
 
 extern __xdata uint8_t cmd_buffer[CMD_BUF_SIZE];
 extern __xdata uint8_t cmd_available;
-extern __xdata char port_names[9][PORT_NAME_SIZE];
 
 uint8_t cmd_tokenize(void) __banked;
 void cmd_parser(void) __banked;
--- a/rtl837x_phy.c
+++ b/rtl837x_phy.c
@@ -367,12 +367,15 @@ void phy_show(uint8_t port) __banked
 		print_string("5G");
 		break;
 	default:
-		print_string("10M");
+		print_string("Down");
 	}
-	if (v & 0x8)
-		print_string(" full duplex");
-	else
-		print_string(" half duplex");
+
+	if ( (((v & 0x0600) >> 7) | ((v & 0x0030) >> 4)) <= 6) { // Link is up
+		if (v & 0x8)
+			print_string(" full duplex");
+		else
+			print_string(" half duplex");
+	}
 
 	phy_read(port,  PHY_MMD_AN, PHY_ANEG_CTRL);
 	v = SFR_DATA_U16;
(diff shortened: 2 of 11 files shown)


rtl837x_phy.c
```diff
@@ -367,12 +367,15 @@ void phy_show(uint8_t port) __banked
 		print_string("5G");
 		break;
 	default:
-		print_string("10M");
+		print_string("Down");
```
logicog, line 370: You do not mention in the PR what you are doing, here, this is not linked to the port names, I assume? Could you move this into a different commit, possibly with a commit message that says what it is going on?

cmd_parser.h
```diff
@@ -7,6 +7,7 @@
 
 extern __xdata uint8_t cmd_buffer[CMD_BUF_SIZE];
 extern __xdata uint8_t cmd_available;
+extern __xdata char port_names[9][PORT_NAME_SIZE];
```
logicog: I think this should go into `rtl837x_common.h`, that way you do not need to include `cmd_parser.h` into `page_impl.c`. We are doing this already for e.g. for the extern declaration of of `sbuf`.

logicog (commented):
Cool idea! Only minor comments...

logicog:
> html/.js has been implemented entirely by MS Copilot - no warranties ;-)

Someone like writing JS as much as I do... But it looks good!

cmd_parser.h
feelfree69: Done.

rtl837x_phy.c
feelfree69: You're right, it has no direct relation to the port names.
I added the output of the port-names also to the `port <port> show` command. 
And then I realized that the output of ports with link down shows "10M", which is obviously wrong.

I'm afraid that my git know-how is not good enough to remove that from the original commit, Adding a commit which reverts this change and then adding it again with a 2nd commit is not very useful, is it?

logicog: OK, let's leave it like that. Just for the record, or next time, this is how I would do it:
```
cp rtl837x_phy.c rtl837x_phy.c_final
git checkout <commit-id before you edited rtl837x_phy.c>
cp rtl837x_phy.c rtl837x_phy.c_before
git checkout <the branch you use to push this PR> # you are now back to where you started with git
cp rtl837x_phy.c_before rtl837x_phy.c
git add rtl837x_phy.c
git commit -m "fix"
git rebase -i HEAD~5 (or less, but needs to include the commit with rtl837x_phy.c)
# now move the fix patch as a fix (line needs to begin with f) directly after the patch with rtl837x_phy.c
# At this point you have reverted the changes to rtl837x_phy.c 
mv rtl837x_phy.c_final  rtl837x_phy.c
git add rtl837x_phy.c
git commit -m "Fix port link-state display" # changes to rtl837x_phy.c are now in a separate commit
git push -f origin <your branch>
```

logicog:
I still need to test-drive this. Will do so tomorrow...

logicog:
Do you think it would be possible to hide the name-line in the tool-tips when the name is empty?
image

I think it does not look too nice, and normally I am rather blind on the designers' eyes...

feelfree69:
Should be fixed now. Thanks Copilot!
grafik

logicog:
Testing now...

logicog:
That thing is scary... Everything works! Good to merge. Very cool stuff, thanks so much for sharing!

TylerDurden-23:
How do you set the port name from the web interface?

feelfree69:
This is not implemented.

Releated: 
I may have missed/not understood some fundamental basics about the Web-If:
I can configure all type of settings and apply them. But how do I store them Non-Volatile? "Save Settings to Flash" has never worked for me. So the Web-If is good for checking status, playing around, but in the end I have to configure it with the Startup Config via command-lines anyway?

PR #201, opened 2026-04-16
Clean up loop in cmd_compare()

logicog:
The for loop in cmd_compare was getting a bit difficult to read. Replace it by a straight-forward while loop.

### DIFF
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -95,20 +95,20 @@ uint8_t cmd_compare(uint8_t start, uint8_t * __code cmd)
 	if ((start > 0) && (cmd_words_b[start] <= 0) )// nothing on this word -> no match
 		return 0;
 
-	signed char i;
 	signed char j = 0;
-	for (i = cmd_words_b[start]; i != cmd_words_b[start + 1] && cmd_buffer[i] != ' '; i++) {
+	signed char i = cmd_words_b[start];
+	while (i != cmd_words_b[start + 1]) {
+		if (cmd_buffer[i] == ' ')
+			break;
 		i &= CMD_BUF_SIZE - 1;
-//		print_byte(i); write_char(':'); print_byte(j); write_char('#'); print_string("\n");
-//		write_char('>'); write_char(cmd[j]); write_char('-'); write_char(cmd_buffer[i]); print_string("\n");
 		if (!cmd[j]) // end of command reached, but cmd_buffer has more characters, so no match
 			return 0;
 		if (cmd_buffer[i] != cmd[j++])
 			break;
+		i++;
 	}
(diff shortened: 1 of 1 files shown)


vDorst:
I would like to propose this version, which is also smaller +100 bytes.
Because `cmd` is guaranteed by the compiler to be NULL-terminated, we can make use of that to ensure the loop always ends.
So we don't need to know when the next words starts.
Because of the wrapping buffer we could technically can have a positive hit.
But I think that is unlikely because `cmd` is hardcoded static strings and not controlled by the user.

```c
uint8_t cmd_compare(uint8_t start, __code uint8_t * cmd)
{
	signed char i = cmd_words_b[start];
	if ((start > 0) && (i <= 0)) // nothing on this word -> no match
		return 0;

	uint8_t j = 0;

	while (1) {
		uint8_t c = cmd[j];
		
		// This ensures that we always read inside `cmd_buffer` if we try to match a long command.
		i &= CMD_BUF_SIZE - 1;
		uint8_t b = cmd_buffer[i];
		
		// cmd is guaranteed to be NULL-terminated.
        if (c == '\0') {
            if ((b == ' ') || (b == '\0')) {
				// Match
                return 1;
            }
            break;
        }
        if (b != c) {
            break;
        }

        j += 1;
        i += 1;
	}

	// No match
	return 0;
}
```

```patch
29c29
---
>    ROM/EPROM/FLASH  0x0000   0x2bfe9  88517    16777216
```



Or with the bufer overflow check:
```c
uint8_t cmd_compare(uint8_t start, __code uint8_t * cmd)
{
	signed char i = cmd_words_b[start];
	if (i < 0) // nothing on this word -> no match
		return 0;

	uint8_t j = 0;

	do {
		uint8_t c = cmd[j];
		uint8_t b = cmd_buffer[i];

		// cmd is garanteerd to be NULL-terminated.
        if (c == '\0') {
            if ((b == ' ') || (b == '\0')) {
				// Match
                return 1;
            }
            break;
        }
        if (b != c) {
            break;
        }

		j += 1;
        i += 1;
	} while (i >= 0);

	// No match
	return 0;
}
```

logicog:
Nice code! I will test tomorrow both versions.

vDorst:
I am working on the parser improvements. See https://github.com/vDorst/RTLPlayground/tree/refactor_cmd_parsing with has this code in it.
But still has some bugs with history.

Also use some Rust magic to test some parts of the implementation see https://github.com/vDorst/clib/blob/main/src/lib.rs#L490

logicog:
@vDorst : Maybe it would be better if you opened a PR for your much more sophisticated solution, while we close this PR. The little code here from my side was merely work I did while trying to figure out what was going on with #169 , the solution you are proposing will bring the command parsing forward by a much bigger step.

logicog:
Closing because superseded by #203

PR #202, opened 2026-04-18
Webif

feelfree69:
As I have no clue about html and Javascript, that's why I thought I let Copilot do some work.
It's not meant for immediate merge, but at least I want to share the results.
The improved User-Experience when uploading a new firmware image has some potential. With my Firefox on Windows the Upload and Redirect to Login-Page works flawlessly.
Well, the colors might not be the first choice.... 

grafik

grafik

What I'm not sure about: I have the feeling that with the "modern look", also the Performance has increased, maybe due to better caching strategies in the browser? Most pages load almost immediately now, where with the old design there were frequent delayed loadings especially at the port-icons at the top of the page.

### DIFF
--- a/html/style.css
+++ b/html/style.css
@@ -98,6 +98,16 @@ input[type=file] {
   width: 70%;
 }
 
+.status-message {
+  margin-top: 1em;
+  padding: 12px 16px;
+  background: rgba(255,255,255,0.95);
+  color: #222;
+  border-radius: 12px;
+  box-shadow: 0 3px 6px rgba(0,0,0,0.12);
+  font-size: 0.95rem;
+}
+
 .action {
   width: auto;
   margin-right: 10px;
--- a/html/style.css
+++ b/html/style.css
@@ -1,53 +1,107 @@
-h1, h2, h3 {
-  color: #226;
+/* Modernes Design für das Webinterface */
+
+body {
+  font-family: 'Segoe UI', Tahoma, Geneva, Verdana, sans-serif;
+  background: linear-gradient(135deg, #667eea 0%, #764ba2 100%);
+  color: #333;
+  margin: 0;
+  padding: 0;
+  min-height: 100vh;
 }
+
+h1, h2, h3 {
+  color: #fff;
+  text-shadow: 1px 1px 2px rgba(0,0,0,0.5);
+  margin-bottom: 20px;
+}
+
 ul {
   list-style-type: none;
   margin: 0;
   padding: 0;
-  width: 12%;
-  height: calc(100% - 50px);
+  width: 15%;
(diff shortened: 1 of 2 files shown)


logicog:
Thanks for sharing this, the upload can definitely profit from a better UI and the redirect is currently indeed broken! A fix for Chrome would however need changes in the C-backend handling the POST request, at least that is what the http-experts say.

> What I'm not sure about: I have the feeling that with the "modern look", also the Performance has increased, maybe due to better caching strategies in the browser? Most pages load almost immediately now, where with the old design there were frequent delayed loadings especially at the port-icons at the top of the page.

I do not see anything touching the ports or related to concurrency in the javascript embedded in the html pages of the code, but if you are comparing the situation to about 1 month ago, then there was a considerable improvement of the loading of the pages with PR #151 . 
If you want to give your hunch some numbers to back it up, have a look at timing diagrams in the debug console of Firefox, like the one posted in that PR.

For proceeding with this request: maybe try to disentangle the design part, i.e. the stuff in the stylesheets, from the functional part like the redirect and the information bar so that we can get the functionality in without necessarily a change of the entire design. Ideally, we anyway want to have different "skins" for the UI so that users can have a UI they like design-wise as design is always rather a personal thing.

feelfree69:
> maybe try to disentangle the design part, i.e. the stuff in the stylesheets, from the functional part like the redirect and the information bar

Both parts are already separated by the 2 commits.

vDorst:
Looks nice.
I was worried that it may try to pull an external library from the web. But lucky it is 
not.

I think we can improve the speed a bit by removing some `__gptrget` call in the IP-stack.
See also https://github.com/logicog/RTLPlayground/issues/169#issuecomment-4121580463
`__gptrget` call is used when the compiler doesn't know which memory is used for the variable.
If we can improve that, it saves overhead, memory and speed.

logicog:
I was not aware that uIP is still using gptr, that needs to go. Still, the `__gptrget` stuff is probably a bit off-topic for this PR. Can we get merely the UI-part in this PR without the change in style? Also, so far all the javascript was always in the .js files, now it is mixed into the html files. I think it would be great to keep that separation, for example if we want the UI to be translated into other languages, we would only need to have different html files for a different language while the .js stay the same as they are just functionality. (OK, some of the .js have text inside, but that could be taken from the .html files, too, instead of coding it into the .js.)

feelfree69:
as written before, i only wanted to share the results of my 1hr playing with copilot. i did not take a single look at the code, and i have no motivation to do so... sorry

feelfree69:
Will close the PR as it was never intended to merge, and as there is already a far more better UI in work (#220).

PR #203, opened 2026-04-18
Refactor cmd parsing

vDorst:
This PR refactor the cmd buffer parsing.

Convert the `cmd_words_b` in to a fixed-capacity vector/bounded buffer.
`cmd_words_len` represents the number of arguments found, the code already did this with `word`. But how it global and useful to check if we have enough arguments for a specific command.

`cmd_words_b` now has only one purpose instead of two.
`cmd_words_b`-entry is only valid when it is inside the `cmd_words_len`-range.
It is converted to `uint8_t`, his means that the 127-bytes limit can be changed to 255-bytes with `CMD_BUF_SIZE`.

`cmd_words_b[]` is reduced from `CMD_BUF_SIZE`/127-bytes to just 15-bytes/entries and is controlled by `N_WORDS`. Technically limit was 64-entries anyway. Reducing entries also saves a lot of xdata.

Refactor the other commands to make use of `cmd_words_len` and `cmd_words_b` now has only one purpose now.

Refactor some functions that made use of that `cmd_words_b` last valid entry was cmd_buffer length.

I hope that I found all the `cmd_words_b[..]` comparison, used `rg -tc "cmd_words_b\[[^]]+\] ?[=<>]"` to find them. Because we don't need to check the index anymore.

Note: this PR supersedes #201.

I still thinking we should rename `cmd_words_b` and `cmd_words_len` to a better name.

Some implementations are tested in Rust, see https://github.com/vDorst/clib/blob/main/src/lib.rs#L490

Codesize and xdata reduction.
```patch
<    EXTERNAL RAM     0x0001   0x2623    9763    16777216
---
>    EXTERNAL RAM     0x0001   0x25b1    9649    16777216
>    ROM/EPROM/FLASH  0x0000   0x2bd55  87857    16777216
```

### DIFF
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -1114,7 +1114,7 @@ void parse_bw(void)
 	__xdata uint8_t port;
 	__xdata uint32_t bw = 0;
 
-	if (cmd_words_b[3] < 0) // Check for at least 2 arguments
+	if (cmd_words_len < 2) // Check for at least 2 arguments
 		goto err;
 
 	port = cmd_buffer[cmd_words_b[2]] - '1';
@@ -1128,7 +1128,7 @@ void parse_bw(void)
 		return;
 	}
 
-	if (cmd_words_b[4] < 0) // Check for at least 3 arguments
+	if (cmd_words_len < 4) // Check for at least 4 arguments
 		goto err;
 
 	if (cmd_compare(3, "drop")) {
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -1187,7 +1187,7 @@ err:
 // Parse command into words
 // cmd_words_len contains the number of words found.
 // cmd_words_b[] contains only start of a word offset.
-uint8_t cmd_tokenize(void) __banked
+void cmd_tokenize(void) __banked
 {
 #ifdef DEBUG
 	print_string("Tokenizing command\n");
@@ -1211,7 +1211,7 @@ uint8_t cmd_tokenize(void) __banked
 
 		if (line_ptr == CMD_BUF_SIZE - 1) {
 			err_status = ERR_CMD_TOO_LONG;
-			return 1;
+			return;
 		}
 
 		if (is_white && c != ' ') {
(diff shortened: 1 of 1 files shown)


cmd_parser.c
```diff
@@ -1175,33 +1195,41 @@ uint8_t cmd_tokenize(void) __banked
...
+
+		if (is_white && c != ' ') {
 			is_white = 0;
+
 			cmd_words_b[word++] = line_ptr;
-		}
-		if (cmd_buffer[line_ptr] == ' ')
+			if (word >= N_WORDS) {
+				cmd_words_len = 0;
+				print_string("\ntoo many arguments, truncated");
```
logicog: I think the error message is wrong, here. The arguments are not merely truncated, the command no longer parses:
```
> stat 1 2 3 4 5 6 7 8 9 10 11 12 13

5       SFP IN  10G     0x0000000e      0x00000000      0x00000026      0x00000000
1       On      2.5G    0x00000048      0x00000000      0x00000014      0x00000000
6       SFP IN  1000M   0x00000029      0x00000000      0x00000022      0x00000000

> stat 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16

too many arguments, truncated
> 
```
So the error message should be something like "Syntax error: too many arguments"

logicog (commented):
Only this very small minor issue. Otherwise really bit improvements, the code reads much easier, now.

cmd_parser.c
vDorst: Yes you are right. Passing stops when too many arguments are found.

Looking with `rg -tc err_status`, it is not used in the code base.
I have the idea `err_status` has lost it's purpose of we missing some code?

logicog: > I have the idea `err_status` has lost it's purpose of we missing some code?

We are just missing code that deals with the error number. This would be for the script execution of `config.txt` at boot-time to identify that a command failed. Currently only too long lines stop execution. The loop could also stop when a command has a syntax error.

vDorst: Updated the error message.
refactor `cmd_tokenize()` a bit. But didn't add err_status handling.
Maybe for an other PR?

logicog: > Maybe for an other PR?

Indeed.

logicog:
Thanks for the cleanup!

PR #204, opened 2026-04-18
Remove `__gptrput()` and `__gptrget()` calls

vDorst:
When no memory type is specified in a call-argument to the reference type, sdcc is
using a helper functions, `__gptrput()` and `__gptrget()`, to access that location.
This can be useful to make functions generic.
But sdcc is using an extra call-register to tell the helper functions which memory-type is used.
This registers must also be preserved until all access to that location is done.

To find all the  `__gptrput()` and `__gptrget()`.
Compile the code as normal.
Then use this command `rg  "__gptr" output/` to see if there is a hit.
Look in the .lst file to see in which file and location this occurs.

note: `cmd_compare()` is not fixed, will be fixed in #201 

Code-size saved: +650-bytes
```patch
<    EXTERNAL RAM     0x0001   0x2623    9763    16777216
---
>    EXTERNAL RAM     0x0001   0x2613    9747    16777216
>    ROM/EPROM/FLASH  0x0000   0x2bf1b  88018    16777216
```

### DIFF
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -169,7 +169,7 @@ uint8_t atoi_hex(uint8_t idx)
 }
 
 
-uint8_t atoi_byte(uint8_t __xdata *out, uint8_t idx)
+uint8_t atoi_byte(__xdata uint8_t *out, uint8_t idx)
 {
 	uint8_t err = 1;
 	uint8_t num = 0;
@@ -185,7 +185,7 @@ uint8_t atoi_byte(uint8_t __xdata *out, uint8_t idx)
 }
 
 
-uint8_t atoi_short(uint16_t __xdata *vlan, uint8_t idx)
+uint8_t atoi_short(__xdata uint16_t *vlan, uint8_t idx)
 {
 	uint8_t err = 1;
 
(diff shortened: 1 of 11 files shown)


vDorst:
with trial-and-error, I found that, passing some arguments also via __xdata, can reduce sram usage. I have the feeling that sdcc puts some arguments on a fix memory location so it knows the location and don't have to pass around the argument value/location when many large arguments are passed or registers are needed for the function itself.
Not sure what the threshold is, and that this still can break in some combinations.

vDorst:
Although it are great "free" improvements, many function are not used much in the hot-path.
So speed gain is minimal.
Because we get 4 extra sram memory, I was trying to move `sfr_data` to sram, because `sfr_data` is used a lot, so I think that is justified to put `sfr_data` in sram.
But it goes out-of-memory.

logicog:
This also contains the commits in PR #203. I  will start going through that now. Since the changes are not interrelated, it should be sufficient to re-base to main once #203 is merged.

vDorst:
Oeps, I remove the merge.

logicog:
I test-drove it and verified the .asm: the web-interface works and there are no more references to gptr left. My understanding is that the speed-improvements in uip are not that relevant, as in practice this is limited by the polling behaviour of the web-interface and the single thread serving it. What matters is the reduction in SRAM usage and there every byte counts.

Great idea and work!

Will merge.

PR #206, opened 2026-04-21
Fix cmd parsing

### LINKED ISSUE #205: cmd parsing broken?
Just built the newest firmware from main. 
While the default config is executed fine, commands on the serial shell are not recognized at all. I get no output at all.
Anybody else with the same problem?

```
Setting ip: 192.168.10.247
Setting gw: 192.168.10.1
Setting netmask: 255.255.255.0

>
<new link: 2000020000, was 0000000000>

<SFP-RX OK>  Slot: 1

<MODULE INSERTED>  Slot: 2  Rate: 67  Encoding: 00  Module: OEM             SFP-H10G-CU1M   03

sds_config sds: 00, mode: 1a

RTL837X_REG_SDS_MODES: 0x000003fa

<SFP-RX OK>  Slot: 2


>

> ip

> eee
```

vDorst:
Fix #205.

### DIFF
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -200,6 +200,7 @@ uint8_t atoi_byte(__xdata uint8_t *out, uint8_t idx)
 uint8_t atoi_short(__xdata uint16_t *vlan, uint8_t idx)
 {
 	uint8_t err = 1;
+	*vlan = 0;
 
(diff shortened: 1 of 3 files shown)


logicog:
Tested and works. Will merge.

PR #207, opened 2026-04-26
ChaCha20 implementation for encryption

logicog:
Add an implementation in 8051 assembler for the generation of a ChaCha20 keystream and an encryption function using the keystream in C.

A test function using the example from RFC7539 is provided in chacha20_test.

Keys, nonce, and block counter and the pointers to the input plaintext, its length and the output pointer are provided via a struct chacha20, which also holds the working state of the ChaCha20 algorithm.

ChaCha20 can be used together with Poly1305 in the Authenticated Encryption with Associated Data (AEAD) implementation CHACHA20POLY1305, which is used e.g. in Wireguard and in TLS, which both Firefox and Chrome support it. In a next step, Poly1305 will be implemented, an 8-bit implementation exists. The final step for use in Wireguard/TLS, will require an implementation for DiffieHellman key exchange using Curve25519, for which an 8-bit microcontroller implementation also exists.

To test, call `chacha20_test()` in `rtlplayground.c`

The output will be:
```
>
Encrypting...
Cypertext:
6e 2e 35 9a 25 68 f9 80 41 ba 07 28 dd 0d 69 81 e9 7e 7a ec 1d 43 60 c2 0a 27 af cc fd 9f ae 0b f9 1b 65 c5 52 47 33 ab 8f 59 3d ab cd 62 b3 57 16 39 d6 24 e6 51 52 ab 8f 53 0c 35 9f 08 61 d8 07 ca 0d bf 50 0d 6a 61 56 a3 8e 08 8a 22 b6 5e 52 bc 51 4d 16 cc f8 06 81 8c e9 1a b7 79 37 36 5a f9 0b bf 74 a3 5b e6 b4 0b 8e ed f2 78 5e 42 87 4d 
```
Currently, no applications are provided, but the code could be used to encrypt e.g. the password.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,11 +26,13 @@ create_build_dir:
 	mkdir -p $(BUILDDIR)
 	mkdir -p $(BUILDDIR)/uip
 	mkdir -p $(BUILDDIR)/httpd
+	mkdir -p $(BUILDDIR)/crypto
 
 SRCS = rtlplayground.c rtl837x_flash.c rtl837x_leds.c rtl837x_phy.c rtl837x_port.c cmd_parser.c html_data.c rtl837x_igmp.c
 SRCS += rtl837x_stp.c rtl837x_pins.c dhcp.c machine.c cmd_editor.c rtl837x_bandwidth.c rtl837x_init.c
 SRCS += uip/timer.c uip/uip.c uip/uip_arp.c uip/uiplib.c uip/uip-fw.c uip/uip-neighbor.c uip/uip-split.c
 SRCS += httpd/httpd.c httpd/page_impl.c
+SRCS += crypto/chacha20.c
 OBJS = ${SRCS:%.c=$(BUILDDIR)/%.rel}
 DEPS := ${SRCS:%.c=$(BUILDDIR)/%.d}
 HTML := $(shell find $(html) -name '*.js' -or -name '*.html' -or -name '*.svg')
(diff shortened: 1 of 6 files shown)


DrDoof:
I had a look at this and ran it against the RFC vector, and I think there is a bug in `_chacha_count` that is worth fixing before this gets used for anything longer than the test.

On the 8051 `inc a` doesn't touch the carry flag, so the three `jnc chacha_count_done` end up checking whatever carry was there before the call. If carry is clear, only the lowest byte of the counter ever changes. After 0xff it wraps to 0x00, and the next block uses counter 1 again, so you get the same keystream as the first block. If carry is set, all four bytes go up at once and even the first block runs with 0x01010101. `inc a` leaves A at zero exactly when the byte overflows, so `jnz` should do what you meant:

```
	inc	a
	movx	@dptr, a
	jnz	chacha_count_done
```

I didn't want to guess, so I assembled chacha_8051.asm with sdas8051, ran it in a small 8051 emulator, calling the routines the same way chacha20_encrypt() does, and compared the output with a ChaCha20 in Python. With the code as it is, the RFC 7539 vector only comes out right when carry is clear going in, and five blocks starting at counter 0xfe break from the third block on. With `jnz` everything matches, also when starting at 0xfffe. In chacha20_encrypt() carry is always clear from the second block on, because the `chacha20.length -= ...` just before the jump back can't borrow. For the first block it depends on what print_string() leaves behind, and I guess that's why the test worked on your board.

A few smaller things I noticed while moving it onto current main to build it for the SWTGW218AS:

The machine.h change looks like it came along from local testing. Makefile, machine.h and rtlplayground.c conflict with main by now, but the move is small: the crypto directory in create_build_dir, the SRCS line and chacha_8051.rel next to crtbank.rel. With that it builds fine.

Nothing calls the code yet, but the image still grows by 1218 bytes of common code (548 in HOME from the assembly, 468 in CSEG and 202 in CONST) and by 512 bytes of xdata for the two test buffers. Maybe chacha20_test() and the buffers could sit behind an #ifdef, or the C part could go into a bank, until something actually uses it.

_chacha_test_1, _chacha_qr_r and print_regs look like debugging leftovers. print_regs has no ret and falls straight into chacha_plus_xor, so I'd drop it before someone calls it by accident. And in chacha_qr the rotated B is written twice, first with `mov dpl, r3` right after rol_b and then again with `mov dpl, r1` after swapping back, so the second store_32 isn't needed.

On the `__at(0x7000)`, I described in #443 how the linker treats it. If it moved to 0xBF00, `--xram-size` could guard it too.

If you push a new version, I can test it on hardware.

logicog:
Was travelling, is on my todo-list. The observations are all valid. Will fix and rebase and provide new commits in the next days.

logicog:
Rebased and fixed the counting issue as suggested.

PR #209, opened 2026-05-01
Add Steamemo IG204 V1 support

hlyi:
Add support for "Steamemo IG204 V1" unmanaged switch.


logicog:
Sorry for looking only now into this, I had planned to do this already yesterday, and then the German DNS broke down... This is great stuff, we have been waiting for a configuration generator from a dump for quite some time!
Still, I believe you are putting 2 different things into this PR, the Steamemo support and the configuration generator. Could you create a second PR just for the configuration stuff? It would also be great to have some documentation on how to use the generator. Maybe with an example?

hlyi:
Could you please check the code? This PR is only for Steamemo IG204 V1 device support. I haven't made the PR request for tools ( as it is not ready for prime time). I think the commit comments you saw was from a different branch. Thanks,

logicog:
Code looks very good! Could you add some documentation for the Steamemo device, such as an .md file in doc/devices, plus pictures of the PCB in doc/devices/photos/ These pictures can help a lot when people need to decide whether the device they have is actually already supported, but e.g. under a different brand name. Have a look at the examples. You can take the photos with your smartphone, the image quality is more than enough.

What would be really great would be if you could clean up this PR a bit: you can use `git rebase -i HEAD~5` (maybe 4 is good enough, maybe you need 6, but try 5) to include a later commit into a previous one (with `f`, rewrite the commit message with `r`or even re-order the commits). Then, you can update this PR with the new commits, wherein only the clean commits are part of the PR, no longer the entire history. To update the PR works with `git push -f <your_name_for_this_repository> <your_branch>`. I believe, your branch is just `pr`. The idea would be, to just have 1 commit with the doc, pictures and the commit to machine.h and machine.c. If git makes trouble, do not hesitate to ask for help, here.

feelfree69:
> What would be really great would be if you could clean up this PR a bit:

I know @logicog is an git expert.... but with such a small PR, I would simply create a new branch from main, copy the changed 2 files in there, commit and then add the docu with an extra commit

Then you're done with that:
> The idea would be, to just have 1 commit with the doc, pictures and the commit to machine.h and machine.c

logicog:
Oops, I think you added several of the latest merges into this PR. There are 2 ways to fix this:
1. checkout main and pull to latest head from upstream, create a new branch and cherry-pick the commits you want to use into that branch. Then force-push into the existing PR. More work, but rather safe.
2. checkout your branch for the PR. `git fetch origin/upstream`, and `git rebase -i origin/main` will play-back your changes on top of the latest HEAD. Easier, but also easier to mess up...

hlyi:
Fixed the PR based upon your commendation option 1. Please check.

logicog:
Do you have a description of the device in a .md file? And everyone loves pictures of the PCB, because that way it is easier to check whether 2 devices are actually identical .

hlyi:
md and jpg files have been added.

hlyi:
Anything for me to work on for this PR?

logicog:
Sorry, I overlooked this. Do you have a photo of the bottom side of the PCP? Could you add that. Otherwise it looks great!

hlyi:
Sorry, the switch is already deployed. Is it okay to accept the PR without bottom side picture?

logicog:
Perfect! Thanks a lot for your contribution.

PR #210, opened 2026-05-01
Fix chrome uploads

logicog:
Chrome sends upload requests using POST with multipart/form-data content type in multiple packets for the header part of the form-data.
    
Introduce a TSTATE_MULTIPART for the httpd server states that denotes that so far only a part of the multipart header has been transmitted. Once the full header has been transmitted, we change to TSTATE_POST as for Firefox which sends all the multipart header in one piece.
    
The main further change required then is to make sure that the parsing of the initial part of the multipart request is only parsed once and initially to distinguish between configuration and firmware uploads.

### DIFF
--- a/httpd/httpd.c
+++ b/httpd/httpd.c
@@ -60,11 +60,12 @@ __xdata uint32_t now;
 __xdata uint8_t *timeptr;
 __xdata uint32_t last_session_use;
 
-#define TSTATE_NONE	0
-#define TSTATE_TX	1
-#define TSTATE_ACKED 	2
-#define TSTATE_CLOSED 	3
-#define TSTATE_POST 	4
+#define TSTATE_NONE		0
+#define TSTATE_TX		1
+#define TSTATE_ACKED 		2
+#define TSTATE_CLOSED 		3
+#define TSTATE_POST 		4
+#define TSTATE_MULTIPART	5
 
 extern __xdata uint16_t crc_value;
 __xdata uint16_t crc_final;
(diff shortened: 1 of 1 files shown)


feelfree69:
Made a quick test with Chrome / Windows.
Updating the image worked fine, but after update, the startup config was screwed up.

grafik
grafik

Need to check of this is reproducable.

httpd/httpd.c
```diff
@@ -355,21 +356,45 @@ void handle_post(void)
...
+			if (flash_size < FIRMWARE_UPLOAD_START*2)
+			{
+				print_string("Flash too small for firmware upload!\n");
+				send_bad_request();
+				return;
+			}
+			print_string("Firmware upload started.");
+			uptr = FIRMWARE_UPLOAD_START;
+			verify_crc = 1;
+			max_upload = 1024576;
```
feelfree69, line 387: Just noticed that magic number, and I don't know where it is coming from.
Its 1024MiB minus 24000 Bytes.
Probably 512MiB (size of RTLPlayground.bin) would be correct?

logicog:
I also saw this issue with the startup configuration with Firefox. I thought this was on purpose and that the configuration that had been subsequently configured was the one retained. I hope I did not break this also for Firefox.

logicog:
> Probably 512MiB (size of RTLPlayground.bin) would be correct?

I think you stumpled over why the configuration is overwritten: The default configuration starts at 0006f000, while the current one starts at 00070000. And I think that must not be overwritten. I was wandering how this update can overwrite the configuration, and always suspected an explicit flash_sector_erase() but I guess this can also have the same effect. I will check. And also de-magic this.

feelfree69:
Can confirm that corruption of config area is reproducable, also with Firefox. 
Went back to main branch - no issue anymore.
Still wondering what causes this.....

logicog:
> Still wondering what causes this.....

I have been interrupting the upload process at various stages all day and then reading out the flash. What I have learned so far is that the current configuration is erased (not overwritten) after the upload is started and before it is finished. I just cannot figure out how this happens. I traced flash_sector_erase() with a printout inside, and that is never called. I do not see any possibility how this can happen. Maybe another unexpected compiler behavior...

logicog:
I finally found it. The configuration was already erased on login(!), because we treated any POST request as a configuration update by default, and login is also a POST request. So the following changes make the difference:
```
diff --git a/httpd/httpd.c b/httpd/httpd.c
index 207ab06..9e33f1a 100644
--- a/httpd/httpd.c
+++ b/httpd/httpd.c
@@ -385,7 +385,7 @@ void handle_post(void)
                        uptr = FIRMWARE_UPLOAD_START;
                        verify_crc = 1;
                        max_upload = 1024576;
-               } else {
+               } else if (is_word(request_path, "config")) {
                        dbg_string("Configuration upload, erasing config mem!\n");
                        uptr = CONFIG_START;
                        verify_crc = 0;
@@ -393,6 +393,7 @@ void handle_post(void)
                        flash_region.addr = CONFIG_START;
                        flash_sector_erase();
                }
+               // Check for other POST requests, which are not multipart, below
        } else {
                dbg_string("Multipart request\n");
        }
```
I updated the PR with the fix.

feelfree69:
Ah, that makes sense! 

Tested again, works like a charm -> ready to merge.

PR #212, opened 2026-05-02
Makefile: Improve filenames of images

feelfree69:
1. `rtlplayground.bin` renamed to reflect MACHINE and git-revision in the filename.
Kept a symbolic link to the latest output for compatibility reasons - makes it also easier for automatic deployment in case one uses additional scripts.

```
~/RTLPlayground$ ls -l output/*.bin
lrwxrwxrwx 1 user user     49 May  2 21:42 output/rtlplayground.bin -> rtlplayground-v0.1.0-09d90d6-dirty-HI_K0402WS.bin
-rw------- 1 user user 524288 May  2 21:42 output/rtlplayground-v0.1.0-09d90d6-dirty-HI_K0402WS.bin
```

2. Output of installer dir renamed `rtlplayground_oem_upgrade.bin` to avoid confusion with `rtlplayground.bin`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,11 +16,21 @@ BUILDDIR = output
 VERSION_HEADER := version.h
 
 ifeq ($(MACHINE),)
+	MACHINE:= $(shell grep "^\s*#define MACHINE_" machine.h | sed "s/^\s*#define MACHINE_//")
 else
 	CC_FLAGS += -DMACHINE_$(MACHINE)
 endif
 
-all: create_build_dir $(VERSION_HEADER) $(SUBDIRS) $(BUILDDIR)/rtlplayground.bin
+GIT_VERSION := $(shell git rev-parse --short HEAD)
+ifeq ($(shell git status --porcelain),) then
+else
+	GIT_VERSION := $(GIT_VERSION)-dirty
+endif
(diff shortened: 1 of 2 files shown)


logicog:
I very much like the improvement of the file names of the images. That is really cool to not mix them up during development. My problem is that my development directory is full of additional stuff like log-output and some other tools. None of that is tracked in git. So my images all end up as being "dirty" although all files used in the build are straight out of git. Would it not be better to mark the image only as "dirty" if tracked files were changed, and not yet committed? 

Now that we have nice image names, would it not be nice if `make clean`did not delete all the images. I was just creating one machine image and then another, but then accidentally did `make clean`. Of course all the previous images were gone. Maybe `clean`could not delete the images themself? And then we could have `make distclean` to do that.

feelfree69:
thanks for your comments and suggestions. Will do that, probably this evening.

feelfree69:
Just added your suggestions.
What I am still unsure about is the "VERSION" variable in the Makefile. It does not bring any benefit. Do we still need it (in the *,bin-filename as well as in the string inside the compiled image)?

logicog:
Perfect! Great work!

PR #213, opened 2026-05-03
Update Makefile to use dynamic build names

dobodu:
Build name are now Machine_Build-Version_Build-Date

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,13 +14,18 @@ SUBDIRSCLEAN=$(addsuffix clean,$(SUBDIRS))
 
 BUILDDIR = output
 VERSION_HEADER := version.h
+MACHINE_NAME = $(shell grep -v "//" machine.h | grep "#define MACHINE" | sed 's/.* MACHINE_//')
+VERSION_NAME = $(shell git rev-parse --short HEAD)
+BUILD_DATE = $(shell date +"%Y-%m-%d_%H-%M-%S")
+BUILDNAME = $(MACHINE_NAME)_$(VERSION_NAME)_$(BUILD_DATE)
+
 
 ifeq ($(MACHINE),)
 else
 	CC_FLAGS += -DMACHINE_$(MACHINE)
 endif
 
-all: create_build_dir $(VERSION_HEADER) $(SUBDIRS) $(BUILDDIR)/rtlplayground.bin
+all: create_build_dir $(VERSION_HEADER) $(SUBDIRS) $(BUILDDIR)/$(BUILDNAME).bin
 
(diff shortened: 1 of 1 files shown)


dobodu:
I'm sorry, just afeter I pulled #213 I've seen #212 we are dealing the same subject 👍

feelfree69:
At least it was not too much work for both of us ;-)

Then let's @logicog and @vDorst decide which way we should go further...

In this PR, I'm missing these aspects (just from reading the diff, not actually tried):
- `make MACHINE=<MY_MACHINE>` won't grab the correct MACHINE name, since only the machine.h is taken into account
- the OEM upgrade is not being touched
- there's no link with a "stable" filename

In my PR, I did not include the build date. Unsure if that brings a benefit.

dobodu:
> In my PR, I did not include the build date. Unsure if that brings a benefit.

Finally, I also think time is unecessary,

For example, the build script i did lead to :

`LIANGUO_ZX_SWTGW215AS_0.1.0_12c98ba_2026-05-03.bin`

logicog:
I have had my own share of submitting what eventually turned out to be competitive pull/patch requests when doing Linux kernel development and that was not a pleasant thing. I would rather see everyone collaborating on a single request. After testing #212 I left some comment there, with further ideas. Maybe you could have a look into that. It is possible for 2 people to collaborate on the same PR, for example by proposing changes in the github web-interface, which then get pulled in by the owner of the PR. Another possibility is by pulling in a commit from a forked version of RTLPlaygound (by done by person A) into an exiting PR (done by person B) by doing cherry-picking in git (by person B).

dobodu:
@logicog We'll make an agreement with @feelfree69 

@feelfree69 : I personnaly prefer to setup the machine.h instead of passing through the machine name to the make CL.
The Makefile I proposed, deals only with standard build, I didn't upgrade the installer Makefile yet. 
I think, we'd better **output directly OEM and non OEM binary versions** with a precise naming, that would be easier for customer to build only once (no more cd installer & make again)... What's your opinion ?

feelfree69:
> I personnaly prefer to setup the machine.h instead of passing through the machine name to the make CL.

You can do that. Both methods do work.

> think, we'd better **output directly OEM and non OEM binary versions** with a precise naming,

As OEM upgrade is only needed once and it is strictly discouraged anyway, I won't do that.
- If you don't have a clamp / flasher, there's no guarantee that it will work. And even if it worked, there's no way back.
- If you *have* a clamp / flasher, you don't need the OEM upgrade image....

PR #214, opened 2026-05-03
Erase flash before writing upload image

### LINKED ISSUE #193: Issue report: Firmware flash failure
Hello, I am experiencing issue with the firmware on my Keeplink KP-9000-9XH-X v3.1 switch:

**Firmware Update Failure**: Whenever I attempt to flash a new firmware, the process consistently ends with a checksum error as shown below:

    Firmware upload started.......................................................
    ...............................................Checksum incorrect!
    Upload to flash done, will reset!
    Initializing Flash controller
    Detecting CPU: RTL8373N
    Software version: v0.1.0-g

After this, the device reboots back into the current version, completely ignoring the new image. I have tested this using the Firefox and chrome browser and attempted to flash both a new build of the current software and the original Keeplink factory firmware. The result is always the same.

feelfree69:
This resolves #193.

### DIFF
--- a/httpd/httpd.c
+++ b/httpd/httpd.c
@@ -14,9 +14,6 @@
 #define SESSION_ID_LENGTH 12
 #define SESSION_TIMEOUT 200
 
-// SPI FLASH MEMORY PAGE SIZE.
-#define FLASHMEM_PAGE_SIZE 0x100
-
 #define CMARK_S 6
 
 #pragma codeseg BANK1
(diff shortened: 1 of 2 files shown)


logicog:
A very smart solution to the issue with the update from installations via OEM installations. Tested and works perfectly! I would have some ideas for cleaning up the current code, though: `FLASHMEM_PAGE_SIZE` is defined in httpd.c, `FLASH_SECTOR_SIZE` is defined in rtl837x_flash.h Maybe it would make sense to have a common prefix (i.e. FLASHMEM_ or FLASH_) for both definitions and have them both in rtl837x_flash. And if we then make sure that FLASH_SECTOR_SIZE is a multiple of the page size, via a preprocessor error, we could guarantee that the check in httpd.c regarding the beginning of a sector always works.

logicog:
Perfect!

Issue #81, opened 2026-01-14
Hi-Source HI-k0801WS


mcaptur:
Will this work as is on this swicth (i will flash with clip) eprom seems to be 2mb so i guess can flash this firmware..

![Image](https://github.com/user-attachments/assets/5af51139-2bce-4d69-8e56-480bf295c698)
![Image](https://github.com/user-attachments/assets/a9624f81-f43b-43b8-b34e-3414cea2c507)
![Image](https://github.com/user-attachments/assets/d76f4ff4-f0f7-4b0d-ad1d-057cdfcea804)
![Image](https://github.com/user-attachments/assets/98192e08-61eb-41e5-9d23-f60b92c2b0fb)

mcaptur:
Thanks as sson as i receive clip i'l backup eeprom and try upload this.. Any modification i need to do to source before compiling (since its 8+1 ports)? if i loose switch no big deal.. I just enjoy playing with this stuff..

vDorst:
In `machine.h`, comment `// #define DEFAULT_8C_1SFP` and uncomment `// #define DEFAULT_8C_1SFP`
Currently I don't think we have a fully working 8x1-switch.
But if it boot fine that you should able to connect to the website on address `http://192.168.10.247` via RJ45 port 5 to 8.
SFP and port 1-4 may not work yet.

Can you also provide pictures from the underside?

mcaptur:
will take picture tomorrow as am out at the moment..

TylerDurden-23:
@feelfree69  
From the photos the PCBs seem to have very similar construction, so the assumption is the config is very similar.
But sure, if this does not work the "Default 8+1" is also worth a try.

mcaptur:
Received spare flash chips and was able to finally test.. here is pull request.. all seems to work fine
https://github.com/logicog/RTLPlayground/pull/216

PR #216, opened 2026-05-07
Add hardware profile and LED fix for Hi-Source HI-k0801WS

mcaptur:
https://github.com/logicog/RTLPlayground/issues/81
Tested, leds and sfp seem to all work fine


logicog:
The code looks perfect. Would it be possible to add the images of the PCB and a brief description of the device such as is done in doc/devices and doc/devices/photos for other switches? I know this is in the original issue, but having it also in the repository to look up would be great: the PCB images help a lot to decide whether a particular device is already supported, because often the same PCB is used by different vendors.

mcaptur:
> The code looks perfect. Would it be possible to add the images of the PCB and a brief description of the device such as is done in doc/devices and doc/devices/photos for other switches? I know this is in the original issue, but having it also in the repository to look up would be great: the PCB images help a lot to decide whether a particular device is already supported, because often the same PCB is used by different vendors.

check now.. sorry for messy commits.. but never used git :)

logicog:
Sweet! Thanks for the nice documentation.

PR #219, opened 2026-05-08
Fix regex and add more command support to config parser

mcaptur:
Changes are rarely saved! this shoudl fix but needs more testing

### DIFF
--- a/html/bandwidth.js
+++ b/html/bandwidth.js
@@ -1,33 +1,33 @@
-const iLayout = '" type="number" maxlength="10" size="10"  onfocus="inputFocus(';
+const iLayout = '" type="number" maxlength="10" size="10" onfocus="inputFocus(';
+
 function createBW() {
   var tbl = document.getElementById('bwtable');
-  const limit = '<input type="checkbox" id="limit_port" onchange="exec();">'
-  if (tbl.rows.length <= 2  && numPorts) {
-     console.log("CREATING TABLE ", tbl.rows.length);
-     for (let i = 2; i < 2 + numPorts; i++) {
-       const tr = tbl.insertRow();
-       let td = tr.insertCell(); td.appendChild(document.createTextNode(`Port ${i-1}`));
-       td = tr.insertCell();
-       td.innerHTML = limit.replaceAll("limit_port", "ilimit_port_" + i).replace("exec()", "iClicked(" + i + ")");
-       td = tr.insertCell();
-       td.innerHTML = 'UNLIMITED';
-       td = tr.insertCell();
-       td.innerHTML = limit.replaceAll("limit_port", "fc_port_" + i).replace("exec()", "document.getElementById('bwapply_" + i + "').disabled=false;");
-       td = tr.insertCell();
-       td.innerHTML = limit.replaceAll("limit_port", "elimit_port_" + i).replace("exec()", "eClicked(" + i + ")");
-       td = tr.insertCell();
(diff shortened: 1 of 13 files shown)


logicog:
Cool. This looks like a real implementation. The original code was more like a proof-of-concept...

mcaptur:
I'l test over the weekend.. haven't touched code since my uni days 30 years ago! These AI tools are getting very helpful..

logicog:
I wanted to mention a point of caution: The JavaScript and HTML end up in the flash of the switch and will be transferred from the switch to the Browser when the page loads. Unlike for C-code which is turned into machine-code bytes without any of the comments and without taking lengths of variables into account in the device's flash, this will include all comment-texts. Variables are stored/transferred with every single byte of their name. The flash memory of most devices is probably large enough not to be worried, but transferring the data will result in longer times for the browser to display the page, because the IP-stack of the devices is very slow. Roughly every frame of up to 2500 Bytes takes 20ms to transfer. Since the Browser sends 6-8 requests in parallel, there can be also timeouts of requests because uIP handles only one request at a time. In that case you see the page being built up by the browser in a couple of seconds, only.

We can probably improve that by using some kind of html/JS compression, but it will add a layer of complexity. On the other hand, we should also think about translated pages, and there, there will also be some kind of processing of the pages, I imagine.

feelfree69:
> We can probably improve that by using some kind of html/JS compression, but it will add a layer of complexity

I guess compressing / minifying html/.js code at compile time shouldn't be a problem at all, there are plenty of tools.

With other words: In any case, we should not omit any comments or better readable code just because we "fear" longer loading times.

logicog:
I did some testing, and so far everything works like a charm. I will need some more time for testing, though, because I see that the changes are across many files. I also only casually looked at the code.

feelfree69:
Don't know if I find some time to test (have no knowledge in html/js code, so reading the diff is useless in my case).

Side question, related to github: What is the easiest/fastest way to get the code of this PR? Just clone the mcaptur repo? For such small repo that's not a problem, but in case of huge repos that seems not very efficient....

mcaptur:
Im Building a new ui.. give me sometime as had to go out and need to fix some kinks.. its much more responsive now.. but will need some ppl to test

logicog:
> Side question, related to github: What is the easiest/fastest way to get the code of this PR? Just clone the mcaptur repo?

As usual, there is solution in git ;-)
```
$ git fetch upstream pull/219/head:pr219
$ git checkout pr219
```
Note that you need to configure the upstream remote repository. You basically get a branch for the PR in your local git repository.

logicog:
> Im Building a new ui.. give me sometime as had to go out and need to fix some kinks.. its much more responsive now.. but will need some ppl to test

Hi Marc, can you separate the different aspects of the code you are going to write into different PRs? This PR here is about fixing the configuration storage in flash. And maybe a different one for the UI? This makes testing easier and also prevents feature-creep, i.e. if you have an omnibus for all your work, people will also start to ask more and more from you. Alternatively, you will put more and more in, just because it fits. But if you have something specific this is much less of a problem.
You can move commits between PRs using git cherry-picking.

mcaptur:
built a new ui.. much faster than previous one.. i think everything mostly works now I know its a diff approach but give it a spin and it is much less taxing on the embeded cpu..  

https://github.com/logicog/RTLPlayground/pull/220

PR #220, opened 2026-05-09
Fastui

mcaptur:
New ui implementation.. much much faster and traps more errors - is dis ome testing and all seems to work.... everything is on one file now.. harder to maintain (migh consider splittig up a bit in future) but avoids avoids webserver overload - for now test out speed and operation..


logicog:
This is absolutely amazing! It is indeed blazingly fast, this feels like a high-end web-server behind. I would never have thought this to be possible.
I love that it is less nerdy than the previous UI, it is truely for Home-users with lots of information. I see plenty of stuff explained where already my wife told me that more explanation would be needed. ;-)
I also love the modern feeling like the notices displayed in the corner when things happen.

This will take a bit of time to test-drive, though, hopefully by several testers, as everything was moved around and a lot of things were refactored.
My first random comments:
- There is a "log3" file in the html directory, that does probably not belong there.
- The mouse-overs with a lot of information on the (SFP-) ports is gone. A lot of though went into defining what information and how to show it. Can this be re-enabled?
- Port-Speeds and EEE are not L2-features, they are L1 and this means unlike L2 these are per-port settings. Maybe have a menu item "Port Settings"? with the speed configuration and EEE?
- While it is much easier to understand than the previous UI, the Nerd-Settings are also important. What about having e.g. a general EEE-on button, but also an advanced menu where one can do this per-port?
- Have you given a thought to translations of the UI? That may be something really important to some people?

BTW: While writing this, I constantly moved between the switch UI and this page, and I was every time amazed discovering more new really nice things in this new UI!

feelfree69:
Just hat a first look at this. I'm speechless. This is a real game-changer.

mcaptur:
unfortunately the save config is corrupting and i messed up my switch and since i dont have a serial port or reset switch i need to unsolder chip to flash (it wont flash when on the board) so will continue when i find sometime.. its only a few hours of claude and gemin did not do much myself.. but speed is much better than before

AccountsTM:
Can confirm the UI is much better both for looks and especially for the speed and usability. There is an issue tho, its pretty much unusable on mobile. Maybe think about this in the next update! Thanks

logicog:
> unfortunately the save config is corrupting and i messed up my switch and since i dont have a serial port or reset switch i need to unsolder chip to flash (it wont flash when on the board) so will continue when i find sometime.. its only a few hours of claude and gemin did not do much myself.. but speed is much better than before

You do not need a switch to work on the UI. There is a simulator for the device, which simulates all of the commands the browser may send:

```
RTLPlayground/html$ ../tools/output/httpd_sim 
=== Waiting for connection on port 8080 === 
bytesRead: 468
GET / HTTP/1.1
Host: localhost:8080
User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:140.0) Gecko/20100101 Firefox/140.0
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8
Accept-Language: en-US,en;q=0.5
Accept-Encoding: gzip, deflate, br, zstd
Connection: keep-alive
Cookie: session=1234567890ab
Upgrade-Insecure-Requests: 1
Sec-Fetch-Dest: document
Sec-Fetch-Mode: navigate
Sec-Fetch-Site: none
Sec-Fetch-User: ?1
Priority: u=0, i


Session: >1234567890ab
Upgrade-Insecure-Requests: 1
Sec-Fetch-Dest: document
Sec-Fetch-Mode: navigate
Sec-Fetch-Site: none
Sec-Fetch-User: ?1
Priority: u=0, i

<. time now: 1778408922 last 0
Session expired
Time now: 1778408922 last 0
=== Waiting for connection on port 8080 === 
bytesRead: 478
GET /login.html HTTP/1.1
```

Just start it in the html directory and it makes a single-threaded web-server that takes and replies to the commands from the browser. Connect at `http://localhost:8080`
When developing the UI I mostly use that, because you can directly edit the html/JS code and it will be directly and immediately served. This is much faster than to have to flash the switch.

mcaptur:
Cool let  me try out this will help immensly!

mcaptur:
broke swith competely as burnt traces when soldering epprom.. need to get new one to continue playing.. which model do you suggest which ideall also has an easy ser
ial interface?

mcaptur:
> * The mouse-overs with a lot of information on the (SFP-) ports is gone. A lot of though went into defining what information and how to show it. Can this be re-enabled?
> * Port-Speeds and EEE are not L2-features, they are L1 and this means unlike L2 these are per-port settings. Maybe have a menu item "Port Settings"? with the speed configuration and EEE?
> * While it is much easier to understand than the previous UI, the Nerd-Settings are also important. What about having e.g. a general EEE-on button, but also an advanced menu where one can do this per-port?

theese shoudl be done

feelfree69:
> broke swith competely as burnt traces when soldering epprom.. need to get new one to continue playing.. which model do you suggest which ideall also has an easy ser ial interface?

Just made a PR for the Keeplink KP-9000-6XH-X2. 
I am not able to link the cheapest offer, but you should find it an Ali below 30€, 
https://de.aliexpress.com/item/1005008058381405.html

logicog:
The keeplink devices definitely have one of the best price/performance ratios. I believe they are all supported.

logicog:
More comments, sorry for the salami-tactics, I was just testing the latest of this PR and tried out the new stuff. The new L1 settings menu already looks great :
image
The device information loads forever, here.

- On the login page, it says "Secure Login". Unfortunately, we are not yet there and the login is merely via http, so in order not to make false promises, we should better call it "Login".
- The order of the ports is in their logical order in the switch, not the physical order on the switch. I.e. my 4+2 device shows port 5, 1, 2, 3, 4, 6 With the SFP ports 5 and 6 actually on the right side. The port information provided by the switch provides this additional information between logical and physical.

vDorst:
Really looks nice and fast.
Fastness is from that is just one page, which invisible elements that shown clicked on the menu.

This first bug that I hit is:
- Login
- Click on "Port Settings (L1)" 

This error shows in the browser java console.
```
Uncaught ReferenceError: configuration is not defined
    parseConf http://192.168.10.247/main.js:631
```

kekopop:
> unfortunately the save config is corrupting and i messed up my switch and since i dont have a serial port or reset switch i need to unsolder chip to flash (it wont flash when on the board) so will continue when i find sometime.. its only a few hours of claude and gemin did not do much myself.. but speed is much better than before

> broke swith competely as burnt traces when soldering epprom.. need to get new one to continue playing.. which model do you suggest which ideall also has an easy ser ial interface?

A couple of suggestions if you want to play around with your new switch once you get it:

Often programmers have some kind of current limiting/overcurrent protection, other circuitry on the board might trip it when trying to program a chip in circuit. Sometimes you can up the current limit in the programming software. Might be worth checking/trying. My switch needed 250mA to program.

Otherwise there are SOIC8 sockets like [this](https://www.adafruit.com/product/4726) available. Just desolder the chip and install the socket instead, then it's a 30 second operation to remove the chip for reprogramming if necessary. 

When it comes to the desoldering, [Chipquik alloy](http://www.chipquik.com/store/product_info.php?products_id=210001) or similar products simplifies desoldering of multi pin SMD components immensely if you don't have access to a proper hot air rework station, using a regular soldering iron/station instead.

Also, the UI really looks amazing :)

Erdnusschokolade:
Tested fastui on my 9XHML-X since your switch is down. The VLAN page has a few issues in `applyVLAN()` (main.js:500) — all ports end up untagged and PVID doesn't get set. Traced it to:

1. **Tagged ports get no `t` suffix** — `allMembers.push(p)` adds tagged and untagged ports identically, so the command is `vlan 10 1 2 3` with no `t` anywhere. Firmware treats suffix-less ports as untagged.
2. **`untag` isn't a firmware command** — it's sent as a separate POST but `cmd_parser.c` has no `untag`, so UART shows "Unknown command". (It's also in `getKey()`, so it gets saved to flash and fails on every boot.)
3. **PVID commands get dropped** — the pvid loop runs synchronously while the vlan request is still in-flight, so `fetchAPI`'s dedup (same url+method) discards them. `untag` survives only because it's sent later from the vlan callback.

The upstream `vlan_sub.js` handles all of this correctly — single `vlan` command with `${p}t` for tagged, pvid queued, everything sent sequentially with `await`. Might be the simplest reference for the fix.

I've also reproduced the config corruption on save and I'm taking a look at that now.

th0m4sek:
@mcaptur did You buy new switch and plan any update ?

vDorst:
Superseded by #429.

Issue #136, opened 2026-02-16
Keeplink KP-9000-6XH-X2 with 2M-PCB43-V2.1: Status?


feelfree69:
Just want to ask what's the integration status of this switch - I found photos in the repo:
https://github.com/logicog/RTLPlayground/tree/main/doc/devices/photos/2M-PCB43-V2.1-unmanaged, but no entry in machine.h.

Using machine Keeplink KP_9000_6XH**ML**_X2 works for RJ45 and SFP2 somehow, but all LEDs are scrambled.

If nobody else is working on this, I would try to find the correct config.

TylerDurden-23:
@feelfree69 
The KP_9000_6XHML_X2 will not work, since this was written for a non-N device, while the P-9000-6XH-X2 / 2M-PCB43-V2.1 is a N device. Seemingly most vendors are going to the N device - sometimes without changing the model number.

`#elif defined MACHINE_KP_9000_6XH_X2
__code const struct machine machine = {
	.machine_name = "keepLink KP-9000-6XH-X2",
	.isRTL8373 = 0,
	.min_port = 3,
	.max_port = 8,
	.n_sfp = 2,
	.log_to_phys_port = {0, 0, 0, 6, 1, 2, 3, 4, 5},
	.phys_to_log_port = {4, 5, 6, 7, 8, 3, 0, 0, 0},
	.is_sfp = {0, 0, 0, 2, 0, 0, 0, 0, 1},
	
	// Left SFP port
	.sfp_port[0].pin_detect = GPIO38, 
	.sfp_port[0].pin_los = GPIO_NA, 
	.sfp_port[0].sds = 1, 
	.sfp_port[0].i2c =  { .sda = GPIO39_I2C_SDA4, .scl = GPIO40_I2C_SCL3_MDC1 }, 

	// Right SFP port
	.sfp_port[1].pin_detect = GPIO37,
	.sfp_port[1].pin_los = GPIO_NA, 
	.sfp_port[1].sds = 0, 
	.sfp_port[1].i2c = { .sda = GPIO41_I2C_SDA3_MDIO1, .scl = GPIO40_I2C_SCL3_MDC1 }, 

	.reset_pin = GPIO_NA,
	.high_leds = { .mux = LED_27 | LED_29, .enable = LED_28 | LED_29 },
	.port_led_set = { 0, 0, 0, 2, 0, 0, 0, 0, 1},
	.led_sets = { 
			{
[...]

PR #221, opened 2026-05-10
Adding Keeplink KP-9000-6XH-X2

feelfree69:
As there was no progress on #136 since a couple of weeks, and on the other hand everything's working for me, I'll just go ahead and add this machine.

Maybe there's an issue with 1G / 2.5G SFP+ ports, but  I can't test them because I have none.


vDorst:
Thanks

Issue #222, opened 2026-05-13
Use of ! in password breaks the login.


jeditec:
The encoding of ! in the password lets say Pass!1234, I was not able to login I had to use curl to not encode it to be able to get in again. It saved me from another desoldering.

curl -X POST http://192.168.10.247/login \
        --data-raw "pwd=Pass!1234" \
        -v -L -c cookies.txt

Then i just modified the cookies in browser and worked. Then i went into the config and removed the ! so now I am sage XD.

logicog:
Thanks for the workaround. I will look into why this character is not correctly handled by the device / browser.

logicog:
Could you try the above fix?

jeditec:
Sorry for the delay but I didnt know how to checkout that fix, now I know how after a couple hours.

I did compile it again and it worked now I can use ! in my password.

But just out of curiosity I tried password !@#$%^&*()_+ and it failed, luckily I was still connected through serial.

logicog:
Thanks a lot for trying out the patch! Thanks also a lot for trying out the other password, you found another bug in the routine. It would be great if you could give it a try with that password again (I am traveling over the long ascension weekend and do not have myself a device to test with me...).

vDorst:
@jeditec other combination should be fixed now.
Otherwise let us know.
Please test it again.

jeditec:
I gave it a try but with some special characters it failed, I worked with my brother to see if we could fix it. Here is our solution.

I was able to use all English KB special characters: `~!@#$%^&*()-_=+[{]}\|;:'",<.>/?

As long as I kept below the 20 characters passwd limit I was able to use all of them.

To achieve this we did a patch in httpd/httpd.c

```
#define hex_val(ch) \
    ((ch) >= '0' && (ch) <= '9' ? (ch) - '0' : \
     (ch) >= 'A' && (ch) <= 'F' ? (ch) - 'A' + 10 : \
     (ch) >= 'a' && (ch) <= 'f' ? (ch) - 'a' + 10 : 0)
 
char is_url_word_x(__xdata uint8_t *c, __xdata uint8_t *d)
{
    __xdata uint8_t i = 0, j = 0, cv;
    while (d[i] && c[j]) {
        if (c[j] == '%') {
            j++;
            cv  = hex_val(c[j]) << 4;
            j++;
            cv += hex_val(c[j]);
        } else if (c[j] == '+') {
            cv = ' ';
        } else {
            cv = c[j];
        }
[...]

PR #224, opened 2026-05-14
Add URL decoding for password comparison.

logicog:
Text sent as a password POST request by a browser is URL-encoded. Decode the URL-encoded password on-the-fly when doing password comparison.
This fixes Issue #222

### DIFF
--- a/httpd/httpd.c
+++ b/httpd/httpd.c
@@ -131,6 +131,31 @@ char is_word(__xdata uint8_t *c, __code uint8_t * __xdata d)
 }
 
 
+char is_url_word_x(__xdata uint8_t *c,  __xdata uint8_t *d)
+{
+	uint8_t i = 0, j = 0;
+
+	while (d[i]) {
+		if (c[j] == '%') {
+			uint8_t v;
+			j++;
+			v  = c[j] - '0' < 10 ? (c[j] - '0') << 4 : (c[j] - 'A' + 10) << 4;
+			j++;
+			v += c[j] - '0' < 10 ? c[j] - '0' : c[j] - 'A' + 10;
+			if (d[i] != v)
+				return 0;
+		} else if (d[i] != c[j])
+			return 0;
+		i++;
+		j++;
+	}
+
+	if (c[j] != ' ' && c[j] != '\t' && c[j] != ':' && c[j] != '?' && c[j] != '=' && c[j] != '\n' && c[j] != '\r' && c[j])
+		return 0;
+	return 1;
+}
+
+
 char is_word_x(__xdata uint8_t *c, __xdata uint8_t *d)
 {
 	register uint8_t i = 0;
@@ -414,7 +439,7 @@ void handle_post(void)
 	} else if (is_word(request_path, "login")) {
 		dbg_string("POST login\n");
 		p += 8; // Read also over "pwd="
-		if (is_word_x(p, passwd)) {
+		if (is_url_word_x(p, passwd)) {
 			dbg_string("Password accepted!\n");
 			read_reg_timer(&last_session_use);
 			gen_random_bytes(session_id, SESSION_ID_LENGTH);


httpd/httpd.c
```diff
@@ -131,6 +131,33 @@ char is_word(__xdata uint8_t *c, __code uint8_t * __xdata d)
...
+		if (c[j] == '%') {
+			uint8_t v;
+			j++;
+			v  = c[j] - '0' < 10 ? (c[j] - '0') << 4 : (c[j] - 'A') << 4;
+			j++;
+			v += c[j] - '0' < 10 ? c[j] - '0' : c[j] - 'A';
+			if (d[i] != v)
+				break;
+		} else if (d[i] != c[j])
+			break;
```
vDorst: If it don´t match, don´t `break` but just return early.

So you don´t need anymore
```c
	if (d[i])
		return 0;
``

File: httpd/httpd.c [hunk 2]
```diff
@@ -131,6 +131,33 @@ char is_word(__xdata uint8_t *c, __code uint8_t * __xdata d)
...
+{
+	uint8_t i = 0, j = 0;
+
+	while (d[i]) {
+		if (c[j] == '%') {
+			uint8_t v;
+			j++;
+			v  = c[j] - '0' < 10 ? (c[j] - '0') << 4 : (c[j] - 'A') << 4;
+			j++;
+			v += c[j] - '0' < 10 ? c[j] - '0' : c[j] - 'A';
```
vDorst: 1. I think just put it in a for loop, or do-while().
2. Hex string is case-insensitive.
3. c - 'A', c = 'A' result in `0`, you need to add +10.
4. check is c is a valid hex, `(c | 0x20) - 'a' < 6` can be sufficient.

vDorst (changes requested):
I also wrote it in Rust to check many inputs.
See https://github.com/vDorst/clib/commit/74bcb702896224e927e72a184f86ec8ef7d6697f

httpd/httpd.c
logicog: Of course! I was so concentrated on not making mistakes while working on the basis of is_word_x()...

httpd/httpd.c
logicog: 1. I wanted to avoid another OSEG byte being used. The compiler does not do loop-unrolling and this function is as deep as it gets with nested functions. Currently, it is already the worst function in httpd.c:
```
; overlayable items in internal ram
        .area   OSEG    (OVR,DATA)
_find_entry_e_10000_106:
        .ds 2
        .area   OSEG    (OVR,DATA)
        .area   OSEG    (OVR,DATA)
        .area   OSEG    (OVR,DATA)
_is_url_word_x_PARM_2:
        .ds 2
_is_url_word_x_c_10000_116:
        .ds 2
_is_url_word_x_i_10000_117:
        .ds 1
        .area   OSEG    (OVR,DATA)
_is_word_x_PARM_2:
        .ds 2
```
it uses already 5 OSEG bytes, which is rather bad for such a little function.

2. Are you sure that lower-case characters are allowed for URL-encoding? Looking at RFC3986 section 2.1, it says that the producers of the URI should produce upper case letters. Should is really "must" in the context of an RFC.
3. Ooops... Fixed
4. I thought about that, too, but we are comparing keywords. If what the browser sends is corrupt, then the words do not match, so no login. There is also no possibility of a buffer overrun or otherwise an attack as we create in any case an unsigned 8 bit character. In my opinion it is safe to just assume the characters are hex.

vDorst: point 2, I quickly looked in https://en.wikipedia.org/wiki/Percent-encoding, and searched for `case`, This `Percent-encoding is not case-sensitive.` pops up so I made the assumption that is is case-insensitive. But `RFC3986` is it not, I say lets follow the `RFC3986`.

httpd/httpd.c
```diff
@@ -131,6 +131,31 @@ char is_word(__xdata uint8_t *c, __code uint8_t * __xdata d)
...
 
+char is_url_word_x(__xdata uint8_t *c,  __xdata uint8_t *d)
+{
+	uint8_t i = 0, j = 0;
+
+	while (d[i]) {
+		if (c[j] == '%') {
+			uint8_t v;
+			j++;
+			v  = c[j] - '0' < 10 ? (c[j] - '0') << 4 : (c[j] - 'A' + 0x10) << 4;
```
vDorst: It is just decimal `10` or `0x0a` in hex.

httpd/httpd.c
```diff
@@ -131,6 +131,31 @@ char is_word(__xdata uint8_t *c, __code uint8_t * __xdata d)
...
+			j++;
+			v += c[j] - '0' < 10 ? c[j] - '0' : c[j] - 'A' + 0x10;
```
vDorst: Also here

vDorst (changes requested):
`0x10` need to be just `10` or `0x0a`.

httpd/httpd.c
logicog: Oh, no! Will fix.

httpd/httpd.c

vDorst:
I found an idea to make a two cycle loop with a `bit`.

```c
char is_url_word_x(__xdata uint8_t *c,  __xdata uint8_t *d)
{
	uint8_t i = 0, j = 0, v;

	while (d[i]) {
		v = c[j];
		if (v == '%') {
			__bit again = true;
			v = 0;

			while(1) {
				j++;
				// Swap instruction is fine for rotation
				v = (v << 4) | (v >> 4);
				v |= c[j] - '0' < 10 ? (c[j] - '0') : (c[j] - 'A' + 10);

				// force `jbc`-instruction.
				if (again) {
					again = false;
				} else {
					break;
				}
			}
		}

		if (d[i] != v)
			return 0;

		i++;
		j++;
	}

	v = c[j];
	if (v && v != ' ' && v != '\t' && v != ':' && v != '?' && v != '=' && v != '\n' && v != '\r')
		return 0;
	return 1;
}
```

```patch
diff ../rtlplayground.mem  output/rtlplayground.mem
29c29
<    ROM/EPROM/FLASH  0x0000   0x2d032  91740    16777216
---
>    ROM/EPROM/FLASH  0x0000   0x2d038  91695    16777216
```

Still have to test it on a real machine.
But what to you think?

vDorst:
Set password to `test!&<` 
Send to device by the browser `pwd=test%21%26%3C`.
My code seems to work.

vDorst:
Thanks, merged to fix the bug and I make a PR with my code.

Issue #134, opened 2026-02-16
System seetings / Update seetings : Not recording correctly


dobodu:
Hi,

I noticed what I believed to be a bug / incorrect behaviour.

Update seeting : Does not work totally

- Updating IP adress : works as expected
- Updating Gateway adress : doesn't work correctly

Wheather I change only the gateway adress, of both (IP+Gateway), only IP adress is taken in consideration.
The gateway adress remains the old one.

I fixed that manually before building, adjusting the "config.txt" straight to the seetings I needed.

TylerDurden-23:
This may be related to: https://github.com/logicog/RTLPlayground/issues/110
If it is, it helps to use another browser - e.g. Firefox.

UAb5eSMn:
Setting ip, netmask and gw was fixed in pull request #226
Note that some other configuration settings may be lost when you save the settings. That has been reported in issue #218

PR #226, opened 2026-05-15
Support for executing multiple commands via /cmd

UAb5eSMn:
Allow multiple commands in a single http /cmd request.
Commands can be separated with \r and/or \n

This should fix issue #134
It could also be useful for atomic changes of vlan settings etc.

Example with curl:
```
#!/bin/bash

IP=192.168.10.247
PASSWORD=1234

IP_NEW=192.168.10.248
PASSWORD_NEW=admin

COMMANDS="
ip 192.168.10.248
gw 192.168.10.2
netmask 255.255.252.0
passwd admin
vlan 1
vlan 101 1 2 8t
vlan 102 3 4 8t
vlan 103 5 6 8t
vlan 104 7 8t 9
pvid 1 101
pvid 2 101
pvid 3 102
pvid 4 102
pvid 5 103
pvid 6 103
pvid 7 104
pvid 8 101
pvid 9 104
ingress 1u
ingress 2u
ingress 3u
ingress 4u
ingress 5u
ingress 6u
ingress 7u
ingress 8t
ingress 9u
vlan 101 mgmt
"

curl --cookie-jar cookie.txt --data "pwd=${PASSWORD}" http://${IP}/login
curl --max-time 5 --cookie cookie.txt --data "${COMMANDS}" http://${IP}/cmd

curl --cookie-jar cookie.txt --data "pwd=${PASSWORD_NEW}" http://${IP_NEW}/login
curl --cookie cookie.txt http://${IP_NEW}/information.json
```

Note that the implementation is far from perfect:
* Commands are processed sequentially and some limited checks are performed on command length and tokenization. If a check fails, the remaining commands are not executed. It would be better if all commands were validated first and then applied atomically if all were valid.
* Settings are applied immediately, so no http response will be received if the IP settings are changed
* A http 400 response does not report what was wrong.

### DIFF
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -1639,3 +1639,37 @@ config_done:
 	clear_command_history();
 	save_cmd = 1;
 }
+
+// Execute multiple commands
+// If a command is too long or can't be tokenized, remaining commands are not executed
+// Returns the status via `err_status`-variable.
+void execute_commands(__xdata uint8_t *p) __banked {
+	err_status = ERR_OK;
+	uint8_t cmd_idx = 0;
+	while (1) {
+		if (*p == 0 || *p == '\n' || *p == '\r') {
+			if (cmd_idx) {
+				cmd_buffer[cmd_idx] = '\0';
+				cmd_tokenize();
+				if (err_status != ERR_OK)
+					return;
+				cmd_parser();
+			}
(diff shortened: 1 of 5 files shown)


logicog:
Tested and works great! Thanks a lot for sharing this great idea!

PR #228, opened 2026-05-16
Add support for FOXNEO FNS-1200P (RTL8372, 4x2.5G PoE+ + 2x SFP+)

zectaiga:
Adds machine support for the FOXNEO FNS-1200P, an unmanaged RTL8372-based switch
with 4×2.5G PoE+ RJ45 ports and 2×SFP+ ports (1G/2.5G/10G).

## What was tested

Tested on hardware:
- All four 2.5G RJ45 ports at 10/100/1000/2500 Mbps
- Both SFP+ ports (module detection, link, I2C EEPROM)
- LEDs: amber (2.5G) and green (1G/100M/10M) per copper port; link/act on SFP ports
- Serial console via S1 through-holes (115200 8N1 3.3V TTL)

## How GPIO and LED values were determined

SFP GPIO assignments were determined by observing GPIO state changes during
module insertion/removal using RTLPlayground's \`gpio\` command, then
cross-checked against an original firmware register dump.

LED register values (led_mux, led_sets, high_leds, port_led_set) were decoded
from the original firmware register dump at 0x65e0–0x65f4 and 0x6540–0x654c,
and verified on hardware.

\`pin_tx_disable\` is GPIO_NA on both SFP ports — confirmed by the original
firmware keeping all GPIOs as inputs (GPIO_00_31_DIRECTION = 0x00000000).


vDorst:
Thanks

PR #229, opened 2026-05-17
Fix and refactor is_url_word_x()

### LINKED ISSUE #222: Use of ! in password breaks the login.

vDorst:
Should fix #222 again.

Fixed the `space` case.
Also check `content_type` is `application/x-www-form-urlencoded`.

### DIFF
--- a/httpd/httpd.c
+++ b/httpd/httpd.c
@@ -131,28 +131,46 @@ char is_word(__xdata uint8_t *c, __code uint8_t * __xdata d)
 }
 
 
-char is_url_word_x(__xdata uint8_t *c,  __xdata uint8_t *d)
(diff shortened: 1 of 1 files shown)


vDorst:
Also changes `__bit` to `bool`.

logicog:
Cool! Great stuff!

PR #230, opened 2026-05-17
Refactor is_word() and is_word_x()

vDorst:
Based on changes/idea of #229, also change `is_word()` and `is_word_x()`.

### DIFF
--- a/httpd/httpd.c
+++ b/httpd/httpd.c
@@ -116,18 +116,24 @@ char strcmp(__xdata uint8_t *c, __code uint8_t * __xdata d)
 }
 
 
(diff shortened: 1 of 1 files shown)


logicog:
From https://github.com/swegener/sdcc/blob/master/device/include/stdbool.h
I take it that SDCC actually supports a `bool`-type implemented via `__bit` on 8051. Would it not be better to use `bool` in the code after including `stdbool.h`? Including the header, would also allow to use `true` and `false` values which would improve readability for non-sdcc experts.

vDorst:
I'll changed it to a bool.

logicog:
Thanks for taking care of these important details!

PR #231, opened 2026-05-18
fix: rtl837x_stp bug, cmpMac always 0 when checking for new root

xristos-sk:
When checking for a new root, a root_bridge with the same priority as STP_I will never be adopted as cmpMAC always returns 0.

This PR fixes this bug by changing the comparison of MACs to what was intended


vDorst:
Good catch, thanks!

PR #232, opened 2026-05-21
Fix VLAN name persistence across rename and delete operations

Erdnusschokolade:
## Problem

VLAN names cannot be changed after they have been set once. Examples:

- `vlan 20 IoT p1u; vlan 20 IoT2` → still shows `IoT`
- `vlan 20 d; vlan 20 NewName p1u` → still shows `IoT`

## Root cause

`vlan_names[]` is an append-only buffer. `vlan_name()` returns the
first matching entry for a given VLAN ID. On update, new entries are
appended without removing the old one; on delete, the name entry is
left behind. The old name always wins.

## Fix

New helper function `vlan_name_remove(vlan)` in `rtl837x_port.c` that
locates the entry for a VLAN ID and removes it via array compaction.
It is called in two places:

- `parse_vlan()` in `cmd_parser.c`, before appending a new name entry
- `vlan_delete()` in `rtl837x_port.c`, when removing a VLAN

The implementation reuses the existing `vlan_name()` lookup, scans
forward to the trailing space of the matched entry, then shifts the
remaining bytes left with a manual `for` loop (SDCC 8051's `memmove()`
is not reliable for the overlap pattern needed here).

Locals are declared as `static __xdata` to avoid the OSEG overlay
segment limit on banked functions — six `uint16_t` autos in a banked
function exceed what SDCC's call-graph analysis can overlay.

## Tested on

KeepLiNK KP-9000-6XH-X (RTL8372 + RTL8221B):

- `vlan 99 AAA p1u; vlan 99 BBB` → name correctly updated to `BBB`
- `vlan 99 d; vlan 99 CCC p1u` → name correctly `CCC`, not stale `AAA`

## Code size impact

- Bank 2: +21 bytes
- Bank 0/1 unchanged

## Notes

Persistence across reboot is a separate issue that lives in the
existing `/config` download/upload workflow and is not part of this
PR.

## Disclosure

This PR was developed with AI assistance (Claude). Code was 
reviewed, hardware-tested on a KeepLiNK KP-9000-6XH-X, and I take 
full responsibility for the changes.

### DIFF
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -331,6 +331,8 @@ void parse_vlan(void)
 	vlan_settings.vlan = 0;
 	vlan_settings.members = 0;
 	vlan_settings.tagged = 0;
+	if (cmd_words_len < 2)
+		goto err;
 	if (!atoi_short(&vlan_settings.vlan, cmd_words_b[1])) {
 		if (cmd_words_len == 3 && cmd_buffer[cmd_words_b[2]] == 'd') {
 			vlan_delete(vlan_settings.vlan);
@@ -351,7 +353,7 @@ void parse_vlan(void)
 			vlan_names[vlan_ptr++] = hex[(vlan_settings.vlan >> 8) & 0xf];
 			vlan_names[vlan_ptr++] = hex[(vlan_settings.vlan >> 4) & 0xf] ;
 			vlan_names[vlan_ptr++] = hex[vlan_settings.vlan & 0xf];
-			while(cmd_buffer[cmd_words_b[w] + i] != ' ') {
+			while(cmd_buffer[cmd_words_b[w] + i] != ' ' && cmd_buffer[cmd_words_b[w] + i] != '\0') {
 				write_char(cmd_buffer[cmd_words_b[w] + i]);
 				vlan_names[vlan_ptr++] = cmd_buffer[cmd_words_b[w] + i++];
 			}
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -347,6 +347,7 @@ void parse_vlan(void)
 		uint8_t w = 2;
 		if (cmd_words_len > w && isletter(cmd_buffer[cmd_words_b[w]])) {
 			register uint8_t i = 0;
+			vlan_name_remove(vlan_settings.vlan);
 			vlan_names[vlan_ptr++] = hex[(vlan_settings.vlan >> 8) & 0xf];
 			vlan_names[vlan_ptr++] = hex[(vlan_settings.vlan >> 4) & 0xf] ;
 			vlan_names[vlan_ptr++] = hex[vlan_settings.vlan & 0xf];
--- a/rtl837x_port.c
+++ b/rtl837x_port.c
@@ -116,8 +116,42 @@ uint16_t port_pvid_get(uint8_t port) __banked
 void vlan_delete(uint16_t vlan) __banked
 {
 	print_string("\nvlan_delete called \n"); print_short(vlan);
+	vlan_name_remove(vlan);
 	REG_WRITE(RTL837x_TBL_DATA_IN_A, 0, 0, 0, 0);
 	REG_WRITE(RTL837X_TBL_CTRL, vlan >> 8, vlan, TBL_VLAN, TBL_WRITE | TBL_EXECUTE);
+}
+
+
+void vlan_name_remove(uint16_t vlan) __banked
+{
+	static __xdata uint16_t name_pos;
+	static __xdata uint16_t entry_start;
+	static __xdata uint16_t pos;
+	static __xdata uint16_t entry_len;
+	static __xdata uint16_t move_count;
+	static __xdata uint16_t j;
+
+	name_pos = vlan_name(vlan);
+	if (name_pos == 0xffff)
+		return;
+
+	entry_start = name_pos - 3;
+	pos = entry_start;
+
+	while (pos < vlan_ptr && vlan_names[pos] != ' ')
+		pos++;
+	if (pos >= vlan_ptr)
+		return;
+	pos++;
+
+	entry_len = pos - entry_start;
+	move_count = vlan_ptr - pos;
+
+	for (j = 0; j < move_count; j++)
+		vlan_names[entry_start + j] = vlan_names[pos + j];
+
+	vlan_ptr -= entry_len;
+	vlan_names[vlan_ptr] = 0;
 }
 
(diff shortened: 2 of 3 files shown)


logicog:
I am not sure this is related, maybe it was broken before, but I just tried, directly after reset:
```
> vlan 10 iot
iot255.255.0��iot255.255.0��iot255.255.0��iot255.255.0��iot255.255.0���
```
and then the device hangs.

I then tried:
```
> vlan 20 bla
bla255.255.0��vlan<

vlan_create called
vlan: 0x0014, members: 0x0200, tagged: 0x0200
vlan_create done 
vlan_ptr 0x0096:>014bla<
```
This also looks strange.

I also tried, again after reset:
```
> vlan 20 bla 2
bla<

vlan_create called
vlan: 0x0014, members: 0x0220, tagged: 0x0200
vlan_create done 
vlan_ptr 0x0007:>014bla <

> vlan 20 aaa
aaa25.255.0���
             vlan<

vlan_create called
vlan: 0x0014, members: 0x0200, tagged: 0x0200
vlan_create done 
vlan_ptr 0x0096:>014aaa<
> 
```

logicog:
> Locals are declared as `static __xdata` to avoid the OSEG overlay
> segment limit on banked functions — six `uint16_t` autos in a banked
> function exceed what SDCC's call-graph analysis can overlay.

That explains a lot of issues I had before. Where did you find that information?

feelfree69:
> I am not sure this is related, maybe it was broken before,

Same behaviour with main branch.

Erdnusschokolade:
Thanks for testing!

I reproduced the crash and traced it. The root cause is actually a
pre-existing bug in `parse_vlan()` at
[`cmd_parser.c:353`](https://github.com/logicog/RTLPlayground/blob/main/cmd_parser.c#L353),
not in `vlan_name_remove()` itself:

```c
while(cmd_buffer[cmd_words_b[w] + i] != ' ') {
    write_char(cmd_buffer[cmd_words_b[w] + i]);
    vlan_names[vlan_ptr++] = cmd_buffer[cmd_words_b[w] + i++];
}
```

The loop only terminates on `' '`, not on `'\0'`. When the VLAN name is
the last token on the line (no port specs after it), the scan runs past
the end of `cmd_buffer` and copies adjacent XRAM (uIP variables, syslog
buffer, etc.) into `vlan_names[]` until something corrupts enough state
for the device to hang. That's where the `255.255.0` bytes in your
output came from — those are `uip_netmask`.

The same pattern is handled correctly a few hundred lines down for port
names ([line 639](https://github.com/logicog/RTLPlayground/blob/main/cmd_parser.c#L639)),
which also checks for `'\0'` and a length bound.

My PR didn't introduce the bug, but it made it reliably reachable: before
the PR, `vlan N name` on its own would silently fail to update the name
(so people probably stopped trying), while my change actively rewrites
the buffer and exposes the trailing OOB scan every time. My own tests
always created VLANs with port specs in the same command
(`vlan N name PORTS`), where the trailing space terminates the loop
cleanly — so I never hit the case you tested. Lesson learned on
covering minimal-input edge cases.


---

On the OSEG question: honestly, I hit the OSEG linker error during
the build, and Claude (which I used as a coding assistant on this PR)
pointed me at `static __xdata` as the workaround. The fix worked, but
I don't have a deep understanding of SDCC's overlay analysis myself.

Erdnusschokolade:
Currently testing this fix:

```diff
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -351,7 +351,7 @@ void parse_vlan(void)
                       vlan_names[vlan_ptr++] = hex[(vlan_settings.vlan >> 8) & 0xf];
                       vlan_names[vlan_ptr++] = hex[(vlan_settings.vlan >> 4) & 0xf] ;
                       vlan_names[vlan_ptr++] = hex[vlan_settings.vlan & 0xf];
-                      while(cmd_buffer[cmd_words_b[w] + i] != ' ') {
+                      while(cmd_buffer[cmd_words_b[w] + i] != ' ' && cmd_buffer[cmd_words_b[w] + i] != '\0') {
                               write_char(cmd_buffer[cmd_words_b[w] + i]);
                               vlan_names[vlan_ptr++] = cmd_buffer[cmd_words_b[w] + i++];
                       }
```

logicog:
I tested with your fix, and everything now works as expected regarding the names. However, there was another bug in the parsing of the VLAN:
```
> vlan 20 eee
eee<

vlan_create called
vlan: 0x0014, members: 0x0200, tagged: 0x0200
vlan_create done 
vlan_ptr 0x0007:>014eee <

> vlan

vlan_create called
vlan: 0x0014, members: 0x0200, tagged: 0x0200
vlan_create done 

```
It looks like `vlan` without arguments parses random memory. So your additional fix and mine for that other bug give:
```
index 7f4004a..ba9ceb3 100644
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -331,6 +331,8 @@ void parse_vlan(void)
        vlan_settings.vlan = 0;
        vlan_settings.members = 0;
        vlan_settings.tagged = 0;
+       if (cmd_words_len < 2)
+               goto err;
        if (!atoi_short(&vlan_settings.vlan, cmd_words_b[1])) {
                if (cmd_words_len == 3 && cmd_buffer[cmd_words_b[2]] == 'd') {
                        vlan_delete(vlan_settings.vlan);
@@ -351,7 +353,7 @@ void parse_vlan(void)
                        vlan_names[vlan_ptr++] = hex[(vlan_settings.vlan >> 8) & 0xf];
                        vlan_names[vlan_ptr++] = hex[(vlan_settings.vlan >> 4) & 0xf] ;
                        vlan_names[vlan_ptr++] = hex[vlan_settings.vlan & 0xf];
-                       while(cmd_buffer[cmd_words_b[w] + i] != ' ') {
+                       while(cmd_buffer[cmd_words_b[w] + i] != ' ' && cmd_buffer[cmd_words_b[w] + i] != '\0') {
                                write_char(cmd_buffer[cmd_words_b[w] + i]);
                                vlan_names[vlan_ptr++] = cmd_buffer[cmd_words_b[w] + i++];
                        }
```
Could you add that fix to the PR? I tested with that fix and now things look great.

Erdnusschokolade:
Pushed as 7d0d852. Tested on 9XHML-X — `vlan` alone now exits with
the usage error, and the name parsing works in all the cases you
tried. Thanks for catching it!

logicog:
Great contribution! Thanks a lot.

PR #233, opened 2026-05-21
Add VLAN selector dropdown and overview table to vlan.html

Erdnusschokolade:
## Summary

Three related changes to the VLAN configuration page, in three atomic
commits:

1. **New `/vlanlist` HTTP endpoint** — returns a JSON array of all
   configured VLANs (ID + name)
2. **VLAN selector dropdown** above the ID input, populated from
   `/vlanlist`, with auto-refresh after Update/Create
3. **VLAN overview table** below the form, showing Member/Tagged/
   Untagged/PVID ports and a delete button per row

## Motivation

The existing UI requires the user to type a VLAN ID to inspect its
configuration. With multiple VLANs this becomes awkward, and there's
no way to get an overview of the full VLAN setup at a glance.

The dropdown solves the lookup problem; the table provides the
overview and convenient inline delete.

## Architecture: hybrid N+1 instead of backend aggregation

The table uses an N+1 request pattern: one call to `/vlanlist` to get
all IDs, then one `/vlan.json?vid=N` per VLAN. This keeps the backend
simple — the new `/vlanlist` endpoint is small and there's no risk of
TCP buffer overflow for setups with many VLANs.

For typical configurations (< 50 VLANs), page load completes in well
under one second.

## Port-to-bit mapping

The table uses the existing `physToLogPort[]` array (populated by
`/status.json` on page load), so port ranges in the table match the
icon view in `fetchVLAN()` exactly, regardless of machine.

## VLAN 1 protection

The delete button is not rendered for VLAN 1, since it's the 802.1Q
default VLAN.

## Code size impact

- Bank 2: +0 bytes (backend endpoint goes in Bank 1)
- Bank 1: +212 bytes (vlanlist endpoint + itoa16_html helper)
- HTML/JS: ~3.2 KB total
  - `vlan.html`: +394 bytes
  - `vlan.js`: +2808 bytes
  - `vlan_sub.js`: +4 bytes
  - `style.css`: +36 bytes

Total flash footprint increase: ~3.4 KB (~0.6% of 512K image).

## Tested on

KeepLiNK KP-9000-6XH-X (RTL8372 + RTL8221B). Tested scenarios:

- Dropdown populates correctly on page load
- Selecting a VLAN in the dropdown triggers the existing
  `fetchVLAN()` flow
- Table reflects all configured VLANs with correct port ranges
- Update/Create refreshes both dropdown and table without F5
- Delete via table button removes VLAN, refreshes views
- Manual ID entry remains functional (dropdown is additive)

## Screenshots

grafik
grafik

## Style note

The new JavaScript uses `fetch()` and `async/await` for readability of
the N+1 loop, while the existing code base uses `XMLHttpRequest` +
`sendXHTTP()`. Happy to convert if the maintainer prefers consistency
over readability here.

### DIFF
--- a/httpd/page_impl.c
+++ b/httpd/page_impl.c
@@ -97,6 +97,19 @@ void itoa_html(uint8_t v)
 	char_to_html('0' + (v % 10));
 }
 
+void itoa16_html(uint16_t v)
+{
+	uint8_t print_zeros = 0;
+	uint8_t d;
+	d = v / 1000;
+	if (d) { char_to_html('0' + d); print_zeros = 1; }
+	d = (v / 100) % 10;
+	if (d || print_zeros) { char_to_html('0' + d); print_zeros = 1; }
+	d = (v / 10) % 10;
+	if (d || print_zeros) char_to_html('0' + d);
+	char_to_html('0' + (v % 10));
+}
+
 void string_to_html(__code char *s)
 {
 	while (*s) char_to_html(*s++);
@@ -824,4 +837,45 @@ void send_cmd_log(void)
 			outbuf[slen++] = cmd_history[p];
 		p = (p + 1) & CMD_HISTORY_MASK;
 	}
+}
+
+
+void send_vlanlist(void)
+{
+	/* Worst case per entry: {"id":4094,"name":"<name>"}, ~45 bytes.
+	 * HTTP header ~50 bytes. TCP_OUTBUF_SIZE=2500 fits ~53 VLANs safely. */
+	__xdata uint16_t i;
+	__xdata uint16_t n;
+	uint8_t first = 1;
+
+	slen = strtox(outbuf, HTTP_RESPONCE_JSON);
+	char_to_html('[');
+
+	for (i = 1; i < 4095; i++) {
+		if (vlan_get(i) < 0)
+			continue;
+		if (!(sfr_data[0] & 0x02))
+			continue;
+
+		if (!first)
+			char_to_html(',');
+		first = 0;
+
+		slen += strtox(outbuf + slen, "{\"id\":");
+		itoa16_html(i);
+		slen += strtox(outbuf + slen, ",\"name\":\"");
+
+		n = vlan_name(i);
+		if (n != 0xffff) {
+			while (vlan_names[n] && vlan_names[n] != ' ')
+				char_to_html(vlan_names[n++]);
+		}
+
+		slen += strtox(outbuf + slen, "\"}");
+
+		if (slen > TCP_OUTBUF_SIZE - 60)
+			break;
+	}
+
+	char_to_html(']');
 }
--- a/httpd/page_impl.c
+++ b/httpd/page_impl.c
@@ -97,7 +97,7 @@ void itoa_html(uint8_t v)
 	char_to_html('0' + (v % 10));
 }
 
-void itoa16_html(uint16_t v)
+void itoa16_html(uint16_t v) /* sufficient for VLAN IDs (max 4094) */
 {
 	uint8_t print_zeros = 0;
 	uint8_t d;
@@ -842,8 +842,9 @@ void send_cmd_log(void)
 
 void send_vlanlist(void)
 {
-	/* Worst case per entry: {"id":4094,"name":"<name>"}, ~45 bytes.
-	 * HTTP header ~50 bytes. TCP_OUTBUF_SIZE=2500 fits ~53 VLANs safely. */
+	/* Worst case per entry: {"id":4094,"name":"<117-char name>"} = 138 bytes
+	 * (name bound: CMD_BUF_SIZE=128 minus command prefix); +1 for closing ']'.
+	 * At worst case ~18 VLANs fit; typical configs with short names fit many more. */
 	__xdata uint16_t i;
 	__xdata uint16_t n;
 	uint8_t first = 1;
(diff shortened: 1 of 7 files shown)


Erdnusschokolade:
Sorry forgot to add:

## Disclosure

This PR was developed with AI assistance (Claude). Code was 
reviewed, hardware-tested on a KeepLiNK KP-9000-6XH-X, and I take 
full responsibility for the changes.

logicog:
Great idea and great stuff! Will test asap!

Erdnusschokolade:
Self-noticed bug in the table: the "Untagged Ports" column was 
showing non-member ports as untagged. The hardware bitmask format 
encodes "untagged member" and "non-member" with the same bit value 
in the upper 10 bits (per doc/vlan.md), and `loadVlanTable()` was 
not masking by membership before display.

Fixed by ANDing with the member bitmask, matching the logic already 
used in `fetchVLAN()`. Pushed as commit c653e4e.

Verified on KeepLiNK KP-9000-6XH-X.

httpd/page_impl.c
```diff
@@ -825,3 +838,44 @@ void send_cmd_log(void)
...
+	__xdata uint16_t n;
+	uint8_t first = 1;
+
+	slen = strtox(outbuf, HTTP_RESPONCE_JSON);
+	char_to_html('[');
+
+	for (i = 1; i < 4095; i++) {
+		if (vlan_get(i) < 0)
+			continue;
+		if (!(sfr_data[0] & 0x02))
```
feelfree69: A comment would be helpful; maybe even better de-magic the 0x02

httpd/page_impl.c
```diff
@@ -825,3 +838,44 @@ void send_cmd_log(void)
...
+
+		n = vlan_name(i);
+		if (n != 0xffff) {
+			while (vlan_names[n] && vlan_names[n] != ' ')
+				char_to_html(vlan_names[n++]);
+		}
+
+		slen += strtox(outbuf + slen, "\"}");
+
+		if (slen > TCP_OUTBUF_SIZE - 60)
```
feelfree69: Why 60 bytes and not 50 or 70? In other words: are really 60 bytes "reserve" required or is it just a guess?

httpd/page_impl.c
```diff
@@ -97,6 +97,19 @@ void itoa_html(uint8_t v)
 	char_to_html('0' + (v % 10));
 }
 
+void itoa16_html(uint16_t v)
+{
```
feelfree69, line 101: I know it's used for VLAN Id (<=4094) only. 
But adding support for values >9999 would be straightforward and make the function more universal.

feelfree69 (commented):
Just read the .c code and added a few minor comments.

httpd/page_impl.c
Erdnusschokolade: Added an inline comment clarifying the bit's meaning. Kept it as a literal rather than a constant since the bit is only checked in this one place.

httpd/page_impl.c
Erdnusschokolade: Good question — digging into it, the post-write check actually didn't prevent overflows for long VLAN names. The 117-char name bound (from CMD_BUF_SIZE) means a single entry can reach 138 bytes, so I replaced the post-check with a pre-write guard using the worst-case size (138 + 1 for the closing ]). Comment updated with the math.

httpd/page_impl.c
Erdnusschokolade: Kept the function scoped to VLAN-ID range with a comment to make the intent explicit. Is there a use case you have in mind where larger values would be needed? Happy to generalize if so.

feelfree69:
Just tested with my SWTGW218AS -> works perfectly. Thanks a lot!

amstan:
Why is the new UI in german?

logicog:
> Why is the new UI in german?

It is not. It looks like just the screenshots were translated, maybe by a translation engine in the browser. When I look a the commits, everything, including the comments, is in English.

amstan:
I'm currently running v0.1.0-5a8a8ed, the latest from the master branch. If i go to /vlan.html, the first setting is labeled with ` VLAN auswählen: `, the dropdown has the `— VLAN wählen —` value in it. Exactly like the screenshot shows.

If i check the code, indeed, vlan.html contains german on line 14: https://github.com/logicog/RTLPlayground/blob/main/html/vlan.html#L14. The [git blame](https://github.com/logicog/RTLPlayground/blame/5a8a8ed91b4008338ff6db23d9579f895306948a/html/vlan.html#L14) sent me to this PR.

logicog:
> I'm currently running v0.1.0-5a8a8ed, the latest from the master branch. If i go to /vlan.html, the first setting is labeled with `VLAN auswählen:`, the dropdown has the `— VLAN wählen —` value in it. Exactly like the screenshot shows.
> 
> If i check the code, indeed, vlan.html contains german on line 14: https://github.com/logicog/RTLPlayground/blob/main/html/vlan.html#L14. The [git blame](https://github.com/logicog/RTLPlayground/blame/5a8a8ed91b4008338ff6db23d9579f895306948a/html/vlan.html#L14) sent me to this PR.

Sorry, I overlooked that: you are right, but I believe this is the only line, everything else appears to be in English. Maybe @Erdnusschokolade  could fix this in a follow-up PR?

PR #239, opened 2026-05-25
doc: Describe other PCB-K0402WS-V3.0 devices

amstan:
I recently brought a "Ztyuav Z-QWYT0402", the insides looked identical to the existing supported "Hisource Hi-K0402WS". Both of them have the "PCB-K0402-V3.0" silkscreen.

I flashed the vanilla image and it seems to work great.

I changed the documentation to be more inclusive of other devices (including adding a picture of mine). I suspect more devices might fall into this "PCB-K0402-V3.0" bucket, for example the "YuanLey YS25-0402" with the vlan switch also looks suspiciously identical from the outside.

I also added some details I noticed while flashing mine: teardown, flash padding, how the selector switch actually works.


machine.c
```diff
@@ -536,9 +536,9 @@ __code const struct machine machine = {
 
 void machine_custom_init(void) { }
 
-#elif defined MACHINE_HI_K0402WS
+#elif defined(MACHINE_PCB_K0402WS_V3) || defined(MACHINE_HI_K0402WS)  // Sold as a variety of devices, see doc/
```
feelfree69, line 539: As you deleted MACHINE_HI_K0402WS in machin.h, it should also be deleted here.

feelfree69 (commented):
Thanks a lot for your contribution!
I agree that the PCB print is the better option for selecting the correct MACHINE in machine.h.

machine.c
amstan: I was thinking to keep the backwards compatibility to not break anyone.

feelfree69: Good point. 
Let's wait for @logicog and see what he thinks.

logicog: I think keeping the backward compatibility would be good. The define can also be set through an environment variable. So better to keep it.

PR #241, opened 2026-05-26
Correct argument count for regset command

tofurky:
Fixes: 2cdab5f9d9038266 ("parse_reg{set,get}() make use cmd_words_len.")


logicog:
Thanks for spotting and sending the fix @tofurky !

PR #242, opened 2026-05-26
Broaden range of SFP+ rates mapped to SDS_10GR

tofurky:
A commonly used SFP+ form-factor XGS-PON ONT (WAS-110) reports 0x63 at power on and then switches to 0x64 after booting up.

0x63 -> 9.9gbps
0x64 -> 10gbps

These are outside of the previous 0x66-0x69 range which prevented the link from ever being brought up.


logicog:
It makes a lot of sense to be as inclusive here, as possible. We have had to widen the bit ranges already in the past several times, especially because modules reported real bit-rates without the error correction overhead.
Thanks so much for contributing this fix!

Issue #245, opened 2026-05-28
SFP port speed change


Pretous:
Hello,

Could we get an option to manually lock the SFP+ port speed in both the GUI and CLI?

Right now, there's no way to hardcode the link speed, which is causing me some real headaches. The stock firmware allowed switching between 1G, 2.5G, and 10G, but here it seems to rely entirely on auto-negotiation.

Because of this, I can't get my 2.5G SFP module to work properly—the link just won't negotiate correctly without forcing the speed.

It would be awesome if we could get:

  -  A simple dropdown in the GUI interface settings (Auto / 1G / 2.5G / 10G)

  -  A quick CLI command to do the same.

logicog:
I'll look into forcing speeds for the SFP+ ports, but I will need some help with testing as I probably do not have the right SFP modules for that. For example, I do not have any 10GBit modules that could be forced to do 2.5GBit, my 10GBit modules only allow 10GBit or 1GBit.

Pretous:
Will do my best but i have just 1/2,5g module or 10g DAC's - yet got 2 switches with RTLPLayground so maybe via dac i can force 2,5g or 5G connection between them

logicog:
Could you try my PR #246 ?

I am able with this to change the speed of a 10GBit module into 1GBit, the same as on the other side and establish a link:
```
<MODULE INSERTED>  Slot: 2  Rate: 67  Encoding: 06  Module: OEM             SFP-10G-SR      2   

sds_config sds: 01, mode: 1a

RTL837X_REG_SDS_MODES: 0x0000035a

<SFP-RX OK>  Slot: 2

<new link: 2000054000, was 2000050000>

<SFP-RX OK>  Slot: 1


> sfp

Slot 1 - Rate: 67  Encoding: 01
Lightron Inc.   WSPXG-ES3LC-IHA 0000
Options: 68
Temp: 2a38
Vcc: 7ede
TX Bias: 3afc
TX Power: 185e
RX Power: 22ef
Laser: 0000
State: 30

Slot 2 - Rate: 67  Encoding: 06
OEM             SFP-10G-SR      2   
Options: 68
Temp: 219b
Vcc: 80a2
TX Bias: 0d24
TX Power: 1583
RX Power: 0dad
Laser: 0000
State: 30

> sfp 2 1g
 1G

<MODULE INSERTED>  Slot: 2  Rate: 0c  Encoding: 06  Module: OEM             SFP-10G-SR      2   

[...]

PR #246, opened 2026-05-28
Add support for configuring SFP modules speeds

logicog:
Add configuration commands for SFP speed configuration, e.g. `sfp 1 1g`, which allow modules which have higher speeds to be configured to use lower speeds, such as 1gbit for a 10gbit module. This has to be supported by the module hardware.
This is supposed to address issue #245 

This is now ready to merge, the 5GBit support is however missing, due to lack of test hardware.

### DIFF
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -791,6 +791,9 @@ void parse_sfp(void)
 	} else if (cmd_compare(2, "1g")) {
 		print_string(" 1G\n");
 		sfp_speed[slot] = SFP_SPEED_1G;
+	} else if (cmd_compare(2, "100m")) {
+		print_string(" 100M\n");
+		sfp_speed[slot] = SFP_SPEED_100M;
 	} else if (cmd_compare(2, "auto")) {
 		print_string(" AUTO\n");
 		sfp_speed[slot] = SFP_SPEED_AUTO;
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -48,6 +48,8 @@ __xdata uint16_t vlan_ptr;
 __xdata char port_names[9][PORT_NAME_SIZE];
 
 extern __xdata uint16_t management_vlan;
+extern __xdata uint8_t sfp_speed[2];
+extern __xdata uint8_t sfp_pins_last;
 __xdata uint8_t gpio_last_value[8] = { 0 };
 
 // Temporatly for str to hex convertion value.
@@ -748,6 +750,61 @@ void sfp_print_measurements(uint8_t sfp)
 }
 
 
+void parse_sfp(void)
+{
+	uint8_t slot;
+
+	if (cmd_words_len != 1 && cmd_words_len != 3)
+		goto err;
+
+	if (cmd_words_len == 1) {
+		for (slot = 0; slot < machine.n_sfp; slot++) {
+			print_string("\nSlot "); write_char('1' + slot);
+			if (gpio_pin_test(machine.sfp_port[slot].pin_detect)) {
+				print_string(" - empty\n");
+				continue;
+			}
+			print_string(" - Rate: "); print_byte(sfp_read_reg(slot, 12));
+			print_string("  Encoding: "); print_byte(sfp_read_reg(slot, 11));
+			write_char('\n');
+			sfp_print_info(slot);
+			sfp_print_measurements(slot);
+		}
+		return;
+	}
+	if (cmd_buffer[cmd_words_b[1]] < '1' || cmd_buffer[cmd_words_b[1]] > '2' || cmd_buffer[cmd_words_b[1] + 1] != ' ' ) {
+		print_string("Illegal SFP slot number\n");
+		return;
+	}
+	slot = cmd_buffer[cmd_words_b[1]] - '1';
+	if (slot >= machine.n_sfp) {
+		print_string("SFP slot not present\n");
+		return;
+	}
+
+	if (cmd_compare(2, "10g")) {
+		print_string(" 10G\n");
+		sfp_speed[slot] = SFP_SPEED_10G;
+	} else if (cmd_compare(2, "2g5")) {
+		print_string(" 2.5G\n");
+		sfp_speed[slot] = SFP_SPEED_2G5;
+	} else if (cmd_compare(2, "1g")) {
+		print_string(" 1G\n");
+		sfp_speed[slot] = SFP_SPEED_1G;
+	} else if (cmd_compare(2, "auto")) {
+		print_string(" AUTO\n");
+		sfp_speed[slot] = SFP_SPEED_AUTO;
+	} else {
+		goto err;
+	}
+	sfp_pins_last |= 0x1 << (slot << 2);
+	handle_sfp();
+	return;
+err:
+	print_string("\nUsage:\n\tsfp\n\tsfp [1|2] [1g|2g5|10g]\n");
+}
+
+
 void parse_regget(void)
 {
 	uint16_t reg = 0;
@@ -1355,18 +1412,7 @@ void cmd_parser(void) __banked
 			print_string("\nRESET\n\n");
 			reset_chip();
 		} else if (cmd_compare(0, "sfp")) {
-			print_string("\nSlot 1 - Rate: "); print_byte(sfp_read_reg(0, 12));
-			print_string("  Encoding: "); print_byte(sfp_read_reg(0, 11));
-			print_string("\n");
-			sfp_print_info(0);
-			sfp_print_measurements(0);
-			if (machine.n_sfp == 2) {
-				print_string("\nSlot 2 - Rate: "); print_byte(sfp_read_reg(1, 12));
-				print_string("  Encoding: "); print_byte(sfp_read_reg(1, 11));
-				print_string("\n");
-				sfp_print_info(1);
-				sfp_print_measurements(1);
-			}
+			parse_sfp();
 		} else if (cmd_compare(0, "stat")) {
 			port_stats_print();
 		} else if (cmd_compare(0, "flash") && cmd_words_len == 2) {
--- a/rtlplayground.c
+++ b/rtlplayground.c
@@ -137,6 +137,7 @@ __xdata char sfp_module_vendor[2][17];
 __xdata char sfp_module_model[2][17];
 __xdata char sfp_module_serial[2][17];
 __xdata uint8_t sfp_options[2];
+__xdata uint8_t sfp_speed[2];
 __xdata bool button_last;
 __xdata uint8_t button_sec_counter_last;
 volatile __bit tx_buf_empty;
@@ -1234,6 +1235,12 @@ void handle_sfp(void)
 				// Read Reg 12: Signalling rate (including overhead) in 100Mbit: 0xd: 1Gbit, 0x67:10Gbit
 				delay(100); // Delay, because some modules need time to wake up
 				uint8_t rate = sfp_read_reg(sfp, 12);
+				if (sfp_speed[sfp] == SFP_SPEED_1G)
+					rate = 0xc;
+				else if (sfp_speed[sfp] == SFP_SPEED_2G5)
+					rate = 0x19;
+				else if (sfp_speed[sfp] == SFP_SPEED_10G)
+					rate = 0x69;
 				print_string("  Rate: "); print_byte(rate);  // Normally 1, but 0 for DAC, can be ignored?
 				print_string("  Encoding: "); print_byte(sfp_read_reg(sfp, 11));
 				print_string("  Module: "); sfp_print_info(sfp);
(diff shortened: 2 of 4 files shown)


cmd_parser.c
```diff
@@ -751,6 +750,60 @@ void sfp_print_measurements(uint8_t sfp)
...
+{
+	uint8_t slot;
+
+	if (cmd_words_len != 1 && cmd_words_len != 3)
+		goto err;
+
+	if (cmd_words_len == 1) {
+		print_string("\nSlot 1 - Rate: "); print_byte(sfp_read_reg(0, 12));
+		print_string("  Encoding: "); print_byte(sfp_read_reg(0, 11));
+		print_string("\n");
```
vDorst: replace with `write_char('\n')`

cmd_parser.c
```diff
@@ -751,6 +750,60 @@ void sfp_print_measurements(uint8_t sfp)
...
+		sfp_print_info(0);
+		sfp_print_measurements(0);
+		if (machine.n_sfp == 2) {
+			print_string("\nSlot 2 - Rate: "); print_byte(sfp_read_reg(1, 12));
+			print_string("  Encoding: "); print_byte(sfp_read_reg(1, 11));
+			print_string("\n");
```
vDorst: replace with `write_char('\n')`

cmd_parser.c
```diff
@@ -1357,19 +1410,10 @@ void cmd_parser(void) __banked
 		if (cmd_compare(0, "reset")) {
 			print_string("\nRESET\n\n");
 			reset_chip();
+		} else if (cmd_compare(0, "sfp") && cmd_words_len == 3) {
+			parse_sfp();
```
vDorst: This can be removed, check below is the same.
This is redundant.

vDorst (commented):
Not tested yet, but small comments.

cmd_parser.c
logicog: Will fix.

cmd_parser.c
logicog: Will fix.

cmd_parser.c
logicog: Will fix.

logicog:
I am very interested in the results of testing with different SFP+ modules. I had to use a fixed 1GBit module on the other side of the link, because I use an intel X520 card in a PC for testing and the linux driver seems not able to change the speed to 1GBit for a 10GBit module. I unfortunately do not have any 2.5GBit capable SFP+ modules at hand, so could not test this at all.

logicog:
I pushed the fixes for the bugs found above. There is a pre-existing issue with the `sfp` command: if a slot is empty, it will still print garbage, as there is no check for whether there is an SFP module inserted. Probably, there should be some kind of `sfp_present()` function, which checks the present pin, but I have not thought this through, yet.

vDorst:
We can make use of `machine.pin_detect`. But also I2C returns an error bit when a `NACK` is received. We can use that to report that the module is not present.

cmd_parser.c
```diff
@@ -751,6 +750,60 @@ void sfp_print_measurements(uint8_t sfp)
...
+		print_string(" AUTO\n");
+		sfp_speed[slot] = SFP_SPEED_AUTO;
+	} else {
+		goto err;
+	}
+	sfp_pins_last |= 0x1 << (slot << 2);
+	handle_sfp();
+	return;
+err:
+	print_string("\nUsage:\n\tsfp\n\tsfp [0|1] [1g|2g5|10g]\n");
```
feelfree69: Usage string says 0 or 1, code checks for 1 and 2 as port numbers.

feelfree69:
With a DAC cable, the link is successfully established when setting sfp speed to 1g, 2g5 or 10g on both sides.
Setting to 5g seems not to work at all, it stays at 10g.

rtlplayground.c
```diff
@@ -1234,6 +1235,12 @@ void handle_sfp(void)
 				// Read Reg 12: Signalling rate (including overhead) in 100Mbit: 0xd: 1Gbit, 0x67:10Gbit
 				delay(100); // Delay, because some modules need time to wake up
 				uint8_t rate = sfp_read_reg(sfp, 12);
+				if (sfp_speed[sfp] == SFP_SPEED_1G)
+					rate = 0xc;
+				else if (sfp_speed[sfp] == SFP_SPEED_2G5)
+					rate = 0x19;
```
feelfree69, line 1249: 5g speed is missing here

cmd_parser.c
```diff
@@ -751,6 +750,60 @@ void sfp_print_measurements(uint8_t sfp)
...
 
+void parse_sfp(void)
+{
+	uint8_t slot;
+
+	if (cmd_words_len != 1 && cmd_words_len != 3)
+		goto err;
+
+	if (cmd_words_len == 1) {
+		print_string("\nSlot 1 - Rate: "); print_byte(sfp_read_reg(0, 12));
```
feelfree69: Consider `for (i=0; i<n_sfp-1; i++)` instead of doubling code?
Also consider n_sfp = 0 ?

rtlplayground.c
logicog: Yes, because I do not know how to configure the SerDes accordingly. The 10GBit Ethernet RTL8261 PHYs always run at 10GBit, even if the link speed is 5GBit. The info for 2.5 GBit comes from the RTL8221B which run at 2.5GBit, and it is really just a guess until someone tests it. 5GBit-capable fiber modules are extremely rare.

cmd_parser.c
logicog: Will fix usage string.

logicog: Will fix usage to stay in line with how we count the Ethernet ports.

logicog:
> We can make use of `machine.pin_detect`. But also I2C returns an error bit when a `NACK` is received. We can use that to report that the module is not present.

I'll try `machine.pin_detect`.

cmd_parser.c
logicog: Will do, that makes especially sense when we detect the module, too.

logicog:
Fixed the above (apart from the 5GBit setting). I added a check for the speed setting, which verifies that the slot is actually present, e.g. if the machine has only 1 slot.
Regarding the 5GBit, I found the code in the SDK, but as usual this is rather arcane. So far I am not able to see any differences to the 2.5GBit mode, but that does not make sense, unless there is not actually any support for 5GBit and it is merely handled like 2.5Gbit.

cmd_parser.c
```diff
@@ -751,6 +750,65 @@ void sfp_print_measurements(uint8_t sfp)
...
+{
+	uint8_t slot;
+
+	if (cmd_words_len != 1 && cmd_words_len != 3)
+		goto err;
+
+	if (cmd_words_len == 1) {
+		for (slot = 0; slot < machine.n_sfp; slot++) {
+			if (gpio_pin_test(machine.sfp_port[slot].pin_detect)) {
+				print_string("\nSlot "); write_char('1' + slot);
```
feelfree69: This line could be moved above the if statement to avoid double code

rtlplayground.c
feelfree69: If 5g is anyhow not supported yet, I would remove the option also from the sfp setting.

feelfree69:
Looks good to me!

rtlplayground.c
logicog: Will fix.

cmd_parser.c
logicog: Good catch! Will fix.

vDorst:
Do we also want to support `100m`? While testing with other switch which has still the original firmware, I see that is support 100mbit.
For the rest it is good to me.

logicog:
I had forgotten that 100MBit fiber modules actually existed. But I found support in the RTL-SDK, there is support for `SERDES_100FX`. For this to work I will need a look how difficult it is to support.

BTW: In the OEM Web-interface, do they support setting 5GBit for fiber modules? I am not sure the API version we have supports that.

feelfree69:
Personally, I would prefer merging PRs that are ready (as this one is) over adding just one more feature (that probably nobody will ever use)....

logicog:
> Personally, I would prefer merging PRs that are ready (as this one is) over adding just one more feature (that probably nobody will ever use)....

You are probably right. This can be merged, and support for 100FX is then a different story. We do not even have support for those modules, yet. Not even speaking of forcing faster modules into that mode.

logicog:
The following appears to work to allow 100MBit modules:
```
$ git diff
index 45308ba..b25a996 100644
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -791,6 +791,9 @@ void parse_sfp(void)
        } else if (cmd_compare(2, "1g")) {
                print_string(" 1G\n");
                sfp_speed[slot] = SFP_SPEED_1G;
+       } else if (cmd_compare(2, "100m")) {
+               print_string(" 100M\n");
+               sfp_speed[slot] = SFP_SPEED_100M;
        } else if (cmd_compare(2, "auto")) {
                print_string(" AUTO\n");
                sfp_speed[slot] = SFP_SPEED_AUTO;
index 3be8c7f..6d6abb5 100644
--- a/rtl837x_common.h
+++ b/rtl837x_common.h
@@ -94,6 +94,7 @@ struct vlan_tag {
 
 enum sfp_speeds {
        SFP_SPEED_AUTO = 0,
+       SFP_SPEED_100M,
        SFP_SPEED_1G,
        SFP_SPEED_2G5,
        SFP_SPEED_5G,
diff --git a/rtl837x_regs.h b/rtl837x_regs.h
index c1a27bb..f5bd951 100644
--- a/rtl837x_regs.h
+++ b/rtl837x_regs.h
@@ -73,6 +73,7 @@
  */
 #define SDS_SGMII              0x02
 #define SDS_1000BX_FIBER       0x04
+#define SDS_100FX              0x05
 #define SDS_QXGMII             0x0d
 #define SDS_HISGMII            0x12
 #define SDS_HSG                        0x16
index 7499f45..6b56db3 100644
--- a/rtlplayground.c
+++ b/rtlplayground.c
@@ -927,6 +927,10 @@ void sds_config(uint8_t sds, uint8_t mode)
                v = 0x0200;
                page = 0x2e;
                break;
+       case SDS_100FX:
+               v = 0x0200;
+               page = 0x26;
+               break;
        default:
                print_string("Error in SDS Mode\n");
                return;
@@ -1170,6 +1174,8 @@ void handle_tx(void)
 
 static inline uint8_t sfp_rate_to_sds_config(register uint8_t rate)
 {
+       if (rate == 0x1 || rate == 0x2)
+               return SDS_100FX;
        if (rate == 0xc || rate == 0xd)
                return SDS_1000BX_FIBER;
        if (rate >= 0x19 && rate <= 0x20)  // Ethernet 2.5 GBit
@@ -1235,7 +1241,9 @@ void handle_sfp(void)
                                // Read Reg 12: Signalling rate (including overhead) in 100Mbit: 0xd: 1Gbit, 0x67:10Gbit
                                delay(100); // Delay, because some modules need time to wake up
                                uint8_t rate = sfp_read_reg(sfp, 12);
-                               if (sfp_speed[sfp] == SFP_SPEED_1G)
+                               if (sfp_speed[sfp] == SFP_SPEED_100M)
+                                       rate = 0x1;
+                               else if (sfp_speed[sfp] == SFP_SPEED_1G)
                                        rate = 0xc;
                                else if (sfp_speed[sfp] == SFP_SPEED_2G5)
                                        rate = 0x19;

```

It seems I can make a 10GBit and a 1GBit module talk 100MBit to each other:
```
 sfp

Slot 1 - Rate: 0d  Encoding: 01
Delock          GLC-SX-MM       1.0 
Options: 08

Slot 2 - Rate: 67  Encoding: 06
OEM             SFP-10G-SR      2   
Options: 68
Temp: 2352
Vcc: 80a2
TX Bias: 0d2a
TX Power: 1512
RX Power: 0e9f
Laser: 0000
State: 30

> sfp 2 1g
 1G

<MODULE INSERTED>  Slot: 2  Rate: 0c  Encoding: 06  Module: OEM             SFP-10G-SR      2   

sds_config sds: 01, mode: 04

RTL837X_REG_SDS_MODES: 0x00000084

> 
<new link: 2000052000, was 2000050000>

<new link: 2000050000, was 2000052000>

<new link: 2200052000, was 2000050000>


> sfp 1 100m
 100M

<MODULE INSERTED>  Slot: 1  Rate: 01  Encoding: 01  Module: Delock          GLC-SX-MM       1.0 

sds_config sds: 00, mode: 05

RTL837X_REG_SDS_MODES: 0x00000085

> 
<new link: 2000052000, was 2200052000>

<new link: 2000050000, was 2000052000>


> sfp 2 100m
 100M

<MODULE INSERTED>  Slot: 2  Rate: 01  Encoding: 06  Module: OEM             SFP-10G-SR      2   

sds_config sds: 01, mode: 05

RTL837X_REG_SDS_MODES: 0x000000a5

> 
<new link: 2100051000, was 2000050000>


> stat

Port    State   Link    TxGood          TxBad           RxGood          RxBad
5       SFP IN  100M    0x004cfe38      0x00000000      0x004cfe37      0x00000000
1       On      2.5G    0x0042c9bf      0x00000000      0x000050dc      0x00000000
6       SFP IN  100M    0x004cfe39      0x00000000      0x004cfe37      0x00000001

```

vDorst:
Thanks

Issue #191, opened 2026-04-05
ZX-SG4T2


JanKiel2000:
I compiled the software for this switch. The diodes work, they show ok. The problem is, for example, that it is not shown whether the interface is up. Software updates via the GUI also do not work, but this switch is unmanageable by default and I am not sure if it should work. Below I publish photos and a snippet of code from machine.h and machine.c

![Image](https://github.com/user-attachments/assets/9c13c7b0-6cb7-481a-87e3-6ac49fa1cb1c)
![Image](https://github.com/user-attachments/assets/686f52df-7836-4986-b0ed-472a3cb5e38e)
![Image](https://github.com/user-attachments/assets/70c6174a-3a98-411e-b986-afc1245c10ed)
![Image](https://github.com/user-attachments/assets/6709ef5c-6129-4fb1-a7a7-4d2486a6c1a7)
![Image](https://github.com/user-attachments/assets/9c9c726d-857c-4d9e-89a6-aede47ed711f)

mahine.h
`
#define MACHINE_ZX_SG4T2
`

mahine.c
`
#elif defined MACHINE_ZX_SG4T2
__code const struct machine machine = {
        .machine_name = "ZX-SG4T2",
        .isRTL8373 = 0,
        .min_port = 3,
        .max_port = 8,
        .n_sfp = 2,
        .log_to_phys_port = {0, 0, 0, 6, 1, 2, 3, 4, 5},
        .phys_to_log_port = {4, 5, 6, 7, 8, 3, 0, 0, 0},
        .is_sfp = {0, 0, 0, 2, 0, 0, 0, 0, 1},

        // Left SFP port
        .sfp_port[1].pin_detect = GPIO38,
        .sfp_port[1].pin_los = GPIO_NA,
        .sfp_port[1].sds = 1,
        .sfp_port[1].i2c =  { .sda = GPIO39_I2C_SDA4, .scl = GPIO40_I2C_SCL3_MDC1 },

        // Right SFP port
        .sfp_port[0].pin_detect = GPIO37,
        .sfp_port[0].pin_los = GPIO_NA,
        .sfp_port[0].sds = 0,
        .sfp_port[0].i2c = { .sda = GPIO41_I2C_SDA3_MDIO1, .scl = GPIO40_I2C_SCL3_MDC1 },

        .reset_pin = GPIO_NA,
        .high_leds = { .mux =  LED_28 | LED_29, .enable = LED_27 | LED_28 | LED_29 },
[...]

PR #248, opened 2026-05-29
Fix SWTG024AS-A 2.0.1 port and LED mapping

zectaiga:
Related to #191.

## Changes

- Correct the SFP logical/physical port mapping.
- Map SDS0 to logical port 3 / physical port 5.
- Map SDS1 to logical port 8 / physical port 6.
- Update LED set definitions to match the stock firmware behavior.
- Enable `LED_GLB_IO_EN` bit 6 for the physical port 5 green LED.
- Adjust `LED_MODE` bits to match the stock firmware register value.

I compared the behavior against stock firmware register dumps. The stock firmware uses `LED_MODE = 0x0023e430`, and the updated initialization now matches that value from the default `0x0021e6b0` setup.

### DIFF
--- a/machine.c
+++ b/machine.c
@@ -786,37 +786,39 @@ __code const struct machine machine = {
     .min_port = 3,
     .max_port = 8,
     .n_sfp = 2,
-    .log_to_phys_port = {0, 0, 0, 6, 1, 2, 3, 4, 5},
-    .phys_to_log_port = {4, 5, 6, 7, 8, 3, 0, 0, 0},
-    .is_sfp = {0, 0, 0, 2, 0, 0, 0, 0, 1},
+    .log_to_phys_port = {0, 0, 0, 5, 1, 2, 3, 4, 6},
+    .phys_to_log_port = {4, 5, 6, 7, 3, 8, 0, 0, 0},
+    .is_sfp = {0, 0, 0, 1, 0, 0, 0, 0, 2},
 
-    // Left SFP port
-    .sfp_port[1].pin_detect = GPIO38,
-    .sfp_port[1].pin_los = GPIO_NA,
-    .sfp_port[1].sds = 1,
-    .sfp_port[1].i2c =  { .sda = GPIO39_I2C_SDA4, .scl = GPIO40_I2C_SCL3_MDC1 },
-
-    // Right SFP port
+    // SFP port on SDS0 / logical port 3
     .sfp_port[0].pin_detect = GPIO37,
     .sfp_port[0].pin_los = GPIO_NA,
+    .sfp_port[0].pin_tx_disable = GPIO_NA,
     .sfp_port[0].sds = 0,
     .sfp_port[0].i2c = { .sda = GPIO41_I2C_SDA3_MDIO1, .scl = GPIO40_I2C_SCL3_MDC1 },
+
+    // SFP port on SDS1 / logical port 8
+    .sfp_port[1].pin_detect = GPIO38,
+    .sfp_port[1].pin_los = GPIO_NA,
+    .sfp_port[1].pin_tx_disable = GPIO_NA,
+    .sfp_port[1].sds = 1,
+    .sfp_port[1].i2c =  { .sda = GPIO39_I2C_SDA4, .scl = GPIO40_I2C_SCL3_MDC1 },
 
     .reset_pin = GPIO_NA,
     .high_leds = { .mux =  LED_28_SYS | LED_29, .enable = LED_27 | LED_28_SYS | LED_29 },
     .port_led_set = { 0, 0, 0, 1, 0, 0, 0, 0, 1},
     .led_sets = {
                     {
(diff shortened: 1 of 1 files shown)


feelfree69:
@kekopop You also own such a switch, I merged your PR recently. Can you have a look at this PR?

kekopop:
Flashed an image with the proposed changes to my Horaco SWTG124AS switch. Unfortunately SFP port 1 (the outward one, numbered 6 on the enclosure) does not work with it. It however seems to detect the SFP module to some extent:

image

Perhaps the SWTG024AS-A 2.0.1 comes in different configurations depending on switch brand etc?

For the moment however, I can't do a lot more than just simple testing. Got very limited time to really dive into this and gain the necessary knowledge to deal with it in a more scientific way the coming month or so. 

Attaching a dump of the Horaco SWTG124AS stock firmware in case @zectaiga wants to check for differences in the meantime.

If there's something specific you want me to check, I'm happy to oblige.

[W25Q16JV@SOIC8.zip](https://github.com/user-attachments/files/28428575/W25Q16JV%40SOIC8.zip)

zectaiga:
I flashed the attached Horaco stock firmware and took a register dump. (One note: the Horaco dump was 0x110 bytes larger than the flash image size, so I trimmed that extra tail before flashing it.)

The register dump result looks equivalent to the dump from my LG-SG4T2 stock firmware. The SFP/LED/SDS-related registers I checked are the same. The only difference I found was one bit in the GPIO input register 0x0048, which looks like a momentary GPIO input state rather than a configuration difference.

I also compared the full flash dump against my LG-SG4T2 firmware dump. The core runtime area is almost identical; the only difference there was a 15-byte random-looking WebKit multipart boundary string. The other differences were in per-device/config areas such as MAC address, factory model string, IP/gateway settings, and the logo/upload area.

So my current impression is that the Horaco HC-SWTG124AS firmware and the LG-SG4T2 firmware are essentially the same core firmware, with mostly branding and device-specific configuration differences. At least from these dumps, I don't see evidence that the Horaco SWTG124AS stock firmware uses different SFP/LED/SDS settings from my unit.


One practical note for testing RTLPlayground builds: please run `make clean` before building for a different machine target.

At the moment the build output directory is shared between machine targets. If you build one machine and then build another without cleaning, some object files from the previous target may be reused. That can produce a firmware image with the new filename/target but partially old machine-specific code inside. I hit this locally while switching between targets.

Example:
```sh
make clean
make SHELL=/bin/bash MACHINE=PCB_SWTG024AS_A_2_0_1
```

feelfree69:
> At the moment the build output directory is shared between machine targets

Sounds like a bug. Will make a PR tomorrow which seperates output directories.
Done. See #251.

kekopop:
> I flashed the attached Horaco stock firmware and took a register dump. (One note: the Horaco dump was 0x110 bytes larger than the flash image size, so I trimmed that extra tail before flashing it.)
> 
> The register dump result looks equivalent to the dump from my LG-SG4T2 stock firmware. The SFP/LED/SDS-related registers I checked are the same. The only difference I found was one bit in the GPIO input register 0x0048, which looks like a momentary GPIO input state rather than a configuration difference.
> 
> I also compared the full flash dump against my LG-SG4T2 firmware dump. The core runtime area is almost identical; the only difference there was a 15-byte random-looking WebKit multipart boundary string. The other differences were in per-device/config areas such as MAC address, factory model string, IP/gateway settings, and the logo/upload area.
> 
> So my current impression is that the Horaco HC-SWTG124AS firmware and the LG-SG4T2 firmware are essentially the same core firmware, with mostly branding and device-specific configuration differences. At least from these dumps, I don't see evidence that the Horaco SWTG124AS stock firmware uses different SFP/LED/SDS settings from my unit.
> 
> One practical note for testing RTLPlayground builds: please run `make clean` before building for a different machine target.
> 
> At the moment the build output directory is shared between machine targets. If you build one machine and then build another without cleaning, some object files from the previous target may be reused. That can produce a firmware image with the new filename/target but partially old machine-specific code inside. I hit this locally while switching between targets.
> 
> Example:
> 
> ```shell
> make clean
> make SHELL=/bin/bash MACHINE=PCB_SWTG024AS_A_2_0_1
> ```

If it matters somehow, it's a raw flash dump done with an external programmer. Did it as a backup before booting the switch for the first time.

Always running make clean before building out of habit, to avoid potential problems caused by not doing it :)

Anyways, gave it another go this morning. It seems it might be some combination of RTLPlayground, the SFP transceivers, and/or the MxL 86282C based switch in the other end causing problems rather than your changes. The link came up after power cycling the MxL switch, and worked just fine with normal throughput for 15 mins or so before things started to behave strangely, the link lagging terribly/dropping out and so on. 

Flashed a build without your changes for comparison, and this time it too behaved the same, as opposed to the other day.

Troubleshooting this and pinpointing the actual problem is out of my league. Since it doesn't seem to be related to your changes in any case, I'd suggest merging your PR as is.

feelfree69:
Thanks you @zectaiga, @kekopop for all your contributions and effort.
Will merge.

PR #249, opened 2026-05-30
Fix SFP led for 2.5gbit for MACHINE_SWGT024_V2_0

vDorst:
While testing #246, I noticd that SFP did not light-up while forcing 2.5gbit link. This is because RJ45 and SFP share the same LED-settings. But RJ45 has two leds-settings.
Solution is to add settings for SFP for both managed and unmanaged variant.


feelfree69 (approved):
Can't test it on my hardware, but code is looking good.

logicog:
Looks good to me, too! Nice work with the LED-sets!

PR #251, opened 2026-05-30
make: seperate output directories by MACHINE

feelfree69:
See [this](https://github.com/logicog/RTLPlayground/pull/248#issuecomment-4585092444) comment.

When compiling images for different machines, things could easily screwed up.

This PR adds a $(MACHINE) output directory to prevent that.


vDorst:
Thanks

PR #252, opened 2026-06-01
Don't disable flash DIO operation before flashing

feelfree69:
After playing with the SPI flash, I wondered if switching to Non-DIO operation is still needed. I once added this beacuse it did not work otherwise.
However, for some reason, it DOES work now without switching the mode, therefore removing the lines.


vDorst:
I merged this. It is hard to test and it is still a bit unclear how SPI controller works without documentation.
So we doing to see if this helps.

PR #253, opened 2026-06-02
Fix conf_overwrite filter removing vlan mgmt entries

Erdnusschokolade:
Follow-up to #237 (the vlan/mgmt lookahead fix). The pattern separation
was correct, but the conf_overwrite filter still used
`startsWith("vlan N ")`, which also matched `vlan N mgmt` — so the
management entry got dropped whenever a VLAN's membership was re-saved.

One-line guard (`!item.endsWith(" mgmt")`) so mgmt entries survive.

### DIFF
--- a/html/config.js
+++ b/html/config.js
@@ -14,6 +14,7 @@ const conf_cmds = [
   /^pvid\s+\d{1,2}\s+\d{1,4}$/,
   /^ingress(\s+\d{1,2}[tua])+$/,
   /^ingress\s+[tua]$/,
+  /^port\s+\d{1,2}\s+(10m|100m|1g|2g5|5g|10g|auto|on|off)(\s+(half|full))?$/,
   /^port\s+\d{1,2}\s+name\s+\S+$/,
   /^eee(\s+\d{1,2})?\s+(on|off)$/,
   /^mirror(\s+\d{1,2})(\s+\d{1,2}[tr]?)+$/,
@@ -36,6 +37,7 @@ const conf_overwrite = [
   /^vlan\s+\d{1,4}(?!\s+mgmt\b)/,
   /^pvid\s+\d{1,2}\b/,
   /^ingress\b/,
+  /^port\s+\d{1,2}(?!\s+name\b)/,
   /^port\s+\d{1,2}\s+name\b/,
   /^eee\s+\d{1,2}\b/,
   /^eee\b/,
@@ -71,7 +73,7 @@ function parseConf(s){
         let m = line.match(x);
         let matchStr = m[0];
         configuration = configuration.filter(item =>
-          !(item === matchStr || (item.startsWith(matchStr + " ") && !item.endsWith(" mgmt"))));
+          !(item === matchStr || (item.startsWith(matchStr + " ") && !item.endsWith(" mgmt") && !item.startsWith(matchStr + " name "))));
         break;
       }
     }
--- a/html/config.js
+++ b/html/config.js
@@ -71,7 +71,7 @@ function parseConf(s){
         let m = line.match(x);
         let matchStr = m[0];
         configuration = configuration.filter(item =>
-          !(item === matchStr || item.startsWith(matchStr + " ")));
+          !(item === matchStr || (item.startsWith(matchStr + " ") && !item.endsWith(" mgmt"))));
         break;
       }
     }
--- a/html/config.js
+++ b/html/config.js
@@ -59,8 +59,7 @@ function parseConf(s){
     const deleteMatch = line.match(/^vlan\s+(\d{1,4})\s+d$/);
     if (deleteMatch) {
       const prefix = "vlan " + deleteMatch[1] + " ";
-      configuration = configuration.filter(c =>
-        c !== "vlan " + deleteMatch[1] && !c.startsWith(prefix));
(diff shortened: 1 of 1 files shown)


vDorst:
I am trying to test is bit more in Rust so I can make more unit tests.
But I seems to fail to replicate the current function.

But I think `parseConf()` has already some issues.

1.
```js
 /^vlan\s+\d{1,4}\s+mgmt$/,
 /^vlan\s+\d{1,4}(?!\s+mgmt\b)/,
```
I think can be combined to `/^vlan\s+\d{1,4}(\s+mgmt)?$/`.
Instead of `!item.endsWith(" mgmt")`, use the capture data to see if it is mgmt, so I think it can be replaced width `m[1].trim == "mgmt"`, not tested so please check. 
2. Line 61 `c !== "vlan " + deleteMatch[1]` is never hit, because such lines, r.g. `vlan <n>` alone, are never added to the `configuration`-list.
3. Sequence `vlan 1 mgmt`, `vlan 2 mgmt` is not reduced to `vlan 2 mgmt`.

Erdnusschokolade:
Thanks for the detailed review @vDorst. I tried your suggestions and tested the parser behaviour:

**Re 1 (combining the regexes):** I went with `/^vlan\s+\d{1,4}(\s+mgmt)?$/` and the `m[1]` capture check, but it doesn't hold up:

- The `$` anchor means the combined pattern no longer matches membership lines like `vlan 1 5t 6t`, so *no* `conf_overwrite` entry matches them and the old member-port lines accumulate instead of being replaced. The second pattern is intentionally left without `$` precisely so `m[0]` yields the `"vlan N"` prefix needed to overwrite membership.
- Replacing `!item.endsWith(" mgmt")` with the capture check affects the whole loop, not just the vlan case — the filter is shared by all patterns. For `ip`/`gw`/`netmask`/`port`/… `m[1]` is `undefined`, so those overwrites break too (old + new entry both kept).

So I'd prefer to keep the two separate patterns as they are.

**Re 2 (dead `c !== "vlan " + deleteMatch[1]`):** confirmed — a bare `vlan N` never matches `conf_cmds`, so it's never stored and that sub-expression can't be hit. Happy to drop it.

**Re 3 (`vlan 1 mgmt`, `vlan 2 mgmt` not reduced):** confirmed, but it's harmless in practice — `management_vlan` is a single variable (`rtlplayground.c`), so on replay the last line simply wins; the runtime state is always correct. It's purely cosmetic, and the web UI never emits a `vlan N mgmt` command anyway (only via CLI). If you consider this undesirable behaviour I could test

```js
if (/^vlan\s+\d{1,4}\s+mgmt$/.test(line))
  configuration = configuration.filter(it => !/^vlan\s+\d{1,4}\s+mgmt$/.test(it));
```

as a fix.

**Separate issue I ran into while testing:** port speed isn't persisted at all. `conf_cmds` only contains `port N name`, so `port N <speed>` lines are treated as unknown and dropped on save — configured speeds don't survive. Fix is small: add the speed command (with optional `half`/`full` suffix) to `conf_cmds`, add a `port N` entry to `conf_overwrite`, and guard the per-port `name` entry so a speed change doesn't wipe the port name. **Should I fold that into this PR, or open a separate one?**

vDorst:
Re1: Thanks for the explanation why it can´t work. `m[1]`is `undefined` needs probably an extra check. But is it fine to keep the old behavior.

Re3: Because we already trying to reduce and sanitize the config, I think it is a nice addition.

port speed: I am fine to add it to this PR.

Erdnusschokolade:
Thanks @vDorst — agreed on all points. Updated this branch accordingly:

- **Re 1:** kept the two separate vlan patterns as they are.
- **Re 2:** dropped the dead `c !== "vlan " + deleteMatch[1]` sub-expression.
- **Re 3:** added the single-management-VLAN reduction. When a `vlan N mgmt` line is parsed, any previously stored `vlan M mgmt` entry is dropped, so repeatedly changing the management VLAN no longer accumulates stale lines:

  ```js
  // Only one management VLAN can be active, so drop any previous mgmt entry
  if (/^vlan\s+\d{1,4}\s+mgmt$/.test(line))
    configuration = configuration.filter(item => !/^vlan\s+\d{1,4}\s+mgmt$/.test(item));
  ```

  Verified that membership overwrite and mgmt survival on member re-save still behave correctly.

- **Port speed:** folded into this PR. `conf_cmds` only contained `port N name`, so `port N <speed>` lines (with optional `half`/`full` for 10/100M) were dropped on save and speeds didn't survive a reboot. Added the speed command to `conf_cmds`, a matching `port N` entry to `conf_overwrite`, and a guard so a speed change no longer wipes the port name.

Erdnusschokolade:
One more note on the port-speed change, regarding duplex specifically:

Duplex is persisted only as the optional `half`/`full` suffix of the speed line, not as a standalone `port N duplex` command. That mirrors the web UI, which only ever emits the combined form, the speed dropdown bundles speed + duplex (e.g. `100m half`). Keeping duplex tied to the speed line avoids having a separate `port N duplex` entry and the speed line overwrite each other, since both would reduce to the same `port N` key in `conf_overwrite`.

vDorst:
That is fine.

Thanks a lot for the work.

Issue #256, opened 2026-06-07
Firmware update through WebUI seems broken on commit 94e34a5


bennydiamond:
Built using command `make MACHINE=LIANGUO_ZX_SWTGW215AS`

Used generated BIN to flash from earlier revision (couple of weeks old).

I changed default config and rebuilt. 

Tried to upload that new bin through web UI. Serial is stuck at 

```
> tcp: found no unused connections.Firmware upload started.
```

Serial interface is stuck at that. 
Serial interface seems unresponsive.
WebUI is unresponsive.
Pinging device fails.

Power cycling the device, does trigger firmware update which  fails:

```
rtl8372_init done
Flash information:
  Manufacturer ID: 0xef
  Memory Type:     0x40
  Capacity:        0x14 = 1 MB
1 MB flash size detected. (1 MB is needed for image updating)
Checking for update image in flash... found update image!
Checking integrity................................................................Checksum incorrect, please upload the image again
Erasing bad uploaded flash image
```

Browsers tried:
Firefox
Chromium
Brave

Also tried to upload bin through curl, same behavior as with browsers.

Example command:
`❯ curl -sS -i -H "Cookie: session=a5bc0a392c65" -F "firmware=@output/LIANGUO_ZX_SWTGW215AS/rtlplayground-v0.1.0-94e34a5-dirty-LIANGUO_ZX_SWTGW215AS.bin;type=application/octet-stream" http://192.168.0.4/upload
`

I'm out of ideas on how to flash my device, other than desoldering the flash!

logicog:
Hi!

```
> tcp: found no unused connections.Firmware upload started.
```
is harmless, it just means that not all concurrent TCP connections opened by the browser were served, but it will retry. Could you, however, post a complete serial console log of the update process, starting with the POST request for the upload to see what was going on when the update image was written into the flash?

feelfree69:
> Used generated BIN to flash from earlier revision (couple of weeks old).

When you update firmware, the update code actually running is the one from the **old** firmware.
That's why it's more interesting what is the old version already running than the new one.


> I'm out of ideas on how to flash my device, other than desoldering the flash!

You don't have clamp with an external programmer?
So you flashed RTLPlayground once directly from stock firmware using the image build in /installer directory? And never updated that image since then?

Then you maybe have the same issue as #193, which is resolved with #214.
But you'll need an external flasher in any case.

bennydiamond:
Just to be clear.

[...]

PR #257, opened 2026-06-07
Revert "Don't disable flash DIO operation before flashing"

TylerDurden-23:
Reverts logicog/RTLPlayground#252 as per: #256


logicog:
Looks good. Thanks for the fast help!

PR #259, opened 2026-06-15
Add support for 100M on RTL8221B ports

logicog:
On machines with 5 Ethernet ports, the 5th port is implemented using an RTL8221B-PHY, which is 2.5. The RTL8221B supports 100M by configuring SGMII on the link to the MAC. In RTLPlayground, this port currently only supports 100M, if previously SGMII was configured, i.e. a 1GBit cable was attached. Make this work also for the case where previously a 2.5Gbit cable was attached.


vDorst:
thanks

PR #260, opened 2026-06-17
Add KP-9000 6XH and 6XHML PCB-revisions to Machine targets

donbernhardo:
## Summary

This PR splits the Keeplink KP-9000 6XH / 6XHML two-SFP targets by PCB revision and updates the documentation so users can select the correct machine target based on the PCB silkscreen instead of the external case label.

The PR discussion showed that the LED mismatch was not caused by a managed vs. unmanaged SKU difference. Instead, these devices exist with different `2M-PCB43` hardware revisions that use different wiring.

## Hardware revisions

Known PCB revisions:

* `2M-PCB43-V1.1`

  * Reported in Mokerlink 2G040210GSM web-managed 4+2 switches.
  * Uses the same V1.x layout as V1.2.

* `2M-PCB43-V1.2`

  * Verified by logicog on a `KP-9000-6XHML-X2`.
  * Uses the existing `KP_9000_6XHML_X2` wiring from `main`.

* `2M-PCB43-V2.1`

  * Verified on my `KP-9000-6XH-X2`.
  * Also reported for `KP-9000-6XHML-X2`.
  * Uses the existing `KP_9000_6XH_X2` wiring from `main`.

## Machine targets

This PR adds explicit PCB-revision based targets:

* `MACHINE_KP_9000_6XH_X2_V1_1`
* `MACHINE_KP_9000_6XHML_X2_V1_1`
* `MACHINE_KP_9000_6XH_X2_V1_2`
* `MACHINE_KP_9000_6XHML_X2_V1_2`
* `MACHINE_KP_9000_6XH_X2_V2_1`
* `MACHINE_KP_9000_6XHML_X2_V2_1`

Legacy compatibility targets are preserved:

* `MACHINE_KP_9000_6XHML_X2` continues to use the V1.x wiring.
* `MACHINE_KP_9000_6XH_X2` continues to use the V2.1 wiring.

This avoids breaking existing deployments while giving new users explicit board-revision choices.

## Rationale

The machine definition should describe the actual PCB wiring: port map, SFP detect pins, I2C buses, SerDes mapping, reset GPIO and LED mux.

The external `6XH` / `6XHML` label, or managed/unmanaged distinction, is not suitable to infer the hardware layout. Users should therefore select the target based on the PCB silkscreen.

## Documentation

Updated the KP-9000 6XH device documentation to:

* list the known PCB revisions,
* show which machine target belongs to each board,
* include both V1.x and V2.1 photo sets,
* document that ML/non-ML labels alone do not identify the wiring,
* update the supported hardware page accordingly.

## Verification

Built and checked:

```text
make machine_check
make MACHINE=KP_9000_6XH_X2_V1_1
make MACHINE=KP_9000_6XHML_X2_V1_1
make MACHINE=KP_9000_6XH_X2_V1_2
make MACHINE=KP_9000_6XHML_X2_V1_2
make MACHINE=KP_9000_6XH_X2_V2_1
make MACHINE=KP_9000_6XHML_X2_V2_1
```

Runtime verification on `2M-PCB43-V2.1`:

* firmware reachable on the configured IP,
* four RJ45 ports mapped correctly to logical ports 4, 5, 6 and 7,
* left and right SFP module detect works,
* SFP I2C EEPROM reads work,
* stable switch operation in production environment observed


logicog:
I have one of these devices (and hope that I did not put it in production somehwere). Will test tomorrow. I am a bit surprised that the LEDs are wrong, though, because I had thought that I tested it thoroughly. Maybe it is a different board version. Did you check that, there are images of the PCB in the docs section?

donbernhardo:
Thanks for looking into this.
My device seems to have the exact same PCB version as in the [documentation](https://github.com/logicog/RTLPlayground/blob/main/doc/devices/KP-9000-6XH-X2.md). HW V2.1
Here is a picture from the Silkscreen of my board
Switch-PCB

feelfree69:
hmm, i think it was me who added the device.
seems like i have not looked at the leds so exactly....
will also test tomorrow.

feelfree69:
Ah, my Device is a KP_9000_6XH_X2, and this PR is about a KP-9000-6XH**ML**-X2.
So I can't test it.
But at least I can confirm that the LED mapping for KP_9000_6XH_X2 is correct.

logicog:
I just verified the LEDs using the exisiting profile for the KP-9000-6XHML-X2, and it works perfectly. And then I saw the version number on the PCB. It is 2M-PCB43-V1.2. Not V2.1 as in your photo. You can see it also in the existing device-photos, also that the PCB has indeed changed. Could you add your device and also update the documentation?

feelfree69:
The PCB43-V2.1 is also used in KP_9000_6XH_X2, maybe this machine definition could also be used for the ML version with that PCB.

logicog:
In fact I was running all the time that machine configuration on the device. But the difference is that the managed switch has a reset button, the KP_9000_6XH_X2 does not.

feelfree69:
> In fact I was running all the time that machine configuration on the device. But the difference is that the managed switch has a reset button, the KP_9000_6XH_X2 does not.

Yes, but I added the pin definition for the reset button also for the unmanaged device.

logicog:
I am not sure how to proceed. For the docs: maybe just have the 6XHML-X2 and mention that there is an unmanaged variant and another PCB with the respective pictures. In the machine definitions, maybe have all 3 defined as MACHINE types and make the _6XH_X2 and alias to the 6XHML. In theory one can make the button dependent on the machine-type. Just a few more ifdefs...

th0m4sek:
PXL_20260616_141216759
PXL_20260616_141223882
PXL_20260616_142911158
Wanted to wait for programer and then start issue but my horaco swtg218as is swtg018as.
So maybe doxumentation should be based around PCB's and devices should only propose what PCB can be in device...

feelfree69:
> Wanted to wait for programer and then start issue but my horaco swtg218as is swtg018as. So maybe doxumentation should be based around PCB's and devices should only propose what PCB can be in device...

You're probably right that there is a lot of confusion, based on different namings on devices and PCBs.
But you're case has nothing to do with this PR, which handles totally different devices.

feelfree69:
> I am not sure how to proceed.

If there are 2 devices with different PCBs out there, I would simply add the PCB revision, as we have it already for the 9-port devices:
```
// #define MACHINE_KP_9000_9XHML_X_V2_2
// #define MACHINE_KP_9000_9XHML_X_V3_1
```

donbernhardo:
Ok now things became a lot clearer. Here is my suggestion:
I will create an update to this PR defining the hardware revision specific configurations just like you mentioned. The repo does that already with other devices. I will also keep the legacy definitions without hardware revision to not break existing deployments.

Since the managed versions of that switch seems to be the exact same PCB as the unmanaged I also consider making all the "MACHINE_KP_9000_9X**ML**... configurations to be just an alias for the non-managed configurations.

Please standby for PR update :)

donbernhardo:
Ok this is good to go now.
I have a KP-9000-6XHML-X2 and a KP-9000-6XH-X2 and both have the 2M-PCB43-V2.1 PCB inside. Other then the reset switch being not populated on the unmanaged version, there is no difference. I tested on both and can confirm that it is working now. Since I can not test the HW V1.2 PCB I kept everything for 1.2 as it was.

Please consider testing and merging this.

riker77:
There seems to be an even older revision of this PCB.
I've got a Mokerlink 2G040210GSM web managed 4+2 Switch.
Its PCB revision  is 2M-PCB43-V1.1.

With the current KP_9000_6XHML_X2 machine definition, all LEDs seem to work fine (short 1G, 2.5G and 10G-T SFP test).

donbernhardo:
Thanks! I think I can rework that so it will cover all the PCB Versions and keep current compatibility. I will update the PR so the machine selection is based on the PCB revision, not on the 6XH_X2 vs 6XHML_X2 label.

The plan is:

- keep the existing legacy targets unchanged for compatibility
- treat 2M-PCB43-V1.1 and 2M-PCB43-V1.2 as the same V1.x layout
- treat 2M-PCB43-V2.1 as the V2.1 layout
- add explicit PCB-revision targets for both 6XH_X2 and 6XHML_X2
- update the documentation table accordingly

So the ML/non-ML naming will only describe the label/SKU, while the actual hardware definition will follow the PCB silkscreen.

donbernhardo:
Updated this PR with the latest KP-9000 PCB-revision mapping logic.

Summary:
- Added explicit PCB-revision machine targets for both `6XH_X2` and `6XHML_X2` variants.
- Kept the legacy aliases unchanged:
  - `MACHINE_KP_9000_6XHML_X2` continues to use the existing V1.x layout.
  - `MACHINE_KP_9000_6XH_X2` continues to use the existing V2.1 layout.
- Made V1.1 and V1.2 explicit targets share the existing V1.x layout.
- Made V2.1 explicit ML/non-ML targets share the existing V2.1 layout with the custom LED mux.
- Updated the KP-9000 device docs and supported-device overview to explain that PCB silkscreen revision, not ML/non-ML labeling, selects the hardware layout.

Files changed:
- `machine.h`
- `machine.c`
- `doc/devices/KP-9000-6XH-X2.md`
- `doc/supported_devices.md`

Verification passed:
- `make machine_check`
- `make MACHINE=KP_9000_6XHML_X2`
- `make MACHINE=KP_9000_6XH_X2`
- `make MACHINE=KP_9000_6XH_X2_V1_1`
- `make MACHINE=KP_9000_6XHML_X2_V1_1`
- `make MACHINE=KP_9000_6XH_X2_V1_2`
- `make MACHINE=KP_9000_6XHML_X2_V1_2`
- `make MACHINE=KP_9000_6XH_X2_V2_1`
- `make MACHINE=KP_9000_6XHML_X2_V2_1`

feelfree69:
> Kept the legacy aliases unchanged:
> 
>     * `MACHINE_KP_9000_6XHML_X2` continues to use the existing V1.x layout.
> 
>     * `MACHINE_KP_9000_6XH_X2` continues to use the existing V2.1 layout.

If it were up to me, I would omit these "legacy" aliases. It adds more confusion than it solves.

donbernhardo:
I agree but I think this should be decided by the maintainer. I did not want to break anything just in case. I would remove them if requested.

feelfree69:
> I agree but I think this should be decided by the maintainer. I did not want to break anything just in case. I would remove them if requested.

Any other opinions on having double definitions with identical contents, just for backward compatibiliy?

feelfree69 (approved):
Apparently no other comments or opinions.
-> will merge.
Thanks a lot!

Issue #261, opened 2026-06-20
LG-SG5T1


Lynn-Becky:
I have successfully compiled the firmware for this model.This PCB is highly similar to the ZX-SG4T2, with the only differences being the removal of one SFP port, which was replaced by an RTL8221B and an RJ45 Ethernet port. The LEDs and display are both working normally.This model is unmanaged out of the factory. Additionally, the status of the 5th RJ45 port can be seen in the management panel, but it does not participate in packet forwarding.
I have submitted a PR with my code modifications and would greatly appreciate it if everyone could help review it.
PCB Board : PCB-SWTG024AS-V2.0_16895

<img width="3072" height="4080" alt="Image" src="https://github.com/user-attachments/assets/759524cd-1f4b-4b54-aad4-9c6d233c0c49" />

<img width="3072" height="4096" alt="Image" src="https://github.com/user-attachments/assets/9b9960db-de98-453b-8af3-ae71f05fe536" />

<img width="3072" height="4096" alt="Image" src="https://github.com/user-attachments/assets/556d22cd-3831-4355-a5c7-b4c9accc5ec7" />

<img width="3072" height="4096" alt="Image" src="https://github.com/user-attachments/assets/67ca4fb9-f974-4a2b-8aeb-ca7214c81d15" />

<img width="4096" height="3072" alt="Image" src="https://github.com/user-attachments/assets/dc2900ce-af12-4a90-98e0-fef8908e33b7" />

<img width="1398" height="765" alt="Image" src="https://github.com/user-attachments/assets/578234f5-7719-4d00-a88a-d6e6d2ed45a6" />

PR #263, opened 2026-06-20
Add PCB-SWTG024AS V2.0 

Lynn-Becky:
Add PCB-SWTG024AS-V2.0_16895 , used in LG-SG5T1
#261


logicog:
I would not listen to Copilot regarding the machine name. Without the suffix is better. Do you have pictures of the PCB?

Lynn-Becky:
> I would not listen to Copilot regarding the machine name. Without the suffix is better. Do you have pictures of the PCB?

IMG_20260620_231515
IMG_20260620_231510

Lynn-Becky:
> I would not listen to Copilot regarding the machine name. Without the suffix is better. Do you have pictures of the PCB?

IMG_20260621_010849
​I tried to fix the switching issue on the 5th RJ45 port, but no way i have to proceed it.

machine.c
```diff
@@ -896,6 +896,56 @@ void machine_custom_init(void)
...
+                            LEDS_1G | LEDS_100M | LEDS_10M | LEDS_LINK | LEDS_ACT,
+                            LEDS_DUPLEX,
+                            LEDS_2G5 | LEDS_LINK | LEDS_ACT
+                    },
+                    {
+                            LEDS_2G5 | LEDS_1G | LEDS_100M | LEDS_LINK | LEDS_ACT,
+                            LEDS_10G | LEDS_LINK | LEDS_ACT,
+                            LEDS_2G5 | LEDS_LINK,
+                            LEDS_COL | LEDS_DUPLEX
+                    },
```
vDorst, line 934: Can you add use comments, which port type and led colour this set is driving?

vDorst (commented):
See review comment.

Also can you also add basic documentation and pictures about this device?

This is a detailed example: https://github.com/logicog/RTLPlayground/blob/main/doc/devices/SWTG024AS.md
But a basic one is also fine like: https://github.com/logicog/RTLPlayground/blob/main/doc/devices/ZX310S-4T2XH.md.

For the picture, rotate the images so the ethernet ports a facing down and crop the image so only the PCB is visible and less of the table/surroundings.

machine.c
Lynn-Becky: done

vDorst:
thanks

PR #265, opened 2026-06-21
Add software-controlled PoE for RTL8238B boards (register-based, generic per-machine layer)

tobik312:
Adds Power-over-Ethernet support for RTL8373 boards whose PSE is an **RTL8238B** on I2C,
with a generic, opt-in per-machine PoE layer. Tested on **keepLink KP-9000-9XHPML-X V3.1**
(8× 2.5G PoE+ + 1× 10G SFP+).

## How it works — and how it differs from existing drivers

Other open PoE drivers (realtek-poe, the OpenWrt `realtek-pse-*` driver) talk to these
PSEs with a 12-byte **host-command** protocol answered by a companion management MCU. On these
boards that MCU layer does not respond — the RTL8238B is reachable only as a raw I2C **register
slave**.

So this driver does everything through **direct register I/O**, exactly as the OEM web UI does:
- uploads the volatile PSE application firmware at boot (a bit-banged I2C stream on GPIO46/47,
  because the image far exceeds the SoC I2C engine's 16-byte/transfer limit),
- enables ports through the chip's admin registers,
- reads telemetry (power-on / class / voltage / current) straight from the register window.

Per-port enable was verified on hardware to be real **auto-detect** (no running MCU app required):
the chip powers only genuine PDs — a Class 4 PD read **51 V / 153 mA**, and an enabled-but-empty
port stays unpowered.

## The register interface (reverse-engineered)

The RTL8238B register file is undocumented; the layout below was reverse-engineered from the OEM
software — the stock firmware's own PoE bring-up code, and the register accesses its web UI makes —
then verified on hardware. Access is plain little-endian I2C; each controller drives four port
channels (`local 0–3`):

- **Bring-up:** `0x1a` = whole-chip reset; arm download mode on `0xec` (payload `03 00 00 00`,
  read-back bit 2 = armed) with an alternate arm on `0x18` (`00 00 20 39`); stream the image to
  data port `0xf4`; `0xec` bit 5 = image accepted.
- **Control:** per-port admin mode is a 2-bit field in `0x12` (byte 2 of the `0x10` block) —
  `3` = on (auto-detect), `0` = off — plus a 1-bit enable in `0x14`.
- **Telemetry:** Power-On status is `0x10` byte 0 bit `4+local`; PD class is the high nibble of
  `0x0c+local`; per-port voltage/current is `0x30+4*local` (volts = top byte × 0.5 V,
  mA = low 16 bits × 125 / 4096).

Full register map + decode in `doc/poe-rtl8238b.md`.

## Generic & opt-in

PoE is described in the machine descriptor (`struct machine.poe` = chip / I2C addresses / port
count) and gated behind `POE_PRESENT`: non-PoE machines compile **no** PoE code at all (~6 KB of
BANK2). Adding another RTL8238B board is descriptor-only; a different PSE chip is a new
`poe_<chip>.c` behind the same chip-agnostic interface (`poe.h`).

## Interfaces

- **Console:** `poe load | port <n> <on|off> | global <on|off>` (chip-agnostic).
- **Web:** a PoE page showing total consumption and a per-port table (state / class / W / V / mA)
  with per-row and global enable/disable.
- **API:** `GET /poe.json` returns the driver-normalized per-port status
  (`port, admin, on, class, v, ma`); it takes no query parameters — the controller is never
  exposed over HTTP.

image

## Firmware image

The RTL8238B's PoE application firmware is volatile and Realtek/OEM-proprietary, so it is **not**
committed. You extract it from your own device's OEM firmware with
`tools/poe/extract_rtl8238b_image.py`, and the Makefile embeds it. An OEM dump can bundle PSE
firmware for several chips; the tool writes the RTL8238B image (identified by size) to
`tools/poe/pse_image.bin`. See `doc/poe-rtl8238b.md`.

## Notes

- The PoE page and nav link live in the shared `html_data`, so they're present in every build;
  on a non-PoE machine the page just shows a "no PoE controller" notice (the `/poe.json` route is
  compiled out). HTML assets are machine-agnostic by design.
- The bring-up internals (the `0x20`/`0x21` two-controller topology, the per-port register
  tables) are hardcoded in the driver as the chip's standard layout; `machine.poe` parameterizes
  the I2C addresses and port count. A board with a non-standard RTL8238B topology would extend the
  driver.

## Testing

- Builds clean for the PoE machine (`KP_9000_9XHPML_X_V3_1`) and for a non-PoE machine (PoE fully
  excluded).
- On hardware: PoE comes up automatically at boot, PDs power up, per-port and global
  enable/disable work, and the web page shows live telemetry (verified Class 4 @ 51 V / 153 mA).


logicog:
> The RTL8238B register file is undocumented; the layout below was reverse-engineered from the OEM
> software — the stock firmware's own PoE bring-up code, and the register accesses its web UI makes —
> then verified on hardware. Access is plain little-endian I2C; each controller drives four port
> channels (`local 0–3`):

This was probably an enormous amount of work. Amazing! I am just browsing though the code, it is great! It will take a bit to go through it, so please bear with all of us a bit.

I so far went through the commits of this PR and I think you are splitting it absolutely the right way, so that is already cool.

poe.h
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+/* Bring PoE up: the driver does whatever its chip needs (e.g. upload firmware + enable ports).
+ * Run at boot (post-PHY) and on `poe load`; afterwards the controller powers detected PDs. */
+void poe_bringup(void) __banked;
+
+/* Enable/disable PoE on a single port (`poe port <n> <on|off>`). */
+extern __xdata uint8_t poe_port_num;	// port 0-based
+extern __xdata uint8_t poe_port_on;	// 1 = enable, 0 = disable
+void poe_port_set(void) __banked;
```
logicog: I believe it is possible to use function arguments, here, without using additional IMEM. 2 bytes as arguments are normally not a problem. So
```
void poe_port_set(uint8_t port_num, uint8_t port_on) __banked;
```
should work.
Normally, the problems start for banked functions with 3 bytes or a short and a byte.

poe.h
```diff
@@ -0,0 +1,40 @@
...
+
+/* Enable/disable ALL ports at once (`poe global <on|off>` + the web UI's "All PoE"). */
+extern __xdata uint8_t poe_global_on;	// 1 = enable all, 0 = disable all
+void poe_global_set(void) __banked;
```
logicog: Same here, it should be possible to use 1 argument.

poe_rtl8238b.c
```diff
@@ -0,0 +1,796 @@
...
+#ifdef POE_CHIP_RTL8238B
+
+extern __xdata uint8_t sfr_data[4];
+extern __xdata struct flash_region_t flash_region;
+extern __code struct machine machine;
+
+// Global interrupt-enable register (IE, SFR 0xA8); EA = bit 7. All interrupts are
+// disabled (CLR EA) for the entire image upload so no ISR can mis-time the
+// bit-banged I2C stream.
+__sfr __at(0xA8) POE_IE;
```
logicog: Is this a general SFR register of the 8051? If yes, could this go into the rtl837x_sfr.h file? It looks like that register is not yet declared, and then we would know in one place what all these registers do. If it is general, is there maybe a different name for the SFR in general contexts?

I also do not understand the comment about the bit-banging. Aren't you using the normal I2C SFR-registers in e.g poe_raw_read()? But maybe I have not read far enough into the PR...

poe_rtl8238b.c
```diff
@@ -0,0 +1,796 @@
...
+
+	for (poe_i = 0; poe_i < poe_rd_len; poe_i++) {
+		if (!(poe_i & 0x3))
+			reg_read_m(RTL837X_REG_I2C_OUT + poe_i);
+		poe_resp[poe_i] = sfr_data[3 - (poe_i & 0x3)];
+	}
+}
+
+// --- PSE application image upload ---
+
```
logicog, line 136: These are a lot of variables. I understand they implement a state machine for I2C bitbanging. Could one group them into structures? This would reduce clutter of the namespace, and possibly grouping could increase readability. I am a bit overwhelmed...

poe_rtl8238b.c
```diff
@@ -0,0 +1,796 @@
...
+static __xdata uint8_t poe_bback;	// 1 = ACK (SDA low) sampled on the last write byte
+static __xdata uint16_t poe_bbk;	// per-byte delay spin counter
+static __xdata uint8_t poe_bbb;		// byte to shift out (bb_write_byte input)
+static __xdata uint8_t poe_bbv;		// byte shifted in (bb_read_byte output)
+static __xdata uint8_t poe_bbn;		// read loop counter
+static __xdata uint8_t poe_bb_ackout;	// 1 = master sends ACK after read byte (else NACK)
+static __xdata uint8_t poe_rdslave = 0x20;	// slave address bb_read_reg() targets (0x20 / 0x21)
+
+// --- Bit-bang I2C on GPIO46 (SCL) / GPIO47 (SDA), push-pull drive ---
+static void bb_scl(uint8_t hi) __banked
```
logicog: This function is only called from within this file, so it does not need the __banked modifier. Calls within the same bank do not need to be banked, so this would in theory also be true for calls across source files in the same bank, but that is very dangerous, in case source code changes its bank...

poe_rtl8238b.c
```diff
@@ -0,0 +1,796 @@
...
+{
+	if (hi)
+		reg_bit_set(RTL837X_REG_GPIO_32_63_OUTPUT, BB_SCL_BIT);
+	else
+		reg_bit_clear(RTL837X_REG_GPIO_32_63_OUTPUT, BB_SCL_BIT);
+}
+
+static void bb_sda(uint8_t hi) __banked
```
logicog: Same here.

poe_rtl8238b.c
```diff
@@ -0,0 +1,796 @@
...
+{
+	// Push-pull drive of both levels (in GPIO mode the pad pull-up may be off and there
+	// may be no strong external pull-up). SDA is released to input only for the ACK/read bit.
+	if (hi)
+		reg_bit_set(RTL837X_REG_GPIO_32_63_OUTPUT, BB_SDA_BIT);
+	else
+		reg_bit_clear(RTL837X_REG_GPIO_32_63_OUTPUT, BB_SDA_BIT);
+}
+
+static uint8_t bb_sda_read(void) __banked
```
logicog: And here.

poe_rtl8238b.c
```diff
@@ -0,0 +1,796 @@
...
+{
+	reg_read_m(RTL837X_REG_GPIO_32_63_INPUT);
+	return (sfr_data[2] & 0x80) ? 1 : 0;	// GPIO47 = bit 15 = sfr_data[2] bit7
+}
+
+/* Route GPIO46/47 to plain GPIO; both lines push-pull, idle-high. */
+static void bb_setup(void) __banked
```
logicog: Here, too.

poe_rtl8238b.c
```diff
@@ -0,0 +1,796 @@
...
+	sfr_mask_data(1, 0x06, 0x00);	// SDA bus0 function bits -> 0b00 = GPIO
+	reg_write_m(RTL837X_PIN_MUX_1);
+
+	reg_bit_set(RTL837X_REG_GPIO_32_63_OUTPUT, BB_SDA_BIT);		// SDA idle high
+	reg_bit_set(RTL837X_REG_GPIO_32_63_OUTPUT, BB_SCL_BIT);		// SCL idle high
+	reg_bit_set(RTL837X_REG_GPIO_32_63_DIRECTION, BB_SDA_BIT);	// SDA output
+	reg_bit_set(RTL837X_REG_GPIO_32_63_DIRECTION, BB_SCL_BIT);	// SCL output
+}
+
+static void bb_start(void) __banked
```
logicog: Ditto.

poe_rtl8238b.c
```diff
@@ -0,0 +1,796 @@
...
+{
+	bb_sda(1);
+	bb_scl(1);
+	bb_sda(0);		// SDA falls while SCL high = START
+	bb_scl(0);
+}
+
+static void bb_stop(void) __banked
```
logicog: Also.

poe_rtl8238b.c
```diff
@@ -0,0 +1,796 @@
...
+{
+	bb_sda(0);
+	bb_scl(1);
+	bb_sda(1);		// SDA rises while SCL high = STOP
+}
+
+/* Clock out poe_bbb MSB-first, then one ACK clock (sampling SDA into poe_bback).
+ * Parameterless (globals) to keep the 8051 overlay segment from overflowing. */
+static void bb_write_byte(void) __banked
```
logicog: Here, too.

poe_rtl8238b.c
```diff
@@ -0,0 +1,796 @@
...
+	bb_sda(1);							// drive latch high, then release
+	reg_bit_clear(RTL837X_REG_GPIO_32_63_DIRECTION, BB_SDA_BIT);	// INPUT to sample the ACK
+	bb_scl(1);
+	poe_bback = bb_sda_read() ? 0 : 1;
+	bb_scl(0);
+	reg_bit_set(RTL837X_REG_GPIO_32_63_DIRECTION, BB_SDA_BIT);	// resume push-pull drive
+}
+
+/* Read one byte MSB-first into poe_bbv; master drives ACK if poe_bb_ackout else NACK. */
+static void bb_read_byte(void) __banked
```
logicog: Also needs no __banked.

poe_rtl8238b.c
```diff
@@ -0,0 +1,796 @@
...
+	bb_scl(1);
+	bb_scl(0);
+}
+
+/*
+ * Bit-bang I2C read of poe_rd_len bytes from register poe_reg of slave 0x20 into
+ * poe_resp:  START | 0x40 | reg | reSTART | 0x41 | read (ACK..ACK,NACK) | STOP.
+ * Status read after each download command (write reg, repeated START, read N).
+ */
+static void bb_read_reg(void) __banked
```
logicog: Same.

poe_rtl8238b.c
```diff
@@ -0,0 +1,796 @@
...
+
+/*
+ * One short bit-bang command transaction: START | 0x40 | cmd | <4-byte payload> | STOP.
+ * Command 0xec then 0x18 (each a separate transaction) are issued before streaming the
+ * image; together they switch the controller from register mode into firmware-download
+ * mode, so the subsequent 0xf4 stream is loaded into the MCU instead of a register window.
+ */
+static __xdata uint8_t poe_bbcmd;
+static __xdata uint8_t poe_ack_cmd;	// ACK of the command byte (0xec/0x18) in the last bb_cmd4
+static void bb_cmd4(void) __banked
```
logicog: Same.

poe_rtl8238b.c
```diff
@@ -0,0 +1,796 @@
...
+	bb_stop();
+}
+
+/*
+ * I2C bus recovery: a slave stuck mid-byte (e.g. after a spurious START it never saw
+ * a STOP for) holds SDA low and wedges the bus. Release SDA to input and pulse SCL 9
+ * times so the slave can finish/abandon its byte, then issue a clean STOP. Cheap
+ * insurance against a wedged bus before the download transaction.
+ */
+static void bb_busrecover(void) __banked
```
logicog: Ditto.

poe_rtl8238b.c
```diff
@@ -0,0 +1,796 @@
...
+/*
+ * Per-port config pass, once per port. Read-modify-write of two per-port register fields
+ * on the port's controller:
+ *   - reg 0x10 block: the port's 2-bit mode field in byte 2 (reg 0x12)
+ *   - reg 0x14 block: the port's 1-bit field      in byte 0 (reg 0x14)
+ * Ports 0-3 are on controller 0x20 (field index 3,2,1,0), ports 4-7 on 0x21 (index 0,1,2,3).
+ * poe_74ca_set selects the clear pass (run before the image, disables the fields) or the
+ * set/enable pass (run after the image, powers the ports). Byte 0 = first I2C byte = LSB.
+ */
+static void bb_74ca_prep(void) __banked
```
logicog: Same.

poe_rtl8238b.c
```diff
@@ -0,0 +1,796 @@
...
+			poe_resp[0] |= (0x01 << poe_n);		// set the port's bit = 1 (enable)
+		else
+			poe_resp[0] &= ~(0x01 << poe_n);	// clear it (disable)
+		poe_reg = 0x14;	bb_write_resp4();
+	}
+	poe_rdslave = 0x20;
+}
+
+/* Dump reg 0x10/0x14 (raw 4 bytes) from 0x20 and 0x21 (defined below; used here too). */
+static void pp_dump(void) __banked;
```
logicog: Same.

poe_rtl8238b.c
```diff
@@ -0,0 +1,796 @@
...
+
+/*
+ * Put the controller into firmware-download mode (prep cmds 0xec, 0x18) then stream the
+ * whole PSE image to slave 0x20 by bit-banging GPIO46/47. The image is one START..STOP
+ * transaction:
+ *   START | 0x40 | 0xf4 | <entire image> | STOP
+ * Runs the whole-chip reset and port-disable first. Restores the I2C engine pin-mux on
+ * exit so normal register access (/poe.json) keeps working.
+ */
+void poe_bb_upload(void) __banked
```
logicog: Same.

logicog (commented):
I am by far not through, yet (limited attention span), but for this commit, I believe I made my comments.

vDorst:
Looks great!

- i2c-bus and firmware upload
  Did you also found this datasheet? https://github.com/jameywine/realtek-doc/blob/main/datasheets/RTL8238B%20Host%20Command%20Guide.pdf
  According to `6.10` shows firmware download which only accepts 12-32 bytes per i2c-frame.
  We can generate 16byte frames so the hardware i2c-bus can be used.
  I think this was expected because almost no i2c-host can generate a single 8kbyte i2c-frame.
- poe
  When no poe is present/controllable, I prefer to see that console `poe` command return `no supported`. Not sure what `poe.json` should return. Just a `404` or `400 Bad Request` or `200` with a valid json content.

logicog:
> * Did you also found this datasheet?

I was not aware of your post and also just hunted this down, should have better invested that half hour...

tobik312:
Thanks both! Pushed the fixes, plus merged latest main:

Function arguments (@logicog): poe_port_set(uint8_t port_num, uint8_t port_on) and poe_global_set(uint8_t on); dropped the poe_port_on/poe_global_on globals. poe_port_num stays only as poe_get_port()'s input.

Driver cleanup (@logicog):

SFR is the standard 8051 IE/EA — dropped the private POE_IE and use EA from <8051.h> (as rtlplayground.c/installer.c do). It can't go in rtl837x_sfr.h since both headers are included together, so it'd redeclare SFR 0xA8.
Removed __banked from functions only called within this bank; the four BANK0 entry points keep it.
Grouped the bit-bang state machine into one struct poe_bb_state.
Clarified the comment: EA is cleared only around the bit-bang upload, not the HW-I2C telemetry path (poe_raw_read).
No PoE (@vDorst): the poe command now replies PoE not supported on this machine on non-PoE builds; /poe.json returns 404 there (200 + JSON only on PoE boards).

HW-I2C vs bit-bang (@vDorst): we did try the HW engine — host-cmds and the 0xC0 framed download (§6.10) both just landed as register writes. The working path is the OEM's legacy ec/0x18/0xf4 handshake streaming the whole image as one continuous transaction, which the ≤16-byte engine can't do — so I'd keep bit-bang as the known-good path.

Also merged latest main and resolved the supported_devices.md conflict — the device moved into the new device table (the bulleted list was dropped upstream).

logicog:
I just looked again a the code and it looks now much cleaner. However, I have some problems reading the commits, because you are correcting the previous ones with the later ones. Could you give `git rebase -i HEAD....` a try to clean up the commits by combining them with `f` or similar? I know this is a bit of work, but it would help quite a bit, this is a lot of code to look through. If you are afraid of messing the repository up, you can tar the entire RTLPlayground directory before you start working with rebase and if you get lost, you can always go back again to the beginning, or wherever you made checkpoints, again using tar.

tobik312:
I rebased the branch into 5 clean, self-contained commits on top of current main and force-pushed — no more corrective commits, and the no-op merge is gone:
- Add chip-agnostic PoE interface and RTL8238B driver
- Add per-machine PoE capability and the KP-9000-9XHPML-X V3.1 machine
- Add poe console command (load/port/global)
- Add PoE web page and /poe.json endpoint
- Add PoE documentation and OEM image-extraction tool

The fixups were cross-cutting (e.g. the arg-passing change touched driver, console and docs together), so a plain fixup would've mixed unrelated changes — I regrouped by area instead. Resulting tree is identical to before; only the history changed.

logicog:
Heads up! The translation-framework that was just merged conflicts with the changes in navigation.js, but that should be easy to fix, I hope.

tobik312:
Thanks for the heads-up! Do you want me to resolve the conflict first and then continue with the changes? Last time there was an issue around this, so I’d rather make sure.

slyalfa:
I have interest in this as I have made a custom switch that use the TPS23880RTQR for POE. And would like to add support for TPS23880RTQR to this code once it is pulled in. Right now I have my esp32 loading FW and controlling TPS23880RTQR.
So keep in mind if possible how to support different POE chips.  At first glance it seems the way the TPS23880RTQR works is very similar to RTL8238B

html/navigation.js
```diff
@@ -10,3 +10,11 @@ document.getElementById('sidebar').innerHTML =
  + "<li><a href='bandwidth.html'>Bandwidth Limits</a></li>"
  + "<li><a href='system.html'>System Settings</a></li>"
  + "<li><a href='update.html'>Firmware Update</a></li></ul>";
+
+// poe.html ships in every build, but the PoE code (and /poe.json) is compiled in only on PoE
+// machines. Show the PoE menu entry only when the firmware has it - i.e. when /poe.json answers.
+fetch('/poe.json').then(function(r) {
```
tofurky: A possible alternative could be to instead return a boolean flag in the initial /information.json request rather than making this extra /poe.json fetch for all switches (PoE *and* non-PoE).

information.json is already fetched upon login -> /index.html, so it's not an extra HTTP request.

```
html/index.html:  <script src="/main_info.js"></script>
```

vs. navigation.js -> poe.json on every page load:

```
html/bandwidth.html:    <script src="/navigation.js"></script>
html/eee.html:    <script src="/navigation.js"></script>
html/index.html:  <script src="/navigation.js"></script>
html/l2.html:  <script src="/navigation.js"></script>
html/lag.html:    <script src="/navigation.js"></script>
html/mirror.html:    <script src="/navigation.js"></script>
html/ports.html:    <script src="/navigation.js"></script>
html/stat.html:  <script src="/navigation.js"></script>
html/system.html:    <script src="/navigation.js"></script>
html/update.html:    <script src="/navigation.js"></script>
html/vlan.html:    <script src="/navigation.js"></script>
```

Even if not combining it with the existing /information.json, I think we should only make the request once and store the result with something like sessionStorage.setItem to avoid continuous re-fetches when navigating between pages. Caching would also make it show up "instantly" rather than waiting for an HTTP request before adding the menu item.

One gotcha with sessionStorage is it's per-tab, so if another tab is opened utilizing the same login session cookie the cached data wouldn't be visible there. localStorage could maybe work around that.

tobik312: Yeah, you're right — went with the /information.json route.

send_basic_info() now reports "poe": true|false from the compile-time POE_PRESENT gate, so it costs nothing extra: no new endpoint, no per-page probe. main_info.js caches it in localStorage (cap_poe) on the Overview page and skips it when building the info table.

navigation.js just reads the cached flag and inserts the menu entry synchronously — instant, no fetch on any page load. localStorage rather than sessionStorage for exactly the cross-tab reason you mention; and if the flag isn't cached yet (fresh browser, or a tab opened before Overview ran), it listens for the storage event once and adds the entry when Overview writes it.

The /poe.json fetch in navigation.js is gone

tobik312:
@logicog

Went ahead and resolved it — the branch is rebased on current main and force-pushed. Still no merge commits; the PoE work is now 6 self-contained commits on top of main.

navigation.js: kept the translation framework's menu markup and the data-i18n pass as-is, and hung the PoE entry off it. The entry deliberately has no data-i18n key — "PoE" is the same string in en/ja/zh, so there's nothing to translate and i18n.js stays untouched.

poe.html was missing <script src="/i18n.js">, which I added in the commit that introduces the page: navigation.js calls t() on DOMContentLoaded now, so without it the sidebar labels broke on the PoE page. The page's own strings aren't wired into the framework yet — it loads i18n.js only so the shared sidebar works correctly. Translating the PoE page itself is a clean follow-up once the PR lands; say the word if you'd rather have it in this one.

logicog:
I am on vacation and cannot test and review. Will have time next weekend.

logicog:
Could you resolve the merge conflict issues, I would like to test the code?

vDorst:
I also want to pick this up again.

Can you rebase this after #315 is merged?

coelner:
just for reference:
Regular 8+1 2.5G Web Managed PoE Switch, however the pcb is marked as `2M-PCB51-Ml-V1.1` and the rtl8238 is used too.

the device information from the web interface, however the label just says: `KP-9000-9XHPML-X`
```
Device Model 	KP-9000-9XHPML-X-AC
Firmware Version 	V100.9.25
Firmware Date 	Jan 20 2026
Hardware Version 	V1.1
```

Image

```
python extract_rtl8238b_image.py /home/user/RTL-switch/dump.bin -o /home/userRTL-switch/
Found 2 PSE image(s) in /home/user/RTL-switch/dump.bin:
  @0x0f8428   21164 B (0x52ac)
  @0x0fd6d4    8028 B (0x1f5c)   <- the KP-9000 uploads this

wrote /home/user/RTL-switch/pse_image.bin (8028 B) - ready to build
```

the boot log from the unmodified firmware:
```
==========Loader start V0.2===========
Press any key to start the normal procedure.
To run SPI flash viewer, press [v]
To enforce the download of the runtime kernel, press [ESC] .....
  cmd -1
    Check Runtime Image.....
    Chksum Correct!
    RunTime Kernel Starting....  
===============  Try to detect the PSE   ===============
===============  PSE IC Existed Result:39  ===============
===============  PSE IC Type: RTL8239  ===============
===============  PSE IC ChipSet Result:C  ===============
Ver8372N=2
Ver8373N=2
NSQ=1 RXIDLE=0 SYNC_OK  Rx reset!
LINE 1472, RL6818C_pwr_on_patch_phy_v007 , patch 0xf finished!
LINE 1472, RL6818C_pwr_on_patch_phy_v007 , patch 0xf0 finished!
===============  Init PSE Starts  ===============
PSE Step0: Reset the whole PSE Chip.
rtk_pse_wholechip_reset1
rtk_pse_wholechip_reset2
PSE Step1: Disable all the PSE ports.
address: 0x20 pse_init regVal=fff0000
address: 0x21 pse_init regVal=fff0000
address: 0x20 pse_init regVal=fff0000
address: 0x20 pse_init regVal=fff0000
=================================
PSE Step2: Begin to load PSE image.
load ok...
===============  Init PSE Ends  ===============
```

`401d91bd543feb573ef9ecb99644c456c6f382cecbfcc346ffe89ea4f0d5676b  pse_image.bin`

Not tested yet.

neuffer:
I have the v1.2 version of that board:
Device Model 	KP-9000-9XHPML-X-AC
Firmware Version 	V100.9.25
Hardware Version 	V1.2
signal-2026-09-17-09-01-45-321_003
signal-2026-09-17-09-01-45-321_002
signal-2026-09-17-09-01-45-321

neuffer:
Better photos of my KP-9000-9XHPML-X AC V1.2 unit (phone originals this time; the earlier ones were recompressed):

  - **PSE marking: `RTL8238C`** (second line `P6H72E1 GP30`), not the RTL8238B this PR targets. That fits the `PSE IC ChipSet Result:C` line in coelner's stock V1.1 boot log above, so the V1.1 is probably a C as well.
  - PoE daughterboard `StandardPOE8-SKY--V1.3` with a single RTL8238C, 8 port fuses, and a `P24C256` EEPROM (U2). Per the RTL8238B-VB datasheet, one chip covers all 8 ports (`0x20`/`0x21` being its port 0-3 and port 4-7 halves), so one chip fits.
  - SPI flash (U8): Winbond `25Q16JVSIQ` (W25Q16JV, 2 MiB).
  - Main PCB `2M-PCB51-ML-V1.2`. 
  - UART on header T8: TX GND RX VCC. 
  - MDIO on T9: MDC GND MDIO.
  - Built-in mains PSU, 52 V / 2.6 A.

 Questions: 
1. does the RTL8238B register map and the extracted `pse_image.bin` apply to the C as-is? I couldn't find a datasheet for the C.
2. Also, the board has a 256 Kb EEPROM next to the PSE, which the B datasheet describes as a firmware autoload option (BOOTSEL strap); does anyone know whether these boards use it, given the OEM firmware uploads the image anyway? Happy to test on this unit.

PXL_20260922_173559803
PXL_20260922_173802585
PXL_20260922_174021989
PXL_20260922_174125417

PR #271, opened 2026-06-25
Add reset switch to SWTG018AS-A V2.0

th0m4sek:
Same board used by Horaco HC-SWTGW218AS with populated reset switch.
Tested and working.


vDorst:
thanks

PR #273, opened 2026-06-27
fix(machine): SWTG024AS V2.0 fix 5th RJ45 port

Lynn-Becky:
Here is some background on the original issue with the 5th RJ45 port:
In prior builds, the 5th RJ45 port could only handle TX traffic, but failed to receive RX traffic. In short, you could see the tx_good count growing in the management panel, but the rx_good count always stayed at 0.

Following further research and comparison with the dumped stock firmware, I fix the 5th RJ45 port issues via a patch. (not for other machine)

Additionally, I noticed a few different between the project and the stock firmware's behavior. Since they aren't causing any bugs at the moment, I have kept them as-is for now and excluded them from this PR.

### DIFF
--- a/rtl837x_init.c
+++ b/rtl837x_init.c
@@ -76,6 +76,16 @@ void static sds_init(void)
 				sds_read(0, 0, 0);
 				pval = SFR_DATA_U16;
 				sds_write_v(0, 0, 0, pval | 0x200);
+#if defined MACHINE_SWTG024AS_V2_0
+				// OEM firmware also sets the companion SDS0 polarity bits.
+				sds_read(0, 0, 0);
+				pval = SFR_DATA_U16;
+				sds_write_v(0, 0, 0, pval | 0x100);
+
+				sds_read(0, 6, 2);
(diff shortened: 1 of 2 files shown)


vDorst:
Thanks

rtl837x_init.c
```diff
@@ -76,6 +76,16 @@ void static sds_init(void)
 				sds_read(0, 0, 0);
 				pval = SFR_DATA_U16;
 				sds_write_v(0, 0, 0, pval | 0x200);
+#if defined MACHINE_SWTG024AS_V2_0
```
feelfree69, line 79: Can't this be moved to machine_custom_init() in machine.c?

So far all machine-dependent configurations are in machine.c, it would be nice if this can be kept.

Lynn-Becky: > Can't this be moved to machine_custom_init() in machine.c?
> 
> So far all machine-dependent configurations are in machine.c, it would be nice if this can be kept.

done.
#274

PR #278, opened 2026-07-04
update document: SKS3200M-4GPY2XF flash size

eraiza0816:
I have the actual device for this switch and have successfully installed RTLPlayground on it. 

I wanted to make a small contribution, so I verified the flash size of this product and updated it accordingly.   
The flash chip used in this device is **FM25Q16A**.

I am available to conduct further investigations or tests on this hardware if needed.  
Please let me know if there is anything else you would like me to check!


vDorst:
Thanks

PR #279, opened 2026-07-04
Support i18n & translate japanese

eraiza0816:
### Summary
This PR implements i18n and adds Japanese language support. 
Please note that common computer/technical terms have been left untranslated as per standard practice.

*Note: This enhancement will become obsolete once the Fast UI from PR #220 is merged. (By the way, I really love the Fast UI!)*

### Background
As a native Japanese speaker, I wanted to contribute this localization. These switches listed in the project are highly popular and widely circulated in Japan, backed by a very active community. (My English isn't perfect, so navigating this has been quite a journey!)

### Concerns & Footprint
This localization adds about 5.2 KiB to the file size. Going forward, supporting additional languages will likely incur a similar increase per language.

Looking at recent commits from other contributors, I noticed that highly efficient techniques are being utilized to save even just a few KB. Given this impressive commitment to optimization, I am aware that adding a substantial binary size increase for UI updates might not be ideal for the project's goals. 

Personally, I feel this specific size is negligible, but if multi-language expansion is not on the project's current roadmap, or if the size increase contradicts the optimization policy, please feel free to close this PR.

image

### DIFF
--- a/doc/support-multi-language.md
+++ b/doc/support-multi-language.md
@@ -1,0 +1,135 @@
+# Supporting Multiple Languages in the Web UI
+
+The firmware uses a client-side i18n approach
+all translations are stored in a single JavaScript dictionary embedded in the firmware.
+No server-side changes are needed.
+
+## Architecture
+
+All translation logic lives in `html/i18n.js`. The file contains:
+
+- A `LANG` object with one sub-object per language (`en`, `ja`, ...)
+- Language auto-detection (browser language → `localStorage` override)
+- `t(key)` — look up a translated string
+- `setLang(lang)` — switch language and update the page
+- `applyTranslation(el)` — apply translation to one DOM element
+
+Translation keys are **flat strings** (no nesting). The English keys in `LANG.en` also serve as the fallback when a key is missing in another language.
+
+## How to Add a New Language
+
+### 1. Add a dictionary entry in `html/i18n.js`
+
+Append a new sub-object to the `LANG` object. Every key from `LANG.en` must be present:
+
+```js
+var LANG = {
+  en: {
+    nav_overview: 'Overview',
+    nav_port_config: 'Port Configuration',
+    // ... all keys for English
+  },
+  ja: {
+    nav_overview: '概要',
+    nav_port_config: 'ポート設定',
+    // ... all keys for Japanese
+  },
+  LANGCODE: {            // ← add your language here
+    nav_overview: '...',
+    nav_port_config: '...',
+    // ... translate every key
+  },
+};
+```
+
+### 2. Add the language to the navigation sidebar
+
+In `html/navigation.js`, add an `<option>` to the language selector:
+
+```js
++ "<option value='en'>English</option><option value='ja'>日本語</option>"
+```
+
+Replace with:
+
+```js
++ "<option value='en'>English</option><option value='ja'>日本語</option><option value='LANGCODE'>Native Name</option>"
+```
+
+### 3. Verify auto-detection
+
+The language detection code in `i18n.js` reads `navigator.language` and normalises it to the first two characters:
+
+```js
+var browser = (navigator.language || navigator.userLanguage || 'en').substring(0, 2);
+return LANG[browser] ? browser : 'en';
+```
+
+If the two-letter code matches a key in `LANG`, it will be auto-selected. No changes needed here.
+
+## Two Translation Mechanisms
+
+### (A) `data-i18n` attribute (declarative — for HTML)
+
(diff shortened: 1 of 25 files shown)


logicog:
> Looking at recent commits from other contributors, I noticed that highly efficient techniques are being utilized to save even just a few KB. Given this impressive commitment to optimization, I am aware that adding a substantial binary size increase for UI updates might not be ideal for the project's goals.
> 
> Personally, I feel this specific size is negligible, but if multi-language expansion is not on the project's current roadmap, or if the size increase contradicts the optimization policy, please feel free to close this PR.

The optimization is mostly for code, not so much for the web-interface. 5.2 is really not much to support another language. I just had fun with an API that was entirely documented in Chinese, and my appreciation for translations has just grown enormously. So I believe this is really great. In the future, when we have 50 :grin: languages to choose from, we may thing of a way to disable some of them.

I just wanted to try out the language switching, but I noticed that you have removed the entry again in the navigation menu in the 3rd commit. Now, I cannot find a way to change the language from the Web.

Then, I thought, I could change it from the CLI, but did not find any command. So I am wondering how the language selection is made persistent across reboots. Is this already implemented?

In any case, bit kudos for taking the translation on. And I like the approach a lot!

eraiza0816:
Thank you for the warm feedback and the kind words! I'm really glad to hear that you like the approach. 

Your story about the Chinese API made me smile—even for us Japanese, who share a similar font/writing system with Chinese, there's definitely a limit to how much of the subtle nuances we can actually grasp. So I completely understand the struggle!

To answer your questions about the language selection:


> I just wanted to try out the language switching, but I noticed that you have removed the entry again in the navigation menu in the 3rd commit. Now, I cannot find a way to change the language from the Web.

I decided to remove the manual toggle UI because the system is designed to automatically detect and switch languages based on the user's Web browser settings. It currently supports English and Japanese, and defaults to English for any other browser language.

How language selection and persistence work  
Please note that the Switch itself does not store any language settings.   
The language is managed entirely on the client side (browser) and is determined using the following fallback order:

1. `localStorage.getItem('rtl_lang')` — Saved user preference.
2. `navigator.language` — Browser's default language setting.
3. `'en'` — Default fallback.

Since it checks `localStorage` first, the language selection remains persistent across reboots once saved in the browser, without needing any storage on the Switch side.

How to test other languages via DevTools, If you are using an English browser and want to test the Japanese UI (or any other language added in the future) without changing your browser settings, you can use the DevTools Console via either of the following methods:

1. Open the Web interface in your browser.
2. Press  F12 to open DevTools.
3. Click on the Console tab.
4. Use either of the following methods to switch the language:

Method 1: Using the function
* **To switch to Japanese:** Run `setLang('ja')`
* **To switch back to English:** Run `setLang('en')`

Method 2: Modifying localStorage directly and reloading
You can also set the language key directly in `localStorage` and refresh the page:
```javascript
localStorage.setItem('rtl_lang', 'ja');
location.reload();
```

logicog:
Thanks for the explanation. It looks great (although of course I do not understand Japanese). One thing I had to figure out was, that setLang() does not work at the login page because it is a Javascript function only loaded later.

I think it may be good if the language could be configured from the web-interface as you had also foreseen. I know quite a few people who have their browser configured for German, but for technical web-pages prefer English, for example. For most web-pages I know there is a selection to change the language, so I think many people will be familiar with that concept, and may not understand that it is the browser's language that is being picket up. They will look for a change-language button and not find it. 
The default language could be the browser-language, though. The language could be made persistent with a CLI command setting the language. If you do not want to program this now, it could be added separately.

eraiza0816:
You make a great point! I agree that having a language selection option directly on the UI is much more user-friendly (as is common with consumer-grade switches). 

To give you a bit of background, the reason I originally removed the UI toggle was due to a few layout and performance constraints in my implementation. The language menu ended up overlapping with the RJ45 and SFP port visuals, and it also caused a slight dip in the UI rendering speed. Because of this, I thought it would be better to keep it clean for now.

As you suggested, I would love to bring the UI toggle back in a cleaner way. I think it makes the most sense to implement the UI language switcher after we resolve the rendering speed/layout issues—perhaps in a separate PR or after looking at upcoming UI updates. 

Would it be alright if we proceed with the current browser-based implementation for this PR, and handle the UI toggle as a follow-up task?

logicog:
> Would it be alright if we proceed with the current browser-based implementation for this PR, and handle the UI toggle as a follow-up task?

Fine with me. Regarding the UI-based toggle, it could also be handled together with the firmware and other configuration options in a sub-menu. This will not slow things down.

PR #280, opened 2026-07-06
Add SFP quirk for devices that misreport DDM capability

tofurky:
On a QSFPTEK QT-SFP+-T (RTL8261C) 10GBase-T module, the diag type field
(92) comes back as 0x00, but there's actually some statistics available
like temperature. Other metrics may be hard-coded values.

Add a basic struct that allows matching on vendor and/or model, using a
bitfield to allow multiple quirks for a given SFP module. Only
SFP_QUIRK_DDM is implemented.

For modules matching SFP_QUIRK_DDM, attempt an I2C read of the MSB of
module voltage during probe if DDM is "unsupported" - if it's not 0xff,
override the reported options so we can pull the diagnostic data.

To allow simpler comparisons, convert the ASCII fields (vendor,
model, serial) from space-padded to standard NULL-terminated strings.

strcmp() is moved from httpd.c to rtlplayground.c alongside other string
functions and shared between them.

The __reentrant keyword is used for the new functions to avoid using up
additional OSEG space. This allocates the variables on the stack, which
is OK for this particular code path.

The JSON assembly in send_status() is slightly modified to treat the
sfp_module_* data as standard NULL-terminated strings, and a repeated
subtraction was moved into a uint8_t to declutter the code.

### DIFF
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -50,6 +50,7 @@ __xdata char port_names[9][PORT_NAME_SIZE];
 extern __xdata uint16_t management_vlan;
 extern __xdata uint8_t sfp_speed[2];
 extern __xdata uint8_t sfp_pins_last;
+extern __xdata uint8_t sfp_options[2];
 __xdata uint8_t gpio_last_value[8] = { 0 };
 
 // Temporatly for str to hex convertion value.
@@ -741,7 +742,7 @@ void parse_mtu(void)
 void sfp_print_measurements(uint8_t sfp)
 {
 	print_string("Options: "); print_byte(sfp_read_reg(sfp, 92)); write_char('\n');
-	if (!(sfp_read_reg(sfp, 92) & 0x40))
+	if (!(sfp_options[sfp] & 0x40))
 		return;
 	print_string("Temp: "); print_byte(sfp_read_reg(sfp, 224)); print_byte(sfp_read_reg(sfp, 225)); write_char('\n');
 	print_string("Vcc: "); print_byte(sfp_read_reg(sfp, 226)); print_byte(sfp_read_reg(sfp, 227)); write_char('\n');
--- a/httpd/httpd.c
+++ b/httpd/httpd.c
@@ -101,21 +101,6 @@ uint8_t find_entry(__xdata uint8_t *e)
 }
 
 
-char strcmp(__xdata uint8_t *c, __code uint8_t * __xdata d)
-{
-	uint8_t i = 0;
-
-	while (d[i] && (d[i] == c[i]))
-		i++;
-
-	if (c[i] < d[i])
-		return -1;
-	else if (c[i] > d[i])
-		return 1;
-	return 0;
-}
-
-
 bool is_word(__xdata uint8_t *xdata_str_p, __code uint8_t * __xdata code_str_p)
 {
 	uint8_t u, c;
--- a/httpd/page_impl.c
+++ b/httpd/page_impl.c
@@ -660,52 +660,53 @@ void send_status(void)
 		slen += strtox(outbuf + slen, "\"");
 
 		if (machine.is_sfp[i]) {
+			uint8_t sfp = machine.is_sfp[i] - 1;
 			slen += strtox(outbuf + slen, ",\"isSFP\":1,\"enabled\":");
-			if (!(sfp_pins_last & (0x1 << ((machine.is_sfp[i] - 1) << 2)))) {
+			if (!(sfp_pins_last & (0x1 << (sfp << 2)))) {
 				bool_to_html(1);
 				slen += strtox(outbuf + slen,",\"sfp_options\":\"0x");
-				byte_to_html(sfp_options[machine.is_sfp[i]-1]);
-				if (sfp_options[machine.is_sfp[i]-1] & 0x40) {
+				byte_to_html(sfp_options[sfp]);
+				if (sfp_options[sfp] & 0x40) {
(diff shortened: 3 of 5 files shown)


tofurky:
Not sure how common this "quirk" is - here's example output. SFP1 is DAC, SFP2 is the quirky 10GBase-T module:
```
> sfp

Slot 1 - Rate: 67  Encoding: 00
OEM             SFP-H10GB-CU1M  R   
Options: 0000
Diag Type: 00

Slot 2 - Rate: 67  Encoding: 00
QSFPTEK         QT-SFP+-T       03  
Options: 0000
Diag Type: 00
Temp: 3940
Vcc: 83bd
TX Bias: 0bb8
TX Power: 1390
RX Power: 18a0
Laser: 0000
State: 00
```

It's a bit of a hack and based on a sample size of 1.

Example of another module that *does* advertise DDM:
```
> sfp

Slot 1 - Rate: 67  Encoding: 00
OEM             SFP-H10GB-CU1M  R   
Options: 0000
Diag Type: 00

Slot 2 - Rate: 64  Encoding: 03
AZORES          WAS-110         V1.0
Options: 001a
Diag Type: 68
Temp: 2c00
Vcc: 7dd0
TX Bias: 13b0
TX Power: 8716
RX Power: 006f
Laser: 7da0
State: 00
```

tofurky:
Tried an older SFP module I had as well; dummy read check seems to work there as well:
```
Slot 2 - Rate: 0d  Encoding: 01
FS              SFP-GB-GE-T     A   
Options: 0012
Diag Type: 08
```

vDorst:
I think this is not the right way to solve this. I personally would like to see how the [linux kernel](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/drivers/net/phy/sfp.c?h=v7.1.3#n502) solves this kind of issues with SFP modules. Al thought I didn't find a quirk for hw-monitoring.

Basics are, use the standard, but look in the quirk-table for a match `vendor_name` and `vendor_partnumber` and apply the quirk when needed.

tofurky:
> I think this is not the right way to solve this. I personally would like to see how the [linux kernel](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/drivers/net/phy/sfp.c?h=v7.1.3#n502) solves this kind of issues with SFP modules. Al thought I didn't find a quirk for hw-monitoring.
> 
> Basics are, use the standard, but look in the quirk-table for a match `vendor_name` and `vendor_partnumber` and apply the quirk when needed.

So, maintain a database of vendor / model strings, then strcmp all of them (just one so far) upon SFP initialization? This would have to happen after sfp_get_info() as that's where those strings are read back.

Are you suggesting a full quirks framework like the kernel or just enough to override the sfp_options value as is done here already?

One problem with solely relying on vendor/model is QSFPTEK is apparently using the same model number for SFP+ with different PHYs; it's not clear if they all behave this way, or if we unconditionally tried to pull metrics from all of them we'd start getting garbage 0xff back for some.

logicog:
I don't think we need a full quirks implementation, as we do not need to treat the PHYs of the SFP modules in any way differently in RTLPlayground (so far at least). Most of the quirks in the kernel are required because Linux uses phylink to configure these modules as PHYs and therefore needs to be really specific about how to configure them. We just need to read the EEPROM correctly (and, yes, there is quite a bit of Linux kernel code to fix that, too).

The danger with your original approach is that we also catch modules where the fix does not work and we may get unwanted consequences, for example incorrect temperature readings.

If QSFPTEK has other modules that may behave similarly, then it may be possible to handle the quirk based on the vendor-id/vendor string.

tofurky:
So, try a dummy read with this specific vendor/model, or just unconditionally enable DDM for this vendor/model? I think it would be safer to still also check that it's returning non-0xff data rather than just basing it on vendor/model.

I thought this could be something generally useful - hence not restricting it to specific modules - but I have no idea how common it actually is.

logicog:
I do not think this is very common. There are a lot of quirky SFP modules out there, but they all have their own quirks. I have not encountered twice the same issue, so far., apart from providing the wrong nominal bitrate for operation, which is fairly common (and we are actually allowing a wide range).

tofurky:
Pushed 1136e4a48a62dde880d57ca77f9a51a04b19a940 which limits this to the particular QSFPTEK SFP module. There is not quite a "framework" but it's written in such a way as to allow more SFPs/quirks to be added.

If it looks OK I can rebase/squash it but left both commits for the time being to allow seeing the changes between the first/second version.

Edit: pushed again to remove remnant of old code that slipped through.

logicog:
It looks great. Can you rebase/squash the commits. I was not aware about the __reentrant keyword. Did you try it out in other places?

tofurky:
> It looks great. Can you rebase/squash the commits. I was not aware about the __reentrant keyword. Did you try it out in other places?

Thanks for reviewing - squashed the 2 with an updated commit message and updated PR title/description to match.

I did not generally try `__reentrant`; stumbled across it when hitting OSEG limits. Apparently, `--stack-auto` does the same thing for *all* functions but could lead to stack overflows if not careful. Tests showed that OSEG space is still taken even for `inline` functions. I think `__reentrant` is kind of misleadingly named. It could be useful for allowing code to be more neatly structured and reused without worrying about tying up the very limited global OSEG.

Candidates could be found by doing some offline analysis of the intermediate .rel files. For example, `phy_modify()` is using 6 bytes which disappear from OSEG after adding __reentrant. Would it run into issues with too much stack? Maybe not, but would need further testing.

PR #281, opened 2026-07-07
Add support for Horaco HC-SWTGW218AS (PCB-V2.0.1_19649)

inno3759:
## Summary

  Adds a machine definition for the **Horaco HC-SWTGW218AS**, a 9-port 2.5G *web managed* switch (8×2.5G RJ45 + 1×10G SFP+).

  The PCB silkscreen reads `HC-SWTGW218AS_PCB-V2.0.1_19649` and is the **same board** as the `SWTG018AS-A V2.0` (the unmanaged
  Ampcom-branded variant) — same `..._19649` layout, RTL8273N + RTL8224N, identical SFP wiring and LED routing. Horaco simply
  ships this hardware with managed firmware. The new `MACHINE_HC_SWTGW218AS_PCB_V2` config is therefore a copy of
  `MACHINE_SWTG018AS_A_V_2_0`.

  > Note: this is **not** the same board as the existing `MACHINE_SWTGW218AS`, which targets the SWTG118AS PCB (Mokerlink
  ZX-SWTGW218AS / XikeStor SKS3200-8E1X) with different SFP/reset GPIOs.

  ## Why a separate definition (problems with `MACHINE_SWTGW218AS`)

  Since this switch is also branded "SWTGW218AS", the obvious first attempt was to build the existing `MACHINE_SWTGW218AS`.
  **That does not work on this PCB** — it targets the unrelated SWTG118AS board, and the differences are not cosmetic:

  - **LEDs are offset by one.** Port 1 lights port 2's LED, port 2 lights port 3's, etc. Root cause: `MACHINE_SWTGW218AS` does
  not set `led_mux_custom`, so the firmware assumes the chip's default LED pad routing; this PCB wires the pads differently and
  needs the explicit `led_mux_custom = 1` + `led_mux[]` remap (inherited from `SWTG018AS_A_V2_0`).
  - **The SFP port does not come up** — different GPIOs:

    | Setting          | `MACHINE_SWTGW218AS` (SWTG118AS) | This board (`..._19649`) |
    |------------------|----------------------------------|--------------------------|
    | SFP `pin_detect` | `GPIO30_ACL_BIT3_EN`             | `GPIO38`                 |
    | SFP `pin_los`    | `GPIO37`                         | `GPIO_NA` (not wired)    |
    | `reset_pin`      | `GPIO54_ACL_BIT2_EN`             | `GPIO48_I2C_SCL1`        |

  Switching to the `SWTG018AS-A V2.0` config (which this definition copies) fixed both — correct per-port LEDs and a working
  SFP+ port.

  ## Changes
  - `machine.h`: add `MACHINE_HC_SWTGW218AS_PCB_V2` define (commented, like the others)
  - `machine.c`: add machine config (copy of `MACHINE_SWTG018AS_A_V_2_0`) with a clarifying comment
  - `doc/devices/HC-SWTGW218AS_PCB-V2.md`: new device documentation with label + PCB photos and the notes above
  - `doc/supported_devices.md`: point the existing Horaco HC-SWTGW218AS row at the new doc

  ## Testing
  Tested on real hardware:
  - All 8 2.5GBASE-T RJ45 ports link at 10/100/1000/2500 Mbps
  - SFP+ port works (1G/2.5G/10G)
  - LEDs match OEM firmware
  - VLANs verified


th0m4sek:
It's not elegant to have two separate docs about same PCB in my opinion.
Supported boards has proper definition for compilation. 
So in my opinion its better do extend 018as PCB page and add info about this board.

Issue #277, opened 2026-06-29
horaco HC-SWTGW215AS. Wrong LED configuration with ZX-SWTGW215AS firmware\Lack of ethernet Port5 for PCB_SWTG024AS_A_2_0_1 firmware


bademux:
Hi,

It looks like there is minor misconfiguration for [horaco HC-SWTGW215AS](https://pl.aliexpress.com/item/1005010660780744.html) (stock firmware V200.1.10) with  Lianguo ZX-SWTGW215AS RTLPlayground firmware (MACHINE=ZX_SWTGW215AS).

Winbond flash chip W25Q16JV

<details>
  <summary>OK\NOK for MACHINE=ZX-SWTGW215AS </summary>
LEDs are not aligned (ports 1-5 left-to-right, 2.5G):
Port1 - green  LED Port2
Port2 - yellow LED Port2
Port3 - green  LED Port3
Port4 - yellow LED Port3
Port5 - green  LED Port1

for 1G - no LED on any port

LED pcb silkscreen
Port1 - LED2
Port2 - LED3
Port3 - LED4
Port4 - LED5
Port5 - LED1
Cage - LED6

WEB UI is not accessible on Port5, although its activity is displayed on WebUI
Cage doesn't work
</details>

board PCB-SWTG024AS-A-V2.0.1_19650
<img width="2120" height="1746" alt="Image" src="https://github.com/user-attachments/assets/b9020886-3544-49be-9fd9-c158f188da67" />

firmware compiled with [the script](https://gist.github.com/bademux/0e8c0702879aae997aae44a5fb86dbfc)

UPD:
tested with `MACHINE=PCB_SWTG024AS_A_2_0_1` firmware - LEDs are OK,
 but board configured 2 cages instead of 5+Cage.
Cage works on MACHINE=PCB_SWTG024AS_A_2_0_1, but 5th ethernet port don't

Lynn-Becky:
use PCB-SWTG024AS-V2.0 to test it. it look like simulate PCB-SWTG024AS-V2.0.only differnt on flash chip 
It looks like this PCB either configures the 8221b differently, or uses a different chip revision, because its  init is different from other 5+1 port switch

bademux:
@Lynn-Becky,  with `MACHINE=SWTG024AS_V2_0` 5th Ethernet is works (LEDs are OK too), but sfp doesn't work (with MACHINE=PCB_SWTG024AS_A_2_0_1 SFP works)

[...]

PR #282, opened 2026-07-08
add SWTG024AS-A-V2.0.1 with 5 RJ45 port 

Lynn-Becky:
#277 
I have no idea about machine name,temporarily use SWTG024AS_A_2_0_1_5_RJ45.


Lynn-Becky:
I creat a simple document with out picture about this device.Need to update it with someone keep this device

feelfree69:
I suggest something like  _5C_1SFP and _4C_2SFP as appendix for the machine definitions. C stands for copper. 
And I would also add the appendix to the existing name, to make it really clear.

Lynn-Becky:
> I suggest something like _5C_1SFP and _4C_2SFP as appendix for the machine definitions. C stands for copper. And I would also add the appendix to the existing name, to make it really clear.

The machine definition has been updated to SWTG024AS_A_2_0_1_5C_1SFP.

logicog:
Looking at the PR, I assume it is OK now (and @feelfree69  is on vacation?). So I will merge.

PR #283, opened 2026-07-08
Translate to Chinese

Lynn-Becky:
Earlier, I created a branch for the Chinese translation, but since it was hard-coded into the project, I didn't open a PR. Now that i18n is available, I have remade the Chinese translation. 

Huge thanks to #279  and author.

image


Lynn-Becky:
In the current version, clicking Detailed Port Statistics does not respond.
The browser console shows:
stat.js:117
Uncaught TypeError: t is not a function
    at xhttp.onreadystatechange (stat.js:117:56)
    
    
`html/stat.js`has a runtime error :
`var t = "<table ...>" + t('stat_counter') + ...`
The local variable `var t ` is hoisted and shadows the global translation function `t()`. As a result, when the right-hand side calls `t('stat_counter')`, it resolves to the local `t`, which is still `undefined`, causing `t is not a function` when opening the “Detailed Port Statistics / All Counters” popup.

Wait to fix

logicog:
@eraiza0816 Can you have a look at the issue?

eraiza0816:
OK. I will check.

eraiza0816:
I have submitted PR #284 to resolve the issue.

@Lynn-Becky, thank you very much for bringing this to my attention.

Could you please review these changes to see if they address your concerns?

Lynn-Becky:
> I have submitted PR #284 to resolve the issue.
> 
> @Lynn-Becky, thank you very much for bringing this to my attention.
> 
> Could you please review these changes to see if they address your concerns?

Thank you very much for your prompt response and for creating #284 to address the issue. 
I think the problem has been successfully resolved with your commit. I truly appreciate your swift support.

logicog:
Thanks for providing this translation! Great work and thanks for sharing.

PR #284, opened 2026-07-10
fix: rename local variable to avoid shadowing global translation

eraiza0816:
Bugs discovered in #283

Commit e75a4c2 introduced i18n support, adding a global t() function for translations. In html/stat.js, the getCounters function already had a local variable var t used to build the table HTML string.
When the translation calls t('stat_counter') and t('stat_value') , JavaScript hoisting caused the local var t declaration to be hoisted to the top of the function scope. 

At the time t('stat_counter') is evaluated, t resolves to the hoisted local variable (still undefined), not the global function — producing Uncaught TypeError: t is not a function.
The fix renames the local variable from t to tableHtml, eliminating the name collision while keeping the global t() function accessible for translations.

### DIFF
--- a/html/stat.js
+++ b/html/stat.js
@@ -114,7 +114,7 @@ function getCounters(port) {
       const s = JSON.parse(xhttp.responseText);
       console.log("Counters: ", JSON.stringify(s));
       const ptext = document.getElementById('popup_text');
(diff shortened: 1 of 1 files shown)


logicog:
Thanks so much for finding and fixing!

PR #285, opened 2026-07-13
feat: add Dockerfile for development environment

eraiza0816:
I have added a `Dockerfile` for contributors whose development environments are not based on Debian. 

Using Docker guarantees a consistent and reproducible environment for both development and builds across different machines. In particular, it makes it much easier to set up `sdcc` and ensure everyone is using the exact same version without manual installation hassle.

### Documentation
Since this setup is optional and not needed for active developers who already have their environment configured, I have placed the Docker instructions inside a collapsible section (`<details>`) in the `README`.

- Debian 13 (trixie) based with sdcc 4.5.0 from apt
- Includes gcc, make, xxd, python3, libjson-c-dev
- Add .dockerignore to exclude build artifacts


logicog:
Tested and works perfectly. Really cool feature! Could you maybe separately add instructions to e.g. configure the necessary Windows/Linux prerequisites (meaning what you need to set up for docker, can be pointers), similarly as is done for normal compilation on Linux?

PR #286, opened 2026-07-14
add language selection to system settings

eraiza0816:
Follow-up to #279 and #283.

Add a language selector to the System Settings page, allowing users to switch between English, Japanese, and Chinese.

### Language selection feature

* **`html/i18n.js`**: Added the `sys_language` key for all three supported languages.
* **`html/system.html`**: Added a language selector dropdown (English / 日本語 / 中文) below the Gateway field in the System tab.
* **`html/system.js`**: Added the `changeLang()` function and implemented initialization of the selector value from `rtlLang` on page load.

The selected language is persisted in `localStorage`.

image

### Bug fixes in `httpd_sim.c` (test server)

I identified and addressed the following issues during manual testing:

1. **Cookie parsing**: `scan_header()` previously used a hardcoded offset (`p+17`) to locate the session ID. This caused failures when the browser sent additional cookies (e.g., `oc_locale=ja; session=...`). This has been fixed by using `strstr()` to search for `session=`.
2. **Set-Cookie response**: The login success response was missing `Path=/` and the terminating `\r\n\r\n`, which caused browsers to reject the cookie.
3. **SO_REUSEADDR**: Added this socket option to prevent "Address already in use" errors during restarts.


logicog:
Works great! Thanks for sharing. What I believe is still missing is a way to make the language selection permanent. For this it would be necessary to define a command for the CLI. Something like `lang cn` or similar. With such a command it will become possible to add it to the configuration file of the switch. But maybe this is for a different PR?

eraiza0816:
I’ve decided to include the language setting save functionality in a separate PR.
Sorry for the sporadic PRs...

tofurky:
Ran into this after pulling in this change via a rebase:

```
gcc httpd_sim.c -Wall -o output/httpd_sim -I/usr/include/json-c -ljson-c
httpd_sim.c: In function ‘scan_header’:
httpd_sim.c:570:35: warning: initialization discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
  570 |                         char *s = strstr(session, "session=");
      |                                   ^~~~~~
```

`const char *s =` makes it go away.

eraiza0816:
I have read it. My apologies.  
I haven't tested it myself yet, but I believe that making this modification will prevent the warning.

```c
char *s = strstr(session, "session=");
```

```c
const char *s = strstr(session, "session=");
```

Please wait while I release the fix.

PR #289, opened 2026-07-19
fix: httpd-sim-const-warning

eraiza0816:
Fix #286 Warn.


logicog:
Thanks for fixing this so quickly!

PR #291, opened 2026-07-20
httpd: fix browser login (cookie parsed by name, Connection: close, self-contained login page)

DrDoof:
## Problem

Logging in from a real browser intermittently (or, with certain cookies, always) bounces back to the login page with "Wrong password!" even though the password is correct — while the same credentials always work via curl. Reproduced and debugged on a SWTGW218AS.

## Root cause

**Session cookie parsed from a fixed offset.** `scan_header()` read the session id at `Cookie:` header offset +17, assuming `session=` is the first and only cookie. Browsers keep stale cookies for a long time — e.g. an `admin` cookie left over from the switch's **vendor firmware** — so the header can arrive as `Cookie: admin=..; session=..`, the fixed offset then points into the `admin` value, authentication silently fails and every authenticated page redirects back to login (with a misleading "Wrong password!", since the redirect chain preserves the login referrer). curl sends only `session=`, which is why command-line tests passed while a real browser (with that stale cookie) failed.

## Changes

- `scan_header()`: scan the `Cookie:` header for the actual `session=` key instead of using a fixed offset (matched as `"session"` — `is_word()` requires a separator after the pattern and `=` is on its list).
- `is_word_x()`: accept `;` as a terminating separator so the session value also matches when it is not the last cookie in the header.
- All HTTP responses now include `Connection: close`. The httpd is built with `UIP_CONF_MAX_CONNECTIONS = 1` and never advertised that a connection is not persistent, so a browser could pool and reuse a connection the server had already closed. This is a general correctness fix independent of the cookie bug above.
- `login.html`: added `autocomplete="current-password"` so password managers recognise the field (kept the existing i18n/`style.css` structure unchanged — no regression to multi-language support).
- The file-serving CSP is now a complete first-party policy (`default-src 'self'; …`) instead of a lone `style-src`, which stopped privacy-shield browsers (Brave, NoScript) from injecting their own noisy report-only probes.

## Verification

On hardware (SWTGW218AS), end-to-end in a real browser **with the stale vendor `admin` cookie present**: login → `index.html`, all pages and JSON endpoints work. curl behaviour unchanged. `make MACHINE=SWTGW218AS` builds clean on this branch.

### DIFF
--- a/html/login.html
+++ b/html/login.html
@@ -19,7 +19,7 @@
   <h1 data-i18n="login_heading"> RTL Switch Login</h1>
   <form method="post" action="login">
   <div class="txt_field">
-    <input name="pwd" type="password" onclick="removeNote()" required />
+    <input name="pwd" type="password" autocomplete="current-password" onclick="removeNote()" required />
     <span></span>
     <label data-i18n="login_password">Password</label>
   </div>
--- a/httpd/httpd.c
+++ b/httpd/httpd.c
@@ -204,28 +204,28 @@ uint8_t parse_short(__xdata uint8_t *p)
 
 void send_not_found(void)
 {
-	slen = strtox(outbuf, "HTTP/1.1 404 Not found\r\nContent-Type: text/html\r\n\r\n" \
+	slen = strtox(outbuf, "HTTP/1.1 404 Not found\r\nConnection: close\r\nContent-Type: text/html\r\n\r\n" \
 			      "<!DOCTYPE HTML PUBLIC>\n<title>404 Not Found</title>\n<h1>Not Found</h1>\n");
 }
 
 
 void send_bad_request(void)
 {
-	slen = strtox(outbuf, "HTTP/1.1 400 Bad Request\r\nContent-Type: text/html\r\n\r\n" \
+	slen = strtox(outbuf, "HTTP/1.1 400 Bad Request\r\nConnection: close\r\nContent-Type: text/html\r\n\r\n" \
 			      "<!DOCTYPE HTML PUBLIC>\n<title>400 Bad Request</title>\n<h1>Bad Request</h1>\n");
 }
 
 
 void send_to_login(void)
 {
-	slen = strtox(outbuf, "HTTP/1.1 302 Found\r\n" \
+	slen = strtox(outbuf, "HTTP/1.1 302 Found\r\nConnection: close\r\n" \
 			      "Location: login.html\r\n\r\n");
 }
 
 
 void send_unauthorized(void)
 {
-	slen = strtox(outbuf, "HTTP/1.1 401 Unauthorized\r\n\r\n");
+	slen = strtox(outbuf, "HTTP/1.1 401 Unauthorized\r\nConnection: close\r\n\r\n");
 }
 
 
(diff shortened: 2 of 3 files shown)


html/login.html
```diff
@@ -1,29 +1,49 @@
...
-  <link rel="stylesheet" href="style.css">
-  <script src="/i18n.js"></script>
+  <title>RTL Switch Login</title>
+  <!-- SELF-CONTAINED on purpose: the uIP httpd serves a single TCP connection
+       at a time (UIP_CONF_MAX_CONNECTIONS=1). If this page pulls style.css,
+       i18n.js and a favicon, a browser opens parallel connections that collide
+       on the single slot - and the login POST (which a browser never retries)
+       intermittently loses that race and fails. With everything inlined the
+       page needs exactly ONE connection and the POST is the only next request. -->
+  <link rel="icon" href="data:," />
```
tofurky: I think the login POST won't be happening at the same time as the CSS/JS loading; POST only happens after the user clicks Login and the DOM should be fully rendered (all static resources loaded) at that point. If login fails, a redirect back to /login.html is done, so it reloads, but the login request has already been processed at that point.

This is also reverting some recent multi-language support changes.

The cookie parser not handling multiple cookies (from prior vendor firmware) seems like it could break login, though, if old/new IPs are the same. Was that the change that actually fixed login for your switch?

DrDoof: You're right on both points, thanks — pushed an update:

- Reverted `login.html` to keep `i18n.js`/`style.css` as-is (just added `autocomplete="current-password"`). The parallel-connection-collision theory for the login POST specifically didn't hold up once I thought it through the way you described — the POST only fires after the page is fully loaded and idle. `Connection: close` stays as a general fix (the server never advertises it's single-connection/non-persistent, so a browser can still pool and reuse an already-closed connection for some later request), but it's no longer tied to the login flow story.
- Yes — the cookie-parsing fix (`scan_header()` reading the session id from a fixed offset instead of the actual `session=` key) is what actually fixed login on my hardware. I had a leftover `admin` cookie from this switch's original vendor firmware still in the browser, and that's what broke the fixed-offset parse. Isolated and confirmed by capturing the raw request and replaying it with/without that cookie.

PR description updated to match.

logicog:
Thanks for spotting this issue and providing the fix. Could you shorten the comments a bit, though, or do away with some of them entirely? The reasons for doing the fix are already in the commit message and can always be retrieved later on, when one wants to understand why the changes were done. Comments are there to explain the unexpected and not self-explanatory in the code or why something that appears strange is nevertheless correct, but basically the lines you add are more-or-less self-explanatory.

DrDoof:
Good point, thanks - trimmed them down. I kept only the bits that explain something the code can't show on its own: that ';' is a cookie separator, why the scan matches "session" rather than "session=" (relying on is_word()'s separator handling), and why Connection: close is required given this httpd closes after every response. The bug-history narrative is gone since it's already in the commit messages.

I put it in a separate "shorten comments" commit for now; happy to squash it into the three originals if you'd prefer a clean history for merge.

logicog:
Thanks a lot for fixing this issue!

Issue #292, opened 2026-07-21
Not work with SFP+ module, but works with DAC/AOC cable.


sempr:
My board is HiSource K0402WS-V3.0

I flashed the latest version,  _481c02c_ and tried using an DAC cable or AOC cable, it works well.

The module is *Hisense LTF1303-BH+* with two “Single-mode Optical Fiber” connected.

But when I use a SFP+ module **it is working with the factory firmware, not with the RTLPlayground though.** , The infomation showed is like this

```
Link speed	:	Down
Vendor	:	Hisense
Model	:	LTF1303-BH+
Serial	:	H8478B******
Temp	:	47.33 ℃
Vcc	:	3.30 V
TX-Fault	:	false
TX-Disabled	:	false
TX-Bias	:	40.2 mA
TX-Power	:	0.599 mW / -2.22 dBm
RX-Power	:	0.073 mW / -11.38 dBm
RX-LOS	:	**false**
```

When the optical fiber is not connected and only the module connected, the information is like this.

```
Link speed	:	Down
Vendor	:	Hisense
Model	:	LTF1303-BH+
Serial	:	H8478B*****
Temp	:	47.14 ℃
Vcc	:	3.31 V
TX-Fault	:	false
TX-Disabled	:	false
TX-Bias	:	39.1 mA
TX-Power	:	0.597 mW / -2.24 dBm
RX-Power	:	0.000 mW / -40.00 dBm
RX-LOS	:	false
```

The infomation seems to say the sfp+ module is successfully recogonized but not turned right working mode and the Link speed is always "Down".

tofurky:
Can you provide output from the "sfp" command via serial?

I ran into an issue similar to this with one of my modules. It was due to the existing range of supported rates not including the rates my SFP+ reported, see https://github.com/logicog/RTLPlayground/commit/73b99b2f081c30513d5d812be81f939f76139cac

Example output - note the "Rate" field 0x64:
```
> sfp

Slot 1 - Rate: 67  Encoding: 00
OEM             SFP-H10GB-CU1M  R   
Options: 00

Slot 2 - Rate: 64  Encoding: 03
AZORES          WAS-110         V1.0
Options: 68
Temp: 34c0
Vcc: 7d30
TX Bias: 14a0
TX Power: 88b6
RX Power: 0071
Laser: 7d00
State: 00
```

Symptom was the module was detected, but the link never came up.

[...]

PR #295, opened 2026-07-24
Fix rate range for SDS_10GR return value

sempr:
Broaden range of SFP+ rates mapped to SDS_10GR
A commonly used SFP+ form-factor 10GBase-LR (Hisense LTF1303-BH+) reports 0x62(BR Nominal: 9800MBd) at power on and then switches to 0x64 after booting up.

0x60 -> 9.6gbps
0x62 -> 9.8gbps
0x63 -> 9.9gbps
0x64 -> 10gbps
0x6f -> 11.1gbps
0x80 -> 12.8gbps

These are outside of the previous 0x63-0x6f range which prevented the link from ever being brought up. So I changed the range to 0x60-0x7f

The issue #292

### DIFF
--- a/rtlplayground.c
+++ b/rtlplayground.c
@@ -1210,7 +1210,7 @@ static inline uint8_t sfp_rate_to_sds_config(register uint8_t rate)
 		return SDS_1000BX_FIBER;
 	if (rate >= 0x19 && rate <= 0x20)  // Ethernet 2.5 GBit
 		return SDS_HSG;
-	if (rate >= 0x63 && rate < 0x70)
+	if (rate >= 0x62 && rate < 0x70)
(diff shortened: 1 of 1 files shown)


tofurky:
The commit message (based on my previous one) here isn't accurate. `Hisense LTF1303-BH+` isn't an XGS-PON ONT, it's 10GBase-LR, and there's not evidence that this one ever reports anything other than 0x62.

I am unsure if the range should be arbitrarily widened past what is necessary for the device (0x62), but it might prevent other outlier hardware from having issues.

sempr:
Only changed the lower range from 0x63 to 0x62 now.
And updated the commit messages.

logicog:
Thanks for fixing this!

PR #297, opened 2026-07-25
system: configurable hostname + show model on the System page

DrDoof:
Two small, self-contained additions to the System Settings page and
`/information.json`, both branched off the current `main`:

**1. Configurable hostname**
A device hostname settable from the CLI (`hostname <text>`) and the
System Settings page. It is:
- sanitized on ingest to JSON-safe printable ASCII (max 23 chars),
- seeded to `RTLPlayground` at boot *before* the startup-config replay,
  so a saved `hostname` overrides the default rather than being clobbered,
- persisted through the startup config like the other settings
  (validator + overwrite rule in `config.js`),
- reported in `/information.json` as `hostname`.

It lives in the shared header (`rtl837x_common.h`) so other modules can
read it — e.g. an LLDP series can advertise it as the System Name TLV.

**2. Show the model on the System page**
`machine.machine_name` is already reported as `hw_ver`; it is now
surfaced read-only on the System Settings page so the exact build target
is visible in the UI. No new JSON field. Includes i18n for en/ja/zh.

Built and verified on `SWTGW218AS`.


rtlplayground.c
```diff
@@ -2145,6 +2146,7 @@ void main(void)
 	vlan_setup();
 	port_l2_setup();
 	igmp_setup();
+	strcpy((__xdata uint8_t *)hostname, "RTLPlayground");	/* default until "hostname ..." replay */
```
plaes: How hard would be to append last two or four bytes from mac?

feelfree69:
Cool feature, thanks for contributing!
The idea of @plaes, having more unique hostnames by default, seems also a good idea.
Do you think you can add this?

cmd_parser.c
```diff
@@ -1524,6 +1524,18 @@ void cmd_parser(void) __banked
...
+			 * printable ASCII (<=23 chars), stored in the shared hostname. */
+			__xdata uint8_t *hp = &cmd_buffer[cmd_words_b[1]];
+			uint8_t hn = 0;
+			if (cmd_words_len >= 2)
+				while (hn < 23 && hp[hn] && hp[hn] != '\r' && hp[hn] != '\n') {
+					hostname[hn] = (hp[hn] < 0x20 || hp[hn] > 0x7e
+						|| hp[hn] == '"' || hp[hn] == '\\') ? '.' : hp[hn];
+					hn++;
+				}
+			hostname[hn] = 0;
```
vDorst: I have tested it and also see some improvement points.
1. Just using `hostname` command without arguments clears the string. 
2. We can just force that `hostname` command only accept none or one argument.
   None: Show the current hostname, more then one show an error.
3. I also see that the code gen is better when `var = *ptr++`. Instead of indexing into an array like `var = hostname[hn]`;

This is what I made while testing and optimizing this PR.
Feel free to use it.
```c
		} else if (cmd_compare(0, "hostname")) {
			/* hostname <text>: rest of the line, sanitized to JSON-safe
			 * printable ASCII (<=23 chars), stored in the shared hostname. */
			__xdata uint8_t *dst = &hostname;
			if (cmd_words_len == 1) {
				print_string_x(dst);
			} else if (cmd_words_len == 2) {
			    __xdata uint8_t *hp = &cmd_buffer[cmd_words_b[1]];

				for (uint8_t hn = 0; hn < sizeof(hostname) - 1; hn++) {
					uint8_t c = *hp++;

					if (c == '\0' || c == '\n' || c == '\r')
						break;
					
					if (c < 0x20 || c > 0x7e || c == '"' || c == '\\')
						c = '.';
					
					*dst++ = c;
				};
				*dst = '\0';
			} else {
				print_string("Hostname invalid. No spaces!");
			}
```

rtlplayground.c
vDorst: Is a nice addition.

vDorst (changes requested):

html/system.js
```diff
@@ -123,6 +131,8 @@ function fetchIP() {
       document.getElementById("ip").value=s.ip_address;
       document.getElementById("netmask").value=s.ip_netmask;
       document.getElementById("gw").value=s.ip_gateway;
+      if (document.getElementById("hostname")) document.getElementById("hostname").value = s.hostname || "";
+      if (document.getElementById("model")) document.getElementById("model").textContent = s.hw_ver || "";
```
vDorst: Are all these checks and default-values needed?
With the other values we don't do that.

Maybe convert it to a generic function, so all the values can have these checks and reduce the JS code.
Some AI-code.
```js
function fillFromJson(data, map) {
  for (const [jsonKey, elementId] of Object.entries(map)) {
    const el = document.getElementById(elementId);
    if (el) el.value = data[jsonKey] ?? "";
  }
}

// usage
fillFromJson(s, {
  ip_address: "ip",
  ip_netmask: "netmask",
  ip_gateway: "gw"
});
```

cmd_parser.c
DrDoof: Thanks — all three fixed in `1312012`.

`hostname` on its own now prints the current name; exactly one argument sets it; anything longer is an error rather than silently keeping the first word. I kept the sanitizer (printable ASCII, `"` and `\` folded to `.`) because the name is emitted into `/information.json`, and switched the copy to the pointer walk you suggested.

rtlplayground.c
DrDoof: Done in `94716ce`: the default is now `RTLPlayground-<last three MAC octets>`, e.g. `RTLPlayground-1ef924`.

I went with three octets rather than two — the first three are the fixed vendor part, so two octets would collide between units of the same batch far too easily, and this keeps the name short enough (20 of 23 usable characters) that a configured name still fits. A `hostname ...` line in the startup config overrides it exactly as before.

One implementation note: the obvious loop over the octets put its locals in the 8051's internal-RAM overlay, which overflows on a build with every feature enabled, so the assignments are spelled out.

html/system.js
DrDoof: Fair point — removed in `1194823`; both fields are now assigned like the neighbouring ones. I left the generic helper out for now: with three fields it would be more indirection than it saves, but it is a sensible refactor once the page grows.

DrDoof:
@feelfree69 @vDorst @plaes — thanks for the review, all points are addressed:

* `hostname` alone no longer wipes the name (it reports it); one argument sets it, more than one is an error. Pointer walk as suggested.
* Default name now carries the MAC tail: `RTLPlayground-1ef924`.
* The System-page fields lost their defensive checks.

The changes are built but not yet exercised on hardware — my switch currently runs a build with other pending work on it, so I did not want to claim more than I verified. Happy to flash and confirm the default name and the CLI behaviour if that helps before merging.

cmd_parser.c
DrDoof: Follow-up on point 3, since it is easy to get wrong by intuition: I measured it rather than assuming.

Same branch, same compiler, only the copy loop differs (everything else in the handler identical), BANK2 code size:

| variant | bytes |
|---|---|
| `hostname[hn]` indexing | 37820 |
| `*ptr++` walk | 37821 |

So in this particular loop the pointer walk is a byte *larger* — i.e. the two are equivalent here, sdcc resolves the constant-base indexing just as well. I kept your version anyway (it reads better and matches the idiom you use elsewhere), but it may be worth knowing before refactoring other loops for the same reason.

For completeness, the whole change costs +67 bytes over the branch point: the extra branches plus two message strings. Trimming the error text to a terse one-liner brings it to +49 if that matters more than the wording.

rtlplayground.c
vDorst, line 1: Some thoughts after reviewing.

I think this can be it's own function.
This should be called after `execute_config()`.
The new function check the hostname length is when zero and then fills it which the default hostname. So it is just reducing some boottime when the hostname is set.

You can use the `itohex()` function to lookup hex-char which also ignore the upper nibble. So you don't to have to byte-shift and mask it.
Just use swap-instuction.


This is what I make to also make it loop and use the `itohex()`.
```patch
index 2675e90..4a02a46 100644
--- a/rtlplayground.c
+++ b/rtlplayground.c
@@ -25,6 +25,7 @@
 #include "machine.h"
 #include "phy.h"
 #include "syslog.h"
+#include "httpd/page_impl.h"
 
 extern __code const struct machine machine;
 extern __xdata uint32_t flash_size;
@@ -2005,6 +2006,51 @@ void check_and_flash_update_image(void)
        }
 }
 
+
+void set_hostname_default() {
+       __xdata uint8_t *ptr = &hostname;
+
+       // return when the hostname is already set.
+       if (*ptr != '\0') {
+               return;
+       }
+
+       __code uint8_t * src = "RTLPlayground-";
+       // opencode strcpy without NULL-termination.
+       while(1) {
+               uint8_t val = *src++;
+
+               if (val == '\0') {
+                       break;
+               }
+
+               *ptr++ = val;
+       }
+
+       // Append last byte mac-addres value as hex-string.
+       for (uint8_t i = 3; i < 6; i++) {
+               uint8_t val = uip_ethaddr.addr[i];
+               __bit again = true;
+               // loops two times so a two byte hex string generated.
+               while (1) {
+                       // force 'swap' instruction.
+                       // itohex() will ignore the upper-nibble.
+                       val = (val >> 4) | (val << 4);
+
+                       *ptr++ = itohex(val);
+
+                       // ideally this should make use of the jbc-instruction.
+                       if (again) {
+                               again = false;
+                       } else {
+                               break;
+                       }
+               }
+       }
+
+       *ptr = '\0';
+}
+
 void main(void)
 {
        ticks = 0;
@@ -2146,19 +2192,10 @@ void main(void)
        vlan_setup();
        port_l2_setup();
        igmp_setup();
-       /* Default name carries the tail of the MAC, so several switches on one
-        * network are distinguishable out of the box (suggested in review).
-        * Overridden by a "hostname ..." line in the startup config. */
-       strcpy((__xdata uint8_t *)hostname, "RTLPlayground-");
-       /* Spelled out rather than looped: locals here land in the 8051's
-        * internal-RAM overlay, which is full on a build with every feature on. */
-       hostname[14] = hex[uip_ethaddr.addr[3] >> 4];
-       hostname[15] = hex[uip_ethaddr.addr[3] & 0xf];
-       hostname[16] = hex[uip_ethaddr.addr[4] >> 4];
-       hostname[17] = hex[uip_ethaddr.addr[4] & 0xf];
-       hostname[18] = hex[uip_ethaddr.addr[5] >> 4];
-       hostname[19] = hex[uip_ethaddr.addr[5] & 0xf];
-       hostname[20] = '\0';
+
+       // init hostname
+       hostname[0] = '\0';
+
        bandwidth_setup();
        uip_init();
        uip_arp_init();
@@ -2182,6 +2219,9 @@ void main(void)
        early_boot_handle_button();
 
        execute_config();
+
+       set_hostname_default();
+
        print_cmd_prompt();
        idle_ready = 1;
```

Also reduces code size.
```patch
diff output/SWGT024_V2_0_MANAGED/rtlplayground.mem ../rtlplayground.mem
29c29
<    ROM/EPROM/FLASH  0x0000   0x2d3cc  95206    16777216
---
>    ROM/EPROM/FLASH  0x0000   0x2d3d2  95137    16777216
```

feelfree69: >             // loops two times so a two byte hex string generated.
>               while (1) {
>                       // force 'swap' instruction.
>                     // itohex() will ignore the upper-nibble.
>                       val = (val >> 4) | (val << 4);
>
>                       *ptr++ = itohex(val);
>
>                       // ideally this should make use of the jbc-instruction.
>                       if (again) {
>                               again = false;
>                       } else {
>                               break;
>                       }
>               }
>       }

Te be honest: I don't like this kind of optimized code at all. 
It's hard to understand why it's written like this (and hard to understand what it does!), and it's not time critical at all. 
So all it does, it saves a few bytes of code size - but we're not running out of flash space, do we?

vDorst: I changed the code to make it loop because it was not possible without this.
Reduce code side is just nice.

Maybe this can be put in a separate function, it is used in other places.
```c
        // loops two times so a two byte hex string generated.
          while (1) {
                  // force 'swap' instruction.
                // itohex() will ignore the upper-nibble.
                  val = (val >> 4) | (val << 4);

                  *ptr++ = itohex(val);

                  // ideally this should make use of the jbc-instruction.
                  if (again) {
                          again = false;
                  } else {
                          break;
                  }
          }
```

feelfree69: But isn't your code-snippet equal to;
```
// itohex() will ignore the upper-nibble.
*ptr++ = itohex(val>>4);
*ptr++ = itohex(val);
```
?

vDorst: Technically it can be used in a other place.

```patch
index cb9125b..d8f1789 100644
--- a/httpd/page_impl.c
+++ b/httpd/page_impl.c
@@ -44,6 +44,7 @@ extern __xdata char sfp_module_vendor[2][17];
 extern __xdata char sfp_module_model[2][17];
 extern __xdata char sfp_module_serial[2][17];
 extern __xdata uint8_t sfp_options[2];
+extern __xdata uint8_t * print_hex_value(__xdata uint8_t * ptr, uint8_t val);
 
 __code uint8_t * __code HTTP_RESPONCE_JSON = "HTTP/1.1 200 OK\r\nConnection: close\r\nContent-Type: application/json\r\n\r\n";
 __code uint8_t * __code HTTP_RESPONCE_TXT = "HTTP/1.1 200 OK\r\nContent-Type: text/plain\r\n\r\n";
@@ -71,12 +72,8 @@ void char_to_html(char c)
 //  Convert uint8_t to ascii HEX char.
 void byte_to_html(uint8_t val)
 {
-       uint8_t cnt = 2;
-       do {
-               val = (val >> 4) | (val << 4);
-               charhex_to_html(val);
-               cnt -= 1;
-       } while(cnt);
+       print_hex_value(&outbuf[slen], val);
+       slen += 2;
 }
 
 /* Converts a uint8_t to raw string.
index 66df88a..36e1200 100644
--- a/rtl837x_common.h
+++ b/rtl837x_common.h
@@ -176,4 +176,5 @@ void sds_write_v(uint8_t sds_id, uint8_t page, uint8_t reg, uint16_t v);
 void sds_config_mac(uint8_t sds, uint8_t mode);
 void sds_config(uint8_t sds, uint8_t mode);
 void handle_sfp(void);
+__xdata uint8_t * print_hex_value(__xdata uint8_t * ptr, uint8_t val);
 #endif
index 2675e90..3c3ea15 100644
--- a/rtlplayground.c
+++ b/rtlplayground.c
@@ -25,6 +25,7 @@
 #include "machine.h"
 #include "phy.h"
 #include "syslog.h"
+#include "httpd/page_impl.h"
 
 extern __code const struct machine machine;
 extern __xdata uint32_t flash_size;
@@ -2005,6 +2006,56 @@ void check_and_flash_update_image(void)
        }
 }
 
+__xdata uint8_t * print_hex_value(__xdata uint8_t * ptr, uint8_t val) {
+               __bit again = true;
+               // loops two times so a two byte hex string generated.
+               while (1) {
+                       // force 'swap' instruction.
+                       // itohex() will ignore the upper-nibble.
+                       val = (val >> 4) | (val << 4);
+
+                       *ptr++ = itohex(val);
+
+                       // ideally this should make use of the jbc-instruction.
+                       if (again) {
+                               again = false;
+                       } else {
+                               break;
+                       }
+               }
+
+       return ptr;
+}
+
+void set_hostname_default(void) {
+       __xdata uint8_t *ptr = &hostname;
+
+       // return when the hostname is already set.
+       if (*ptr != '\0') {
+               return;
+       }
+
+       __code uint8_t * src = "RTLPlayground-";
+       // opencode strcpy without NULL-termination.
+       while(1) {
+               uint8_t val = *src++;
+
+               if (val == '\0') {
+                       break;
+               }
+
+               *ptr++ = val;
+       }
+
+       // Append last byte mac-addres value as hex-string.
+       for (uint8_t i = 3; i < 6; i++) {
+               uint8_t val = uip_ethaddr.addr[i];
+               ptr = print_hex_value(ptr, val);
+       }
+
+       *ptr = '\0';
+}
+
 void main(void)
 {
        ticks = 0;
@@ -2146,19 +2197,10 @@ void main(void)
        vlan_setup();
        port_l2_setup();
        igmp_setup();
-       /* Default name carries the tail of the MAC, so several switches on one
-        * network are distinguishable out of the box (suggested in review).
-        * Overridden by a "hostname ..." line in the startup config. */
-       strcpy((__xdata uint8_t *)hostname, "RTLPlayground-");
-       /* Spelled out rather than looped: locals here land in the 8051's
-        * internal-RAM overlay, which is full on a build with every feature on. */
-       hostname[14] = hex[uip_ethaddr.addr[3] >> 4];
-       hostname[15] = hex[uip_ethaddr.addr[3] & 0xf];
-       hostname[16] = hex[uip_ethaddr.addr[4] >> 4];
-       hostname[17] = hex[uip_ethaddr.addr[4] & 0xf];
-       hostname[18] = hex[uip_ethaddr.addr[5] >> 4];
-       hostname[19] = hex[uip_ethaddr.addr[5] & 0xf];
-       hostname[20] = '\0';
+
+       // clear hostname
+       hostname[0] = '\0';
+
        bandwidth_setup();
        uip_init();
        uip_arp_init();
@@ -2182,6 +2224,9 @@ void main(void)
        early_boot_handle_button();
 
        execute_config();
+
+       set_hostname_default();
+
        print_cmd_prompt();
        idle_ready = 1;
```

feelfree69: Sorry to bother you again with the same question:

Isn't this function doing the same as yours:

```
_xdata uint8_t * print_hex_value(__xdata uint8_t * ptr, uint8_t val) {
  *ptr++ = itohex(val>>4);
  *ptr++ = itohex(val);
  return ptr;
}
```

vDorst: No problem.

yes, but I can't find this function you point too.
In which file did you find that function?
```
# rg -tc itohex
httpd/page_impl.c
54:     outbuf[slen++] = itohex(c);
143:            // force the swap instruction, itohex() ignores the upper nibble.

httpd/httpd.c
309:            b[--bytes] = itohex(sfr_data[i]);
311:            b[--bytes] = itohex(sfr_data[i] >> 4 | sfr_data[i] << 4);

httpd/page_impl.h
22:inline char itohex(uint8_t val) {

# rg -tc print_hex_value
<no results>
```

feelfree69: I was referring to your code/patch with *your* function `_xdata uint8_t * print_hex_value(__xdata uint8_t * ptr, uint8_t val)`, in the posting directly above mine.

vDorst: ah

```c
__xdata uint8_t * print_hex_value(__xdata uint8_t * ptr, uint8_t val) {
  *ptr++ = itohex(val>>4);
  *ptr++ = itohex(val);
  return ptr;
}
```

This function reduced the code with 17-bytes.
```c
__xdata uint8_t * print_hex_value(__xdata uint8_t * ptr, uint8_t val) {
	__bit again = true;
	// loops two times so a two byte hex string generated.
	while (1) {
		// force 'swap' instruction.
		// itohex() will ignores the upper-nibble.
		val = (val >> 4) | (val << 4);
		*ptr++ = itohex(val);
		// ideally this should make use of the jbc-instruction.
		if (again) {
			again = false;
		} else {
			break;
		}
	}
    return ptr;
}
```

feelfree69: Couldn't believe that 3 lines of C-code produce significant more code than 9 lines.
But then I found out that it is due to 'itohex()' is an inline function.... that explains it.

vDorst: the results are without inline itohex() BTW.

> Couldn't believe that 3 lines of C-code produce significant more code than 9 lines.

That is why it is writing like that, to write code so it makes better use of the instruction that can do more thing at once like `jbc`-instruction.
Or the `swap`-instruction vs 4-bit-shifting. Which result in a `swap` and `anl a, 0x0F`-instruction.

feelfree69: Ok, understood. 
Personally, I wouldn't write such code if there's no clear benefit. Better asm code for me is not a benefit, if it's not in a busy loop, irq routine or similar.

But that's just a matter taste, so I'm fine with it.

DrDoof:
Following up on my earlier note that this was built but not yet exercised — I have now run it on the switch (SWTGW218AS). Two of the three points are confirmed on hardware:

* `hostname` with no argument **reports** the name instead of wiping it — returns `SWTGW218AS` here.
* `hostname foo bar` is rejected: `Error: hostname [name] - the name must not contain spaces`.

The third one, the MAC-derived default, I could not observe on this unit: its startup config carries an explicit `hostname SWTGW218AS` line, so the default never gets a chance to apply. For the record, this switch's MAC is `06:05:16:1e:f9:24`, which by the code in this PR yields `RTLPlayground-1ef924` — the value I quoted earlier. Confirming it for real means clearing the config line and rebooting; I can do that and report back if you would like it verified before merging, but I did not want to touch a working unit's configuration unasked.

Also, on the `itohex()` discussion between @vDorst and @feelfree69 — I have no strong preference and am happy either way. If you want the hand-optimised form in, say the word and I will fold it in.

vDorst:
It tested with no hostname set and with a hostname set in the config.

I prefer the hand-optimized one for now, already put some effort in to make it.
It doesn't directly save any SRAM space, that is what I hoped for.
But it does on code-size.

I do agree with @feelfree69 statement. When it is not time-critical hand-optimizing-code is mostly not worth it.

DrDoof:
Folded in your version — thanks, it is better than what I had in a way I had not considered.

`set_hostname_default()` is now its own function, called right after `execute_config()` and returning immediately if the config already set a name, so a configured switch does no work for it at all. The hex tail goes through `itohex()` with the swap, as you wrote it.

The part I appreciate most: my version spelled out all six assignments by hand, with a comment explaining that a loop was not an option because the locals land in the internal-RAM overlay, which is full on a build with everything enabled. Moving the code into its own function makes that constraint simply not apply — so the comment documenting the obstacle could go away along with the obstacle. Working around a limit and changing the context where it holds are not the same thing, and you picked the better one.

On the wider point you and @feelfree69 settled: agreed, hand-optimising outside hot paths is usually not worth it. Here it is a boot-time one-shot, so the argument for it is code size rather than speed — and it happens to read better too, which is not always how that trade goes.

Also, thank you for testing both cases (no hostname set, and one set in the config). That covered the one thing I could not check here without clearing a working unit's configuration, and it was the last open item from my side.

DrDoof:
Verified on hardware now, both cases, on an integrated build (LACP + STP + this change) on a SWTGW218AS:

* **Config wins**: with `hostname SWTGW218AS` in the startup config, `information.json` reports `SWTGW218AS` — the derived name never runs.
* **Derived default**: booted with no `hostname` line in the config, the switch came up as `RTLPlayground-1ef924`. Its MAC is `06:05:16:1e:f9:24`, so that is exactly the expected tail.

One thing worth recording, because it decided the final shape of the patch. Your loop version does not link in an integrated image:

```
?ASlink-Error-Could not get 8 consecutive bytes in internal RAM for area OSEG.
```

Locals here — counters and pointers alike — land in the 8051's internal-RAM overlay, and with LACP and STP both enabled that overlay is exhausted. I tried three ways out before giving up on the loop: hoisting the counters to `__xdata`, hoisting the pointers as well, and relying on the move into a separate function. None helped; the overlay is shared across the whole image, and `itohex()` is inline so it brings its own frame regardless.

So I kept the two structural improvements you suggested — the separate function, and calling it after `execute_config()` with an early return when the config already set a name — but the body stays without local variables. I was wrong in my earlier comment when I said moving the code into its own function makes the constraint not apply; it does not, and this only shows up when LACP and STP are compiled in together, which is why the branch builds fine on its own.

The comment in the code now carries the linker error verbatim and lists what was tried, so the next person has something falsifiable rather than a bare "do not loop here".

DrDoof:
I tried your `print_hex_value()` and it does not fit — but the reason is narrower than I expected, and the page_impl.c half of your patch may well be fine on its own.

**What I built.** The helper exactly as you wrote it (the loop variant, the one you measured 17 bytes smaller), placed in `rtlplayground.c` rather than `page_impl.c`: that file is `#pragma codeseg BANK1` while `rtlplayground.c` is HOME, and since HOME is always mapped, putting it there lets both sides call it without `__banked`. Declared in `rtl837x_common.h` as you suggested, `byte_to_html()` refactored to `print_hex_value(&outbuf[slen], val); slen += 2;`, and `set_hostname_default()` calling it three times with constant addresses so it needs no locals of its own.

**The hostname branch builds fine.** The integrated image — LACP and STP compiled in together — does not:

```
?ASlink-Error-Could not get 8 consecutive bytes in internal RAM for area OSEG.
```

**Then I narrowed it.** Reverted `byte_to_html()` to its original loop, leaving the helper with `set_hostname_default()` as its only caller. Still fails, identically. So the blocker is not the HTML refactor: it is the call itself out of `set_hostname_default()`. That function apparently cannot afford a callee with parameters at all on a full image.

That also corrects something I got wrong when I reasoned about it beforehand. I expected a shared helper to *relieve* overlay pressure, since `byte_to_html()` would give up its `cnt`. It does not work that way — SDCC allocates the overlay along the call graph, so turning a leaf into a non-leaf adds a frame to every path that reaches it rather than trading one for another.

**What this does not rule out:** your `page_impl.c` change on its own, with the hostname code left as it is. I did not build that combination, and `byte_to_html()` already had a local, so it may well cost nothing. If you want that part in, it stands independently of this PR and I would be glad to see it land.

For this branch I have kept the version without locals — the spelled-out assignments — since that is the only form that links in an integrated image. The comment above it now carries the linker error verbatim and lists what was tried, so nobody has to rediscover it.

One small thing while I was in there: `hex[]` is not dead if the hostname code stops using it — `cmd_parser.c` and `rtl837x_port.c` still index it for VLAN name encoding.

DrDoof:
@vDorst when you get a chance, could you take another look here? The review that's blocking is from 30 July and predates the changes, so I think it's just stale rather than unaddressed.

Since then: `hostname` with no argument reports the current name instead of clearing it, the MAC-derived default is verified on hardware both ways (set and unset), and I looked into folding the hex printing into a shared `print_hex_value()` — it doesn't fit, OSEG is full and the call graph can't take the extra overlay. I wrote that up in the thread rather than quietly dropping it.

No rush, just don't want it sitting there because of a review nobody meant to leave open.

vDorst:
I am going merge it now as is.
We can always improve it with a new PR.

Thanks

PR #298, opened 2026-07-25
uip: cap TCP MSS to 1460 to survive jumbo-MTU clients

DrDoof:
### Symptom

Opening the web UI from a host on a **jumbo-MTU** link (e.g. a NIC at MTU 8192) hangs the management CPU or returns blank pages: the browser fetches `/`, the first small JSON responses arrive, then a larger asset (`i18n.js`, `config.js`, …) never completes and the switch stops answering ICMP until a cold power-cycle. From `curl` on a standard-MTU host everything works, which is what makes it easy to miss.

### Root cause

`UIP_TCP_MSS` was derived from the buffer size: `UIP_BUFSIZE - UIP_LLH_LEN - UIP_TCPIP_HLEN` ≈ **2158**. A jumbo-MTU peer advertises a large MSS, so the switch settles on its own 2158 for egress. The CPU-port TX path cannot send a segment that big — the frame is silently dropped, so every multi-segment reply is lost. Worse, when the management VLAN is configured, the outbound 802.1Q tag is inserted into `uip_buf` after the payload is placed, and a full-MSS frame plus that tag runs past the end of the buffer, corrupting adjacent state and wedging the CPU.

Captured with `tcpdump`: the SYN-ACK advertised `mss 2158`, `curl` on a jumbo link advertised `mss 8152`, and the first multi-segment GET was retransmitted with no response. Forcing the client path MTU to 1500 (MSS 1460) made every asset serve correctly, confirming the segment size as the trigger.

### Fix

Cap `UIP_TCP_MSS` at one standard Ethernet segment (1460). The receive buffer is unchanged, so this only bounds the segment size the switch *emits* (and the window it advertises); 1460 stays well within the buffer ceiling. This fixes both the truncated responses and the hang for every client, regardless of the client's MTU.

### Verification

Built for `SWTGW218AS` and flashed. With the host NIC back at MTU 8192 (MSS negotiated to 1460): `config.js` (3071 B), `i18n.js` (19685 B), `main.js`, `stat.js` all serve in full and the switch stays responsive across repeated fetches. Previously the same requests under jumbo returned an empty body and/or hung the CPU.

### DIFF
--- a/uip/uip-conf.h
+++ b/uip/uip-conf.h
@@ -106,9 +106,25 @@ typedef unsigned short uip_stats_t;
 /**
  * uIP buffer size.
  *
+ * Sized to the largest frame the CPU port accepts on ingress. The NIC drops
+ * anything above that in hardware, so a larger buffer only costs XDATA, while
+ * a smaller one would be overrun by the receive DMA.
+ *
+ * Measured on a SWTGW218AS with ICMP, which bypasses MSS and so probes the
+ * hardware directly: a 1502-byte payload is answered and 1503 is not, putting
+ * the largest frame we are handed at UIP_LLH_LEN + 20 + 8 + 1502 = 1556 bytes
+ * of uip_buf.
+ *
+ * UIP_TCP_MSS derives from this as 1490, which is also the largest TCP payload
+ * that gets through, so the switch advertises exactly what it can receive.
+ *
+ * The limit belongs to the NIC rather than to an ingress port: RX_CTRL sits in
+ * the NIC register block and the CPU tag is inserted for every trapped frame,
+ * so how the frame was tagged on the wire does not change the figure.
+ *
  * \hideinitializer
  */
-#define UIP_CONF_BUFFER_SIZE     2200
+#define UIP_CONF_BUFFER_SIZE     1556
 
 /**
  * CPU byte order.
--- a/uip/uip-conf.h
+++ b/uip/uip-conf.h
@@ -106,21 +106,12 @@ typedef unsigned short uip_stats_t;
 /**
  * uIP buffer size.
  *
- * Sized to the largest frame the CPU port accepts on ingress. The NIC drops
- * anything above that in hardware, so a larger buffer only costs XDATA, while
- * a smaller one would be overrun by the receive DMA.
- *
- * Measured on a SWTGW218AS with ICMP, which bypasses MSS and so probes the
- * hardware directly: a 1502-byte payload is answered and 1503 is not, putting
- * the largest frame we are handed at UIP_LLH_LEN + 20 + 8 + 1502 = 1556 bytes
- * of uip_buf.
- *
- * UIP_TCP_MSS derives from this as 1490, which is also the largest TCP payload
- * that gets through, so the switch advertises exactly what it can receive.
- *
- * The limit belongs to the NIC rather than to an ingress port: RX_CTRL sits in
- * the NIC register block and the CPU tag is inserted for every trapped frame,
- * so how the frame was tagged on the wire does not change the figure.
+ * Sized to the largest frame the CPU port accepts on ingress: anything above
+ * that the NIC drops in hardware, so a larger buffer only costs XDATA. Measured
+ * with ICMP, which bypasses MSS and so probes the hardware directly: a 1502-byte
+ * payload is answered and 1503 is not, which puts the frame at 1556 bytes of
+ * uip_buf. The limit is the NIC's rather than a port's, so how the frame was
+ * tagged on the wire does not change it.
  *
  * \hideinitializer
  */
--- a/uip/uip-conf.h
+++ b/uip/uip-conf.h
@@ -118,6 +118,13 @@ typedef unsigned short uip_stats_t;
 #define UIP_CONF_BUFFER_SIZE     1556
 
 /**
+ * Bytes of the buffer kept out of the advertised MSS.
+ *
+ * \hideinitializer
+ */
+#define UIP_CONF_BUFFER_EXTRA    30
+
+/**
  * CPU byte order.
  *
  * \hideinitializer
--- a/uip/uipopt.h
+++ b/uip/uipopt.h
@@ -299,16 +299,10 @@
 /**
  * The TCP maximum segment size.
  *
- * Deliberately below the buffer-derived ceiling of
- * UIP_BUFSIZE - UIP_LLH_LEN - UIP_TCPIP_HLEN, not at it. Advertising the
- * exact ceiling makes a full segment fill uip_buf to its last byte, and a
- * sender that takes us at our word then corrupts every large upload: the
- * firmware image arrives fully acknowledged, no retransmissions on the wire,
- * yet the CRC over the streamed body never matches. Measured with nothing but
- * the segment size changing: 1490-byte segments fail four times out of four,
- * 1460-byte segments succeed. Only macOS ever sends the full advertised size -
- * Linux halves its segments against a window this small - which is why the
- * failure hides so well.
+ * Deliberately a step below the buffer-derived ceiling, not at it: a segment
+ * that fills uip_buf to its last byte corrupts large uploads. Measured with
+ * nothing but the segment size changing, 1490 fails four times out of four and
+ * 1460 succeeds.
  */
 #define UIP_TCP_MSS     1460
 
--- a/uip/uipopt.h
+++ b/uip/uipopt.h
@@ -447,7 +447,7 @@ void uip_log(char *msg);
 #ifdef UIP_CONF_LLH_LEN
 #define UIP_LLH_LEN UIP_CONF_LLH_LEN
 #else /* UIP_CONF_LLH_LEN */
-#define UIP_LLH_LEN     ETHER_HEADER_SIZE + RTL_FRAME_DESC_SIZE
+#define UIP_LLH_LEN     (ETHER_HEADER_SIZE + RTL_FRAME_DESC_SIZE)
 #endif /* UIP_CONF_LLH_LEN */
 
 /** @} */
--- a/uip/uipopt.h
+++ b/uip/uipopt.h
@@ -299,10 +299,18 @@
 /**
  * The TCP maximum segment size.
  *
- * This is should not be to set to more than
- * UIP_BUFSIZE - UIP_LLH_LEN - UIP_TCPIP_HLEN.
+ * Deliberately below the buffer-derived ceiling of
+ * UIP_BUFSIZE - UIP_LLH_LEN - UIP_TCPIP_HLEN, not at it. Advertising the
+ * exact ceiling makes a full segment fill uip_buf to its last byte, and a
+ * sender that takes us at our word then corrupts every large upload: the
+ * firmware image arrives fully acknowledged, no retransmissions on the wire,
+ * yet the CRC over the streamed body never matches. Measured with nothing but
+ * the segment size changing: 1490-byte segments fail four times out of four,
+ * 1460-byte segments succeed. Only macOS ever sends the full advertised size -
+ * Linux halves its segments against a window this small - which is why the
+ * failure hides so well.
  */
-#define UIP_TCP_MSS     (UIP_BUFSIZE - UIP_LLH_LEN - UIP_TCPIP_HLEN)
+#define UIP_TCP_MSS     1460
 
 /**
  * The size of the advertised receiver's window.
--- a/uip/uipopt.h
+++ b/uip/uipopt.h
@@ -298,13 +298,8 @@
 
 /**
  * The TCP maximum segment size.
- *
- * Deliberately a step below the buffer-derived ceiling, not at it: a segment
- * that fills uip_buf to its last byte corrupts large uploads. Measured with
- * nothing but the segment size changing, 1490 fails four times out of four and
- * 1460 succeeds.
  */
-#define UIP_TCP_MSS     1460
+#define UIP_TCP_MSS     (UIP_BUFSIZE - UIP_LLH_LEN - UIP_TCPIP_HLEN - UIP_BUFFER_EXTRA)
 
 /**
  * The size of the advertised receiver's window.
@@ -382,6 +377,17 @@
 #else /* UIP_CONF_BUFFER_SIZE */
 #define UIP_BUFSIZE UIP_CONF_BUFFER_SIZE
 #endif /* UIP_CONF_BUFFER_SIZE */
+
+/**
+ * Bytes of uip_buf kept out of the advertised MSS.
+ *
+ * \hideinitializer
+ */
+#ifndef UIP_CONF_BUFFER_EXTRA
+#define UIP_BUFFER_EXTRA 0
+#else /* UIP_CONF_BUFFER_EXTRA */
+#define UIP_BUFFER_EXTRA UIP_CONF_BUFFER_EXTRA
+#endif /* UIP_CONF_BUFFER_EXTRA */
 
 
 extern __xdata uint8_t uip_buf[UIP_CONF_BUFFER_SIZE+2];


logicog:
Great that you identified this problem. Nice catch!
I am not sure about the root-cause though:
> The CPU-port TX path cannot send a segment that big — the frame is silently dropped, so every multi-segment reply is lost.

Have you investigated how large TX packets can actually be? Looking at the registers, the TX buffer size can be 0x800 = 2048, it is set to:
```
rtlplayground.c
1632:   REG_SET(RTL837X_REG_NIC_BUFFSIZE_TX, 0x7fe);
```
Not sure where the 2 bytes safety come from.

Also:
```
uip/uip-conf.h
111:#define UIP_CONF_BUFFER_SIZE     2200
```
The NIC-RX seems to have the same limits as the TX engine. So probably UIP_CONF_BUFFER_SIZE is too large and needs to be adjusted to the capabilities of the NIC.

What worries me is that we have a mechanism for fragmenting TX packets in httpd_appcall(), which allows to server large web-pages (but JSON is limited to TCP_OUTBUF_SIZE nevertheless). But there is no mechanism for re-assembling fragmented packets that are received. If we are too agressive with MSS, will this break receiving large chunks of JSON, i.e. are there already packets that may be received in the size range 1460-2048? 

My understanding would be that we need to adjust UIP_CONF_BUFFER_SIZE and TCP_OUTBUF_SIZE to work with the limitations of the TX/RX engine of the NIC, not artificially limit MSS.

DrDoof:
Sorry for the slow reply, and thanks for pushing on this — you were right to. I went and measured it, and my original wording was wrong, though not in the direction you'd expect.

The two engines don't have the same limits. RX is the tighter one at 1534, TX sits at your 2046, and it's RX that makes the cap necessary rather than cosmetic.

Taking TX first, since that's what you asked. You're right that `0x7fe` is 2048 minus 2 and I have no idea where those two bytes come from either. What I can tell you is that it behaves as a hard frame limit and not just a partition boundary. I built with `UIP_TCP_MSS=2100` so the wire frame (~2158) would land in the gap between TXSTOP 2046 and `uip_buf` 2202, then pulled a 19 KB asset over a jumbo link. The switch really did transmit 2100-byte segments — tcpdump on the peer has them, the NIC doesn't clamp to 1500 — but httpd came out of it corrupted and stayed corrupted until a reboot, including for requests arriving over a perfectly normal path afterwards. Since `uip_buf` was still under 2202 at that point, TXSTOP is what binds. Also worth flagging the asymmetry, because it bit me: RX loses frames cleanly, TX corrupts. The "silently dropped" in my PR text describes RX and is just wrong about TX.

RX is 1534, which I'm fairly confident about because it shows up four different ways. The firmware clears bits 16/17 of `RX_CTRL` — `sfr_mask_data(2, 0x03, 0)` at `rtlplayground.c:1727` — which selects index 0 of the SDK's `rtk_nic_RxMTU_t`, and the options there are 1534 / 2048 / 4096. Live `regget 0x785c` reads back `0x0400201b`, RXMTU zero. And behaviourally, with a build advertising MSS 2000 and a GET carrying a big `X-Pad` header, there's a sharp edge: payload up to 1490 B arrives in about 0.05 s, 1492 B and above never arrives and the sender retransmits after ~0.54 s. The datasheet is NDA so none of this comes from Realtek, but four independent things agreeing is good enough for me.

The overhead works out around 42–44 (IP 20 + TCP 20 + the small CPU tag). The Ethernet header isn't in there — it's stripped when the frame is trapped to the CPU — and there's no VLAN tag on RX, only on TX. So the largest MSS we can honestly advertise is about 1490, and 1460 is just the standard Ethernet number sitting a comfortable 30 bytes under it.

Where I'd push back is on adjusting the buffers instead of capping. I agree the buffer is oversized, and I did shrink it separately — 1600 works and saves 600 bytes of XDATA, 1450 hangs the 8051 outright, which brackets the floor nicely. But the two knobs aren't interchangeable. `uip_buf` is one buffer doing two jobs, RX DMA target and TX assembly, so it has to hold `max(RX, TX)`, and the RX floor is 1534. Your relation, `MSS = BUFSIZE − 66`, needs BUFSIZE around 1526 to yield MSS 1460 — under that floor, so it overflows on receive. The cap has to be decoupled from the buffer size. Hardcoding it isn't me dodging the real fix; it's the only way both constraints hold at once.

Your worry about large JSON receives I think resolves itself. The NIC drops anything over 1534 regardless of what we advertise, so segments in the 1460–2048 range were never receivable in the first place — advertising them just bought us retransmits. Capping MSS means we advertise what the hardware will actually take, which is also why the missing RX reassembly doesn't bite.

One practical note if you want to reproduce the RX number: ICMP bypasses MSS entirely, so `ping -s` probes the hardware limit directly without TCP hiding it behind its own windowing. My first attempt at this didn't do that and was worthless — RX MTU and MSS were both throttling the input, so every variant I tried looked identical and I convinced myself of something I hadn't actually shown.

Happy to split the buffer change into its own PR on top of this, or fold it in here, whichever you prefer.

logicog:
> The two engines don't have the same limits. RX is the tighter one at 1534, TX sits at your 2046, and it's RX that makes the cap necessary rather than cosmetic.

Did you test this in hardware, e.g. with packets of a given length being sent or received? I would assume the packet engine is symmetric. 
My preference would be to adjust the size of UIP_CONF_BUFFER_SIZE to fit into the send/receive buffers. It does not make sense to have data that we cannot send at the end of the buffer without implement our own TCP fragmentation. This will be only an issue for the dynamically generated JSON content. The rest of the data is already send piece by piece out of the the flash.

I have a further comment: my personal attention span to arrive at the decisive point in a comment on PRs is about 30 seconds. It appears that the comments have been made using AI, which shows in their considerable length. I personally am not able to read through all that text, especially when I get the feeling I am listening to an AI, I am thinking also about PR #303 , here, with endless text. Although I am very interested in that, I really am not able to work through all that text. No problem to write the code or the commit description with AI, but in my opinion, the comment discussions should be between humans. After all, for all Open Source projects, the reviewer's time is always the most precious (everyone wants to code and maybe to test), and reviewers want to be able to figure out quickly how much the human behind a PR has understood the problem in order to trust that the problem is fully understood and solved and merge the PR.

DrDoof:
Yes — `ping -s`, since ICMP skips MSS and you get at the hardware limit directly. 1534 arrives, 1536 doesn't. TX I checked separately with a build advertising MSS 2100: those frames really do go out, but httpd is corrupt afterwards. So the two directions aren't the same, though it turns out that doesn't matter here.

You're right about the knob. I was working backwards from 1460 as the target and never asked what MSS the RX floor actually permits — it's 1480, out of BUFSIZE 1546, and a full segment is then exactly 1534 on the wire. Buffer fits the largest frame the NIC takes, nothing unsendable at the end of it, no hardcoded constant. I'll redo it that way.

And point taken on the length.

DrDoof:
Reworked the way you suggested, and rebased onto current main.

The buffer is sized to what the hardware hands us rather than to a round number:
ICMP answers a 1502-byte payload and drops 1503, putting the frame at 1556 bytes
of uip_buf, which frees 644 bytes of XDATA. UIP_LLH_LEN was a bare sum too, so
everywhere it got subtracted the second term was added instead.

One thing I couldn't follow all the way. The advertised MSS stays a step below
the buffer edge rather than at it, 1460 rather than the 1490 it would derive to,
because a segment that fills uip_buf to its last byte corrupts large uploads.
Fully acknowledged, nothing retransmitted, CRC still fails. With only the segment
size changing, 1490 fails four times out of four and 1460 succeeds. I don't know
yet where it breaks.

DrDoof:
Ran into this on hardware today by accident, and it makes the case better than my
description does.

I flashed a build off current main onto my switch. The first page I opened,
/l2.json, killed management outright: HTTP gone, ping at 100% loss, still gone two
minutes later. The only way back was a reset over the serial console.

The same request from a host on a 1500 byte MTU answered 200 with 2221 bytes in
42 ms. The difference was the machine I was working from, which sits on an 8192
byte MTU interface. Main still derives UIP_TCP_MSS from UIP_BUFSIZE with nothing
capping it.

Worth adding that the switch never stopped switching. Traffic through it was
untouched while the management CPU was dead, which is probably why this is easy to
miss until it bites.

uip/uipopt.h
```diff
@@ -447,7 +449,7 @@ void uip_log(char *msg);
 #ifdef UIP_CONF_LLH_LEN
 #define UIP_LLH_LEN UIP_CONF_LLH_LEN
 #else /* UIP_CONF_LLH_LEN */
-#define UIP_LLH_LEN     ETHER_HEADER_SIZE + RTL_FRAME_DESC_SIZE
+#define UIP_LLH_LEN     (ETHER_HEADER_SIZE + RTL_FRAME_DESC_SIZE)
```
logicog, line 452: Good catch!

uip/uipopt.h
```diff
@@ -299,10 +299,12 @@
...
  *
- * This is should not be to set to more than
- * UIP_BUFSIZE - UIP_LLH_LEN - UIP_TCPIP_HLEN.
+ * Deliberately a step below the buffer-derived ceiling, not at it: a segment
+ * that fills uip_buf to its last byte corrupts large uploads. Measured with
+ * nothing but the segment size changing, 1490 fails four times out of four and
+ * 1460 succeeds.
  */
-#define UIP_TCP_MSS     (UIP_BUFSIZE - UIP_LLH_LEN - UIP_TCPIP_HLEN)
+#define UIP_TCP_MSS     1460
```
logicog, line 307: I do not understand why you set this and also UIP_CONF_BUFFER_SIZE directly. UIP_BUFSIZE is calculated from UIP_CONF_BUFFER_SIZE, so when requiring UIP_TCP_MSS <= 1460 to make the NIC happy, we can calculate backwards to get the buffer size:
```
UIP_TCP_MSS  = 1460 =  (UIP_BUFSIZE - UIP_LLH_LEN - UIP_TCPIP_HLEN)
-> 1460 + UIP_LLH_LEN + UIP_TCPIP_HLEN =  UIP_BUFSIZE
-> UIP_CONF_BUFFER_SIZE = 1460 + UIP_LLH_LEN + UIP_TCPIP_HLEN
```
If we need the buffer to be larger than that, we can introduce somehing like UIP_BUFFER_EXTRA and subtract it from UIP_BUFSIZE so that we definitely arrive at the right MSS size.

logicog:
While trying to understand which buffer size definitions come from where in uip. I noticed the following: uip is able to do de-fragmentation of packets. It is just not turned on:
```
/**
 * Turn on support for IP packet reassembly.
 *
 * uIP supports reassembly of fragmented IP packets. This features
 * requires an additonal amount of RAM to hold the reassembly buffer
 * and the reassembly code size is approximately 700 bytes.  The
 * reassembly buffer is of the same size as the uip_buf buffer
 * (configured by UIP_BUFSIZE).
 *
 * \note IP packet reassembly is not heavily tested.
 *
 * \hideinitializer
 */
#define UIP_REASSEMBLY 0

/**
 * The maximum time an IP fragment should wait in the reassembly
 * buffer before it is dropped.
 *
 */
#define UIP_REASS_MAXAGE 40
```
So, we may be able to have the cake and eat it in terms of accepting larger incoming packets, as long as we have a sufficiently large uip buffer. For that we could make use of something like UIP_BUFFER_EXTRA.

DrDoof:
Good find, I didn't know that code was sitting there. I went back over the measurements before answering, though, because I think reassembly ends up solving a different problem than the one this PR runs into.

Reassembly puts an IP datagram back together out of fragments, and each fragment still arrives as its own frame inside the link MTU. TCP doesn't produce fragments: it sizes its segments from the MSS we advertise, so a client on a jumbo link sends one oversized segment rather than several fragments. Turning reassembly on wouldn't change what that client puts on the wire.

There's also a ceiling under all of this that a bigger buffer can't lift. I measured the ingress limit with ICMP, which bypasses MSS and so probes the hardware directly: a 1502 byte payload gets answered and 1503 does not, which puts the frame at 1556 bytes of uip_buf. Anything above that never reaches uIP at all, so a larger buffer only costs XDATA.

And the failure here isn't "the packet was too large to receive". It's that a segment which fills uip_buf to its very last byte corrupts a large upload. With nothing but the segment size changing, 1490 failed four times out of four and 1460 succeeded. As long as the MSS stays derived from the buffer size, a larger buffer just moves that same edge somewhere else rather than removing it.

Cost is worth weighing too, since this part is tight: the reassembly code is around 700 bytes, and the reassembly buffer is another whole UIP_BUFSIZE of XDATA on top of the one we already carry, for a path I can't currently make the hardware take.

If you have a case in mind where fragments actually do arrive, I'm glad to go and measure it. I couldn't produce one over TCP.

uip/uipopt.h
DrDoof: You're right that writing both of them out by hand is odd, and the second suggestion works exactly. I've given the gap a name, `UIP_CONF_BUFFER_EXTRA`, and put `UIP_TCP_MSS` back to being derived. With the buffer at 1556 and the extra at 30 it lands on 1460, and the built image comes out byte for byte identical to the one that spelled 1460 out, so nothing actually moved.

The other direction doesn't hold, though. Sizing the buffer from the MSS gives 1460 + 26 + 40 = 1526, and that is 30 bytes under the frame the NIC hands us. A 1502 byte ICMP payload takes 1556 bytes of uip_buf and gets answered today, so it would stop fitting.

Which is really what the two constants are for: the buffer says what the hardware delivers, and the MSS says what a peer may put into one segment. They looked like the same number right up until the uploads started failing.

logicog:
> Reassembly puts an IP datagram back together out of fragments, and each fragment still arrives as its own frame inside the link MTU. TCP doesn't produce fragments: it sizes its segments from the MSS we advertise, so a client on a jumbo link sends one oversized segment rather than several fragments. Turning reassembly on wouldn't change what that client puts on the wire.

True, but it changes what the receiver can handle. If, say on the sender side the client sends a big data buffer using the send() function of a socket, then this gets fragmented down to fragments by the OS to fit the MTU. If our device cannot reassemble it, then that's it, the buffer content cannot be handled. Of course we can control our clients, because we control what we put into the JS that runs on the clients. So we do not need de-fragmenation, we can always work around that by splitting packets at application level. However, the MSS is now disappointingly small, so we may run into that issue at some point if we increase the size of JSON sent by the webclient. At that point de-fragmentation may be nice to have, otherwise we need to work on the JS. For now, I guess it is out of scope.

uip/uipopt.h
```diff
@@ -299,12 +299,13 @@
 /**
  * The TCP maximum segment size.
  *
- * Deliberately a step below the buffer-derived ceiling, not at it: a segment
- * that fills uip_buf to its last byte corrupts large uploads. Measured with
- * nothing but the segment size changing, 1490 fails four times out of four and
- * 1460 succeeds.
+ * Derived from the buffer as before, less the headroom the hardware turned out
```
logicog: I do not think this comment is needed at all. The formula speaks for itself.

uip/uip-conf.h
```diff
@@ -117,6 +117,19 @@ typedef unsigned short uip_stats_t;
  */
 #define UIP_CONF_BUFFER_SIZE     1556
 
+/**
+ * Bytes of the buffer kept out of the advertised MSS.
+ *
```
logicog, line 122: Anything below this line is not needed in terms of the comment. It could go into the commit comment, though.

uip/uipopt.h
DrDoof: Gone in 4385b9c.

uip/uip-conf.h
DrDoof: Trimmed to the first line, and the measurement went into the commit message instead. Also took the one above UIP_TCP_MSS out entirely, as asked.

I checked that dropping both leaves the build alone: same HEAD, same flags, byte for byte the same image.

DrDoof:
Agreed it's out of scope, and thanks for digging into it.

Worth separating one thing though, because it changes what would actually fix it. A `send()` of a big buffer over TCP doesn't reach us as IP fragments. The sending stack cuts the stream into segments at the MSS and uIP hands each of them over fine, so reassembly isn't the piece that's missing.

The worry itself is real, just in a different shape. Only the multipart upload path gathers data across packets today, in `stream_upload()`, which is how half a megabyte of firmware gets in. Something like `/cmd` reads whatever was in the packet it got handed, so a body bigger than one segment would need that endpoint to keep reading across packets the way `/upload` does. That's the same application level work you were describing, only on our side of the wire instead of in the JS.

As for the MSS being small, I think 1460 is what an ordinary client was using anyway. Both ends advertise and the smaller one wins, and a peer on a 1500 byte MTU offers 1460 itself once you take off the 40 bytes of IP and TCP header. Our old 1490 would have needed a 1530 byte IP packet, so only a jumbo link could ever reach it, and that is the link that was corrupting uploads. Nobody else loses anything.

UDP is where reassembly would earn its 700 bytes, if we ever want to take big datagrams in.

logicog:
Thanks for investigating this so thoroughly!

Issue #294, opened 2026-07-22
LACP and LAGG question


r0bb10:
Hi,

i've looked a bit at the code but it seems that LACP was never implemented, so only static LAGG with 4 trunks?

Compared to the stock firmware (i have a WAMJHJ-8125MNG from Ampcom, correctly flashed and aligned with the build with SWTGW218AS model) RTL gives the ability to have 4 groups instead of the hardcoded 2, but testing it LACPDU from one of my dev machine against RTL brings nothing up while stock firmware it responds indeed.

I've made Terra do some researches and found out usefull stuff in [RuijieNetworksCommunity/rtl837x-gsw-driver](https://github.com/RuijieNetworksCommunity/rtl837x-gsw-driver) + [JiaY-shi/rtl83xx](https://github.com/JiaY-shi/rtl83xx) and [halmartin/zyxel-gs1900-gpl](https://github.com/halmartin/zyxel-gs1900-gpl) to resolve the lacp trap2cpu, just found and confirmed acl and rma but i'm just not capable of making it work as it should, also the 8051 does not help at all.

Also my CP210x TTL seems that is not capable of reading correctly the serial port (correctly soldered and wired), any recommendation on how to make a stable connection? My first idea was to use a cheap esp32 with stream-server but it also has problems :(

Thanks for the great work so far! I'm impressed, and btw apart of LACP i've tested the rest and works as the original FW no problems so far!

logicog:
> i've looked a bit at the code but it seems that LACP was never implemented, so only static LAGG with 4 trunks?
Yes. I think the stock firmware I looked at early on did not have LACP, also anything that requires reading packets is way more difficult to implement than just setting hardware registers. So this never got implemented.

> I've made Terra do some researches
Indeed this has been part of the API, but early stock firmware did not necessarily make it available. I think it can be done now, but it will be some work.

> Also my CP210x TTL seems that is not capable of reading correctly the serial port (correctly soldered and wired), any recommendation on how to make a stable connection? My first idea was to use a cheap esp32 with stream-server but it also has problems :(
In my own experience, I made the most mistakes when soldering the GND pin. Because this is connected to a large ground plane on the PCB, it is difficult to solder and it has happened to me already a couple of times that I got a bad solder joint. Using hotter temperature and leaded solder, not the lead-free one, has helped me in the past.

r0bb10:
[...]

PR #299, opened 2026-07-25
lacp: implement IEEE 802.3ad (LACP)

DrDoof:
This adds dynamic link aggregation (IEEE 802.3ad / LACP) on top of the static LAGG we already have, for RTL8372/8373. It started from #294 and grew into a bit of a journey.

The part that took longest was simply getting LACPDUs to the CPU. The RMA trap action for `01:80:c2:00:00:02` never reaches the 8051 on this chip, and neither does an ACL trap or a redirect to the CPU port, so I went with the RMA **forward** action and a static L2 multicast entry for that address with a CPU-only member mask, one per PVID in use. The frames reach the CPU and nowhere else. Everything that was tried and measured along the way is written up in `doc/lacp.md` and `doc/link_aggregation.md`, so nobody has to repeat it.

A big thank you to @r0bb10 for testing the early version hard enough to show that plain forwarding leaks LACPDUs across the VLAN, and for the OEM register dumps that settled the acl59 question. And to @vDorst for asking for a howto in the first place; the doc exists because of that.

What's in here:
- Receive, selection and mux state machines (coupled mux, one aggregator per group), with `lacp show` for a look inside
- Per-group aggregation: each of the 4 trunk groups runs its own aggregator on a candidate set you assign with `lag <n> lacp <ports>`; the engine starts with the first such group and stops with the last
- LACPDU containment through the CPU-only FDB entries, rewritten after the startup config and on `pvid`; `lacp off` puts the slow-protocols drop default back
- Transmit periods measured against a Linux 802.3ad bond (1.00 s fast period, no expiries, churn state `none` on both ends). The partner timeouts now sit just above the standard's 3 s and 90 s; those I have not measured on the wire yet
- Web UI: Static/LACP mode per group on the LAG page, live status and a per-port table; the startup-config validator knows the per-group syntax
- The L2 page lists static multicast entries, so you can see the containment entries there
- A host-side sandbox for the state machines, now run by the test harness in CI together with the LAG hash check

Builds for all machines with the module in BANK3 (SWTGW218AS: BANK1 48094, BANK2 43618, BANK3 16348 of 49152 each), no cross-bank calls. Happy to split or trim anything that makes review easier.


vDorst:
Do you have also some resource learn and setup so we can test this?
For me this is new and have to lookin to it.
Maybe make a basic howto to use this. Like this https://github.com/logicog/RTLPlayground/blob/main/doc/igmp.md

r0bb10:
I've looked at this implementation but i dont think FORWARD is the correct way of doing this, it works but its partially broken, the isolation workaround only fixes some scenarios but i'm testing this with 2 bond0 (two different hosts each with a bond of two nics) and they leak the frames.

the poisoning and leakage of the lacpdu frames even happen with another switch (that still has stock firmware) uplinked to the custom firmwared one (using sfp).. i've tested moving one of the hosts from custom switch to stock switch, the flood problem even reaches the host on custom switch from the stock switch host masqueraded as uplinked mac.

i'm doing some more test but this approach is not safe to use.

DrDoof:
@r0bb10 You're right — thanks for testing this so thoroughly. Plain RMA-forward floods the PDUs across the VLAN, and the per-LAG isolation only masked the sibling loop, not the other leaks.

I've reworked the delivery. The RMA action stays "forward" (trap genuinely never reaches the NIC RX ring on this chip — re-verified, also with the L2_CTRL CPU port mask widened), but it is now constrained by a **static L2 multicast entry for `01:80:c2:00:00:02` with a CPU-only member mask**, one per PVID of the LACP candidate ports (lookups are IVL on this chip — a VID-0 entry is not matched). The forward lookup hits that entry instead of the VLAN flood mask, so LACPDUs reach the CPU and nothing else: no sibling leak, nothing crosses the uplink, and the whole isolation workaround is gone (user-configured port isolation is no longer overridden either).

Verified on a SWTGW218AS with a 2-NIC Linux 802.3ad bond: with the isolation removed, neither bond member sees the other's PDUs (tcpdump on the slaves shows only the switch's own LACPDUs) and no "illegal loopback" is logged; the aggregate converges from a cold boot. The steering entry is visible on the L2 page as a static entry on the CPU port (a second commit extends the listing to include multicast entries), so the containment can be checked straight from the web UI.

Could you re-test your 2-bond and uplinked-stock-switch scenarios on the updated branch?

r0bb10:
> @r0bb10 You're right — thanks for testing this so thoroughly. Plain RMA-forward floods the PDUs across the VLAN, and the per-LAG isolation only masked the sibling loop, not the other leaks.
> 
> I've reworked the delivery. The RMA action stays "forward" (trap genuinely never reaches the NIC RX ring on this chip — re-verified, also with the L2_CTRL CPU port mask widened), but it is now constrained by a **static L2 multicast entry for `01:80:c2:00:00:02` with a CPU-only member mask**, one per PVID of the LACP candidate ports (lookups are IVL on this chip — a VID-0 entry is not matched). The forward lookup hits that entry instead of the VLAN flood mask, so LACPDUs reach the CPU and nothing else: no sibling leak, nothing crosses the uplink, and the whole isolation workaround is gone (user-configured port isolation is no longer overridden either).

indeed it is a better solution than plain rma forward with isolation but still feels like a workaround, is that verified with other PVID setted up also?

> Verified on a SWTGW218AS with a 2-NIC Linux 802.3ad bond: with the isolation removed, neither bond member sees the other's PDUs (tcpdump on the slaves shows only the switch's own LACPDUs) and no "illegal loopback" is logged; the aggregate converges from a cold boot. The steering entry is visible on the L2 page as a static entry on the CPU port (a second commit extends the listing to include multicast entries), so the containment can be checked straight from the web UI.

analyzing many firmwares (v100, v1.9, 5 and 8 ports variants) all have the trap2cpu usage using acl59, oem firmware uses also others like acl62 for arp, so there must be a clean way to backport the oem usage to use trap.. rma trap is not the same as the oem acl trap, the rtl837x sdk also covers them

acl is more precise but my reverse study dont expose correctly what 59 actually does to make trap work and backport it as it should

> Could you re-test your 2-bond and uplinked-stock-switch scenarios on the updated branch?

will report once i set everything up again to test this solution, thanks

DrDoof:
> is that verified with other PVID setted up also?

Yes — and I checked it rather than assuming. I put a spare port on PVID 30 and added it as an LACP candidate; a second steering entry appeared immediately, alongside the existing one:

```
{"mac":"01:80:c2:00:00:02","vlan":"002","type":"s","port":9,...}
{"mac":"01:80:c2:00:00:02","vlan":"01e","type":"s","port":9,...}   <- VLAN 30
```

`lacp_fdb_update()` walks the PVIDs of the candidate ports and writes one entry per distinct VID, so a mixed-PVID LAG is covered. The known gap is that it runs at configuration time: changing a port's PVID while LACP is already up needs an `lacp off`/`on` to refresh. Both entries are visible on the L2 page.

> rma trap is not the same as the oem acl trap, the rtl837x sdk also covers them

You are right, and this is worth stating precisely, because I over-generalised earlier. What I verified is that the **RMA** trap never reaches the internal 8051: its destination is an *external* CPU on a physical port (`cpuTag_externalCpuPort_set`, `EXT_CPU_CTRL 0x6724`), which these boards do not have — I tried it with the L2_CTRL CPU mask widened and with the external-CPU port set to both 0xF and the NIC port; the rx counters stayed at zero in every combination.

The ACL path is a different mechanism, and the SDK makes that explicit — `acl.h` has three separate actions:

```
FILTER_ENACT_TRAP_INT_CPU
FILTER_ENACT_TRAP_EXT_CPU
FILTER_ENACT_TRAP_INT_EXT_CPU
```

`TRAP_INT_CPU` is exactly what the RMA action lacks, so your reading of rule 59 looks right and this would be the clean way: a rule matching the slow-protocols DA with an internal-CPU trap action, no static FDB entry and no flooding at all. Relevant registers, if it helps your reverse engineering: `ACL_CTRL 0x4810`, `ACL_PORT_EN 0x4818` (10-bit port mask), `ACL_TEMPLATE_CTRL 0x4820` (5 templates), `ACL_ACT_CTRL(n) 0x4848 + n*4` for n = 0..95 — 96 actions, which fits "rule 59" — and `ACL_HIT_INDICATOR 0x49C8`. The firmware currently touches none of them.

Two things would speed this up a lot, if you have a stock unit at hand: a dump of `0x4818`, `0x4820`-`0x4844` and `0x4848 + 59*4` with the stock firmware running, and whatever the rule table itself lives behind (the match fields are not in the register map I have, so they are presumably reached through the table-access port like the L2 and VLAN tables).

Until that exists, the FDB steering is at least strictly better than the flooding it replaces: nothing leaks to other ports, port isolation is no longer hijacked, and it is verifiable from the web UI. Happy to switch it over to an ACL trap the moment the rule layout is understood.

DrDoof:
Following up on the ACL trap, since you asked for it: I went digging in the SDK and then tried it on hardware. Sharing the recipe and where it currently fails, in case it saves you time.

**The rules live in the same table port as the L2 and VLAN tables** (`ITA_CTRL0 = 0x5CAC`), only the table type differs: `TB_TARGET_ACLRULE = 1`, `TB_TARGET_ACLACT = 2` — and `TB_TARGET_L2 = 4`, `TB_TARGET_CVLAN = 3`, which match what the firmware already uses, so the numbering checks out.

Layout, from `dal_rtl8373_acl.c` / `dal_rtl8373_acl.h` / `rtl8373_asicdrv.h`:

* rule address: `(type << 7) | rule`, with care bits at `type = 0` and data bits at `type = 1` — so rule 0 is address 0 (care) and 128 (data)
* rule = 5 words in `ITA_WRITE_DATA0..4` (`0x5CB8 + i*4`): words 0-3 hold the 8 fields as 16-bit halves, word 4 is `templateIdx` bits 0-2, `activePmsk` bits 11-20, **valid bit 21**
* action = 3 words, table type 2, address = index; word 1 carries `fwdAct` at bits 17-19, and **`RTL8373_ACL_FWD_INT_TRAP = 3`**, i.e. word1 = `3 << 17` = `0x00060000`
* templates: `0x4820 + i*8`, one byte per field, and note fields 0-3 go to `base + 4` while fields 4-7 go to `base`
* **the action must also be armed separately**: `ACL_ACT_CTRL(n)` at `0x4848 + n*4`, bit 5 = forward/trap action enable (`FILTER_ENACT_FWD_MASK`). The action words alone do nothing.
* enable per port: `ACL_PORT_EN 0x4818` (10-bit mask), hits visible in `ACL_HIT_INDICATOR 0x49C8`

I configured template 0 as the destination MAC (`FILTER_FIELD_RAW_DMAC_15_0/31_16/47_32` = 0/1/2), rule 0 matching `01:80:c2:00:00:02` (fields `0x0002`, `0xc200`, `0x0180` — the byte order from `acl_dal.c:380`), action 0 = internal trap, armed, enabled on the two bond ports only.

**Result: the rule matches everything, not just the BPDU group.** With the ACL enabled, LACPDU delivery to the CPU stops *and* ordinary traffic through those ports stops as well (the host behind the bond goes unreachable); disabling it restores both. I tried both care-bit polarities (all-ones and all-zeros on the MAC fields) with identical results, so the mask is not the problem.

So I cannot tell you yet whether `TRAP_INT_CPU` reaches the internal CPU — I never had a selective rule to test it with, and it would have been easy to report a false negative here. What is confirmed is that the ACL engine is live and does intercept frames.

My suspicion is that the field selectors are missing: without `igrAcl_fieldSel_set` (`FILTER_FIELD_RAW_FIELD_SELECT00 = 0x40` and friends) the rule fields are probably never populated from the frame, so every comparison trivially matches. That is where I would look next, and if your dump of a stock unit shows what rule 59 puts in `0x4820`-`0x4844` and `0x4848 + 59*4`, it would settle it quickly.

Everything above was done with `regset` over the existing CLI, no firmware changes, so it is easy to reproduce or continue from.

DrDoof:
Correction and progress on the previous comment — I found two real configuration mistakes, and my field-selector guess was wrong.

**1. `ACL_PORT_UNMATCH_PERMIT (0x481C)` defaults to DROP.** `FILTER_UNMATCH_DROP = 0`, so the moment you enable the ACL on a port, everything that does *not* match a rule is discarded. That, not the rule, is what killed all traffic in my earlier test. Setting `0x481C = 0x3ff` first makes ordinary traffic flow normally with the ACL enabled.

**2. The action has to be armed separately.** `ACL_ACT_CTRL(n)` at `0x4848 + n*4`, bit 5 (`FILTER_ENACT_FWD_MASK`), is a bitmap of which action groups are active. The three action words written through the table port do nothing on their own.

**3. My field-selector suspicion was wrong.** `PARSER_FIELD_SELTOR_CTRL(0-15)` at `0x6F58 + i*4` only backs the `FILTER_FIELD_RAW_FIELD_SELECT00..15` (0x40+) fields; the destination MAC uses the predefined raw fields 0/1/2 and needs no selector. Sorry for sending you down that path.

**Where it stands now:** with both fixes applied, the rule still does not match. The decisive test: set action 0 to redirect with an empty port mask (a match makes the frame vanish), leave the CPU-only FDB entry in place, and enable the ACL — the LACPDUs keep arriving, so they are not being intercepted. I also tried a redirect to the CPU port (`word1 = (1 << 17) | (0x200 << 21)`); no delivery either.

So the honest status is unchanged on the important question: **I still cannot tell you whether `TRAP_INT_CPU` reaches the internal CPU**, because I have never had a rule that matches. What is now certain is that the ACL engine is live and does inspect frames — the unmatch-drop behaviour proves it.

Remaining candidates for why the match fails, in the order I would try them: the template may need activating (or my write order for it is off — fields 0-3 go to `base + 4`, fields 4-7 to `base`, which is easy to get backwards), the `templateIdx`/`activePmsk` semantics in the care vs data halves of the rule word 4, and possibly `ACL_CTRL 0x4810` (TABLE_RST) needing a pulse before the table is written.

A stock dump would cut straight through this: `0x4810`, `0x4818`, `0x481C`, `0x4820`-`0x4844`, `0x4848 + 59*4`, and if you can reach it, rule 59 itself through the table port (`ITA_CTRL0 0x5CAC`, type 1, addresses 59 and 128+59).

DrDoof:
Resolved, and the answer to your question is now definitive. First a correction: two things I said earlier were wrong.

**What I got wrong.** `ACL_ACT_CTRL(n)` at `0x4848 + n*4` reads back as **`0xFF` on a factory-reset switch**, not `0`. I had been writing `0x20` to it (just bit 5, `FILTER_ENACT_FWD_MASK`), which clears the other seven bits — and with those cleared the rule never matches anything. The SDK does a read-modify-write there; I read the macro name as "set this bit" and overwrote the whole register. **The fix is to leave `ACL_ACT_CTRL` alone.** That single mistake is what made every one of my earlier experiments come back negative, so my previous claims about templates, field selectors and rule contents being ruled out are void — they were all measured through a dead rule.

I only saw this after a cold power-cycle. Every prior read showed a value I had overwritten myself.

**Positive control, so the negative result below means something.** A match-all rule with a `REDIRECT` action and `ACL_PORT_UNMATCH_PERMIT = 0x3ff` removes traffic from the port, and restores it when disabled. Since unmatched frames are explicitly permitted, the drop cannot come from a miss — the rule genuinely matches.

**The rule/care encoding is a two-mask TCAM, not value+mask.** `dal_rtl8373_acl.c` transforms the pair before it reaches the hardware:

```c
care_tmp = ~care | ~data;
data_tmp = ~care | data;
```

So per bit: must-be-1 is `(0,1)`, must-be-0 is `(1,0)`, **don't-care is `(1,1)`**, and `(0,0)` is a rule that can never match. A "match everything" rule is therefore all-ones in *both* words, not all-zeros. `rule_info` is 22 bits wide — writing `0xFFFFFFFF` reads back as `0x003FFFFF`.

**Answer on `TRAP_INT_CPU`: it intercepts, but it does not deliver.** With a rule on DMAC `01:80:c2:00:00:02` and `fwdAct = RTL8373_ACL_FWD_INT_TRAP (3)`, the LACP rx counters freeze while the ACL is enabled and resume the moment it is disabled — so the frames are being taken out of the normal path, but they never reach the 8051. I also tried `REDIRECT` with the port mask set to the CPU port (`word1 = 1<<17 | 0x200<<21`); same outcome — intercepted, not delivered. This lines up with the RMA trap on this board, whose destination is an *external* CPU (`cpuTag_externalCpuPort_set`, `EXT_CPU_CTRL 0x6724`).

**So the FDB steering stays.** The ACL engine works and matches correctly, but no delivery path I can find brings the frames to the internal CPU, and shipping a mechanism that doesn't deliver would be worse than the static L2 entry we have. If you know whether the action port mask uses physical port numbering, or whether redirect-to-CPU needs additional cpuTag setup, that would be the missing piece — but it is no longer blocking this PR.

For anyone continuing: the working recipe is templates at `0x4820 + i*8` (fields 0-3 go to `base+4`, 4-7 to `base`; they are **zero after reset**, you must fill them), rules through the table port `ITA_CTRL0 0x5CAC` with `TBL_ADDR<<16 | TYPE<<8 | ACT<<1 | EXEC<<0`, care at address `(0<<7)|idx` and data at `(1<<7)|idx`, valid at bit 21 of data word 4, write order invalidate → care → data+valid, `ACL_PORT_UNMATCH_PERMIT 0x481C` set to `0x3ff` first because it defaults to drop, and `ACL_ACT_CTRL` left untouched. The rule table holds uninitialised garbage after reset, but with the valid bit clear.

r0bb10:
You can give this to the agent to study, i did it quickly with just Terra and live OEM fw.

I told it to verify what changed on the registers when a trunk1 of 2 ports (other side 2x 2.5gb nics in LACP layer2+3) was disabled and enabled, this is the result.

## OEM RTL8373 LACP A/B register delta — `lacp_port_enable` 0x00C3 (on) vs 0x0000 (off)

Source: `oem-lacp-on.log` vs `oem-lacp-off.log`, both dumps are 513 lines.

[oem-lacp-off.log](https://github.com/user-attachments/files/30750166/oem-lacp-off.log)
[oem-lacp-on.log](https://github.com/user-attachments/files/30750167/oem-lacp-on.log)


## Result: only 2 registers differ, both live NIC/CPU RX queue words

| Register | Name (per `rtl837x_regs.h`) | LACP on | LACP off | Delta |
|---|---|---|---|---|
| `0x7878` | NIC RX current packet | `0x0204` | `0x0083` | RX queue active |
| `0x787c` | `RTL837X_REG_CPU_RX_CURR_PKT` | `0x0228` | `0x0083` | CPU RX queue active |

Nothing else differs. No ACL, no RMA, no action-control, no global changes.

## Configuration is bit-identical (LACP on vs off)

| Register | Function | Value |
|---|---|---|
| `0x4818` | ACL port enable | `0x01ff` |
| `0x481c` | ACL unmatched frame action | `0x01ff` |
| `0x4ed4` | RMA2 (RMA forwarding control) | `0x00000020` |
| `0x4934` | ACL Rule 59 action control | `0x00000000` |
| `0x7854` | NIC buffer size TX | `0x2` |
| `0x7858` | NIC reserved | `0x0` |
| `0x785c` | RX_CTRL | `0x400201b` |
| `0x7860` | TX_CTRL | `0x1` |

## Interpretation

1. The OEM enables LACP without writing a single configuration register. There is **no** config-detectable "trap" state: Rule 59 and the RMA2 block are untouched when LACP is toggled on/off.
2. The only observable difference is that the **NIC RX and CPU RX current-packet pointers** are live while LACP is on — i.e. LACPDUs are delivered over the normal CPU-port RX queue path (the same `0x7878`/`0x787c` polling loop the RTLPlayground firmware already uses), **not** via a dedicated trapped/isolated path.
3. Consequence: a static or FDB/CPU-forward of `01:80:c2:00:00:02` into the normal NIC RX queue is functionally indistinguishable from the OEM mechanism in register space, the OEM does not reconfigure ACL slot 59 when LACP is active.

DrDoof:
This is the measurement that settles the acl59 question, thank you — an A/B against live OEM is worth more than everything I inferred from static analysis. And there is one row in your table that I think is the most interesting thing in the whole dump.

## `0x4ed4 = 0x00000020` — the OEM has the LACP group set to *drop*

That register is `RMA2_CONF`, the reserved-multicast entry for `01:80:c2:00:00:02` (`0x4ecc + 2*4`), and the action field is bits [5:4]: `0x00` forward, `0x20` drop. So in your dump the OEM has that group on **drop**, identically with LACP on and off — and LACP still works.

That rules out RMA-forward as the OEM's delivery mechanism, which is worth stating explicitly because it is what this branch currently relies on.

I tested the converse here, on the current branch, this evening:

| `RMA2` (`:02`) | LACP rx counters |
|---|---|
| forward | advancing |
| **drop** | **frozen, zero increments over 10 s** |
| forward (restored) | advancing again |

So on our side the RMA drop pre-empts the L2 lookup entirely: the static CPU-only multicast entry does **not** deliver on its own. Same register, same chip family, opposite outcome. Something on the OEM gets those frames to the CPU while the reserved-multicast action says drop, and I cannot see what from here.

**The ask, if you are up for one more dump:** the L2/FDB table on the OEM unit while the trunk is up, specifically any entry for `01:80:c2:00:00:02` — its member mask and, more importantly, its type/flag bits. If the OEM has a static entry with some attribute we are not setting (something that takes priority over the RMA action), that is the missing piece. It would let this branch drop RMA-forward altogether, which removes the flooding failure mode you originally reported rather than merely constraining it.

## On acl59, your reading is confirmed from the other side

`ACL_ACT_CTRL` for rule 59 reads `0x00000000` in your dump — that slot is present but **disarmed**. The action words in the table do nothing on their own; the action group has to be armed in `ACL_ACT_CTRL(n)` at `0x4848 + n*4`. I lost a dozen rounds to this from the opposite direction: that register reads `0xFF` on a factory-reset switch, and I had been writing `0x20` to it (just the FWD bit), clearing the other seven and killing every rule I wrote. So `0x00` there means inert, which is exactly your "no config-detectable trap state" conclusion, reached independently.

Your `0x4818`/`0x481c` at `0x01ff` fit the same picture: the ACL engine enabled on all nine user ports (bit 9, the CPU port, is not set) with unmatched frames permitted. Note `ACL_PORT_UNMATCH_PERMIT` reads `0x00000000` after a factory reset and `0` means *drop*, so `0x01ff` is a deliberate write, not a leftover — an engine switched on with nothing armed.

For completeness on the trap itself: I did get ACL rules matching here once the arming was right. The other trip hazard is the encoding — the SDK transforms the pair as `care = ~c | ~d`, `data = ~c | d`, so it is a two-mask TCAM where "don't care" is `(1,1)` and `(0,0)` is a rule that can never match; a match-all rule is all-ones in *both* words, not all-zeros. With a matching rule on `01:80:c2:00:00:02`, `FWD_INT_TRAP` intercepts the frames — the rx counters freeze while it is enabled and resume when it is disabled — but they never reach the 8051. `REDIRECT` aimed at the CPU port behaved the same, though I would not lean on that one: I have not verified whether the action's port mask uses physical or logical numbering, and a wrong bit there would look identical.

Still very interested in the two-bond and stock-switch-uplink retest when you get to it — those were the scenarios that broke the previous approach, and they are the ones that matter for whether this is actually fixed.

DrDoof:
Rebased onto current main. The only conflict was in send_l2(), where #340 rewrote the same block I had changed to list multicast entries.

I kept both sides rather than picking one. The buffer bound and the separator handling from #340 stay as they are, and so does taking the VLAN from the register that had already been read. The entry test keeps the L3 filter on top of that, so an IP entry is still skipped instead of printing addresses where a MAC belongs.

Builds for both machines on sdcc 4.5.0. Not run on hardware since the rebase.

DrDoof:
@vDorst sorry this took so long — you asked on 30 July and five weeks is not a reasonable turnaround.

`doc/lacp.md` is in, written to the shape of `doc/igmp.md`: what the protocol does here, how it is configured from the console and the web UI, and a section on trying it against a Linux 802.3ad bond with what to look for on both ends.

The long part is how the frames reach the CPU, because the obvious route is the wrong one on this chip and anyone reading the code would trip over it. The RMA trap action aims at an *external* CPU these boards do not have, so it takes the frames out of the forwarding path and nothing arrives anywhere; forward is what actually reaches the 8051, and the static CPU-only L2 entry is what stops forward from also flooding the group to every other port. That is what @r0bb10 found the hard way, and it now lives in the tree instead of in this thread.

Your comment on #377 shaped the section on what has to match. I took the standard's line, full duplex and identical speed, and kept it separate from the wider test, which is that anything changing how a frame is treated has to match because the hash picks the member per frame. Your 2.5G-and-SFP example is worth having in mind precisely because it reads as reasonable and the standard still rules it out, so the doc gives the reason rather than only the rule. It also says plainly that none of it is enforced today.

One warning about the branch itself, since it is behind main. I merged current main locally and all twelve conflicts resolve, but the result does not build, and not for the reason I expected: the image no longer fits the two code banks. BANK1 comes out 235 bytes over its 0xC000 window and BANK2 562. A third bank is the next step and I would rather do that carefully than quickly, so the doc lands on its own ahead of it.

DrDoof:
Merged and building again, so this is out of conflict.

The image no longer fits two code banks, so there is a third now. That is one linker flag — `crtbank.asm` already masks five bits of bank number and `imagebuilder` derives the count from the file size — plus moving `rtl837x_lacp.c` out of BANK2 and `rtl837x_igmp.c` out of BANK1. Those two are the self-contained ones, and it buys far more room than either bank was short of: BANK1 has 1320 bytes free now, BANK2 3161, BANK3 43874.

The check is the part worth stating, because this is the failure mode that does not announce itself. An `lcall` carries sixteen bits and no bank number, so a plain call across a bank boundary lands at the same low address in the *caller's* bank and runs whatever happens to be there. It builds clean and bricks at runtime. I walked every `lcall` in the generated assembly against the symbol's bank from the map: 2904 calls, none crossing.

A clean result from a check that cannot fail is worth nothing, so I pointed it at a violation I already knew about first. Moving `syslog.c` to BANK3 makes its call to the non-banked `print_ip()` in BANK2 show up:

```
syslog.asm:279  [BANK3] -> _print_ip  [BANK2, cmd_parser]
```

It found it, so the zero above means the check works rather than that it is blind.

Also confirmed the new bank's bytes really are in the `.bin` at `0x1c000` rather than being dropped on the way out, and `machine_check` builds every machine.

Not run on hardware since the merge.

DrDoof:
Hardware tested after the merge, and it turned up the reason the bond never quite looked right.

LACP converges and the aggregate works, but against a Linux bond the churn state sat on `monitoring` forever instead of settling to `none`. Traffic went over both links the whole time, which is presumably why nobody chased it.

Reading the frames explained it. We were transmitting every 2 to 3 s to a partner that had asked for the fast rate, and the fast rate comes with a 3 s expiry. Every time a gap crossed the line the partner's receive machine timed us out and recovered: 13 of its 93 LACPDUs came back with `EXPIRED` set and our SYNC cleared. We record a received actor state verbatim and send it straight back, so the flap bounced between both ends and neither could settle.

The transmit period is the one timer here that is not ours to choose freely. Our own timeouts only decide how patient we are with a partner that has gone quiet; the transmit period decides when the partner gives up on *us*. A timer unit measured out at 1/50 s, so fast periodic is 50 units now and slow 1500. The slow one needed it too — at 7680 units it came to 94 s against a 90 s long timeout, already past the line whenever the jitter went the wrong way.

```
interval between our LACPDUs      1.00 s in all 116 intervals, none over 3 s
partner LACPDUs carrying EXPIRED  0 of 115        (was 13 of 93)
churn state, both ends            none            (was monitoring, indefinitely)
```

Reading a frame by hand also turned up the reserved fields going out with whatever the buffer held from the previous frame — three bytes of our own MAC, in most samples. Zeroed now.

The rest of what I checked, on a SWTGW218AS against a two-NIC 802.3ad bond:

- converges from a cold boot; the partner's MAC is elected as the aggregator and both ports land in the hardware trunk
- it fixed a bond that was already broken. The host had its two slaves sitting in separate aggregators before the flash and in one after
- containment holds: each slave sees our LACPDUs and its own, and never the other slave's, over 30 frames apiece
- nothing leaks past the group either — no slow-protocol frames on two hosts elsewhere on the network
- traffic distributes. Six parallel flows split two and four across the members
- pulling a member down costs 5% of a 120-packet flood, and that at the transitions rather than the half you would lose if the ASIC kept hashing onto the dead link. It rejoins by itself
- 15 of 15 host sandbox scenarios pass

What I have not tested is the pair of scenarios you raised, @r0bb10: two bonds on one switch, and the uplink to a stock switch. I only have one bond here, so all I can say is that the leak does not happen in this topology — not that yours is fixed. The offer to re-test stands whenever you have the setup back up.

DrDoof:
Rebased after the recent merges to main; the module moved to BANK3.

DrDoof:
Rebased after #397 and #424.

DrDoof:
Rebased on current main and pushed five more commits. I sat down with the branch once more before asking anyone to spend time on it, and I am glad I did, because two real bugs were hiding in plain sight. Sorry for the extra round; better now than after the merge.

- Once a partner was elected, the aggregator held on to it as long as *any* port was still hearing *someone*, even a different system. Move a LAG to a new host one cable at a time and it just sits there, nothing aggregated, until you toggle LACP. Sandbox T16 plays it through; the fix is one line in the release loop.
- The CPU-only steering entries only covered the LACP ports' PVIDs, but the forward action is not that picky, so a LACPDU arriving on some other port in another VLAN was still flooded. @r0bb10, this is a cousin of what you found; entries now cover every port's PVID and are refreshed after the startup config and on `pvid`.
- `lacp on` is gone. It quietly replaced group 1's ports with all of them, and the doc told people to use it. My mistake, on both counts.
- The partner timeouts were still on the old timer scale, 15 s and close to 8 minutes. They now sit just above the standard's 3 s and 90 s. Not measured on the wire yet; that is my next evening.
- The sandbox and the LAG hash check now run from `make -C test`, which promptly showed the hash check had stopped compiling after `LAG_HASH_RESET` landed. Fixed, and a good argument for wiring tests in.

All machines build, no cross-bank calls, harness green.

DrDoof:
Rebased on main, which now has #482 and #483. Those two change how long a pass of the main loop takes, and the LACP timers counted passes, so under load they ran fast or slow with the traffic; the 4 s partner timeout is the one that matters. The timers step by the system tick now, the same way #488 does it for STP, and the branch carries the two commits of #488 until that one is in. The sandbox advances the tick counter instead of calling the timers, so its scenarios keep their timing.

DrDoof:
Three more commits, from one more read of the branch with the sandbox at hand. Sorry for yet another round; each one is small and each comes with a sandbox check that fails on the code before it.

- A group that already had a static trunk kept it when put under LACP, so the ASIC hashed onto ports the partner had not aggregated yet, and a reboot brought the group up differently. The LAG page's Static -> LACP switch sends exactly that sequence. The trunk is cleared now.
- A port named in a second LACP group stayed in the first group's candidate set; taking the second group out then left the port orphaned. It moves between groups now.
- lacp_fdb_set() wrote the steering entry without resetting the table access mode in TBL_DATA_0, which the L2 page leaves on "next address"; every other L2 writer clears it first. Found by reading, not reproduced on hardware, so I would not call it verified.

SWTGW218AS: xdata +1 B, stack 149 -> 147 B, no cross-bank calls; the harness passes.

DrDoof:
Two more small commits, on top of the squash.

- `lacp off` (or taking a port out of a group) left the port's last partner in place: `lacp show` kept reporting the partner system, state and counters as if the port still heard it. `lacp_port_release()` clears them now.
- Re-applying the same candidate set to a group tore the trunk down and renegotiated it, one to two seconds without the aggregate. The LAG page resends the `lag N lacp ...` line on every Apply, also when only the hash changed, so this happened on every Apply. Same ports now only refresh the steering entries.
- `lacp_in()` now checks the frame length against the PDU it reads and the slow-protocols ethertype, so a runt on 01:80:c2:00:00:02 cannot record stale buffer bytes as a partner.
- LAG page: with mode LACP and no port checked, Apply sent `lag N lacp`, which the firmware takes as "no candidates" but the config validator accepted as a line. It sends `lag N lacp off` now, and the validator requires at least one port after `lacp`.
- Timing comments next to the transmit periods still described the loop-pass era; the timers step by the tick now.

Sandbox T20-T22 cover the three behaviour changes and fail on the code before this. Checked on SWTGW218AS: re-applying `lag 1 lacp 3 4` no longer calls `port_lag_members_set`, and `lacp show` after `lacp off` shows partner 00 / psys 000000000000. Code +164 B, xdata and stack unchanged, no cross-bank calls.

PR #301, opened 2026-08-01
Add FG-4GT-2SX_V2.0

zytstudio:
Add FG-4GT-2SX_V2.0, used in RY-4GT-2SX.


vDorst:
Thanks for the PR, it is fine.
Some small changes I would like to see.
1. Add led color as comment in the `machine.c`. So we know which led-set controlled which color.
2. Edit the photos so the non-PCB parts are reduced to minimum.

zytstudio:
> Thanks for the PR, it is fine. Some small changes I would like to see.
> 
>     1. Add led color as comment in the `machine.c`. So we know which led-set controlled which color.
> 
>     2. Edit the photos so the non-PCB parts are reduced to minimum.

Done, thanks!

vDorst:
Thanks

PR #302, opened 2026-08-04
stp: contain BPDUs to the CPU while STP runs

DrDoof:
With STP enabled the switch participates in the spanning tree, so BPDUs must be consumed by the bridge, not relayed — yet `01:80:C2:00:00:00` was flooded across the VLAN like any multicast, so every BPDU leaked to all other ports. This is the same defect class as the LACPDU flood being addressed in #299, applied to the BPDU group.

**Mechanism.** The RMA "trap" action cannot deliver to the internal NIC on this hardware (its trap destination is an external CPU on a physical port — this board has none), so the working delivery is RMA *forward* constrained by a **static L2 multicast entry with a CPU-only member mask**: the forward lookup hits the entry's portmask instead of the VLAN flood mask. Hardware-verified in both directions on a SWTGW218AS (a mask without the CPU bit stops delivery, a CPU-only mask delivers with no port egress). The first commit adds the shared `port_l2mc_set()` helper; #299 adopts it for LACPDUs as a follow-up once either lands.

**Per-VLAN entries.** BPDUs can arrive VLAN-tagged and classify into the tag's VID (lookups are IVL), so `stp on` writes one CPU-only entry per VLAN present in the VLAN table, plus each port's PVID for the untagged case.

**`stp off` restores transparency.** The entries are retargeted to all ports + CPU, restoring the previous flood behaviour: with STP disabled an unmanaged-style switch is expected to be transparent to BPDUs so a surrounding spanning tree can span through it (topologies genuinely rely on this), and dropping them instead would partition that tree.

Known pre-existing interaction (unchanged here, noted in the commit): with STP enabled all ports start out blocking, which also stops egress of CPU-originated LACPDUs, so an active LACP aggregate drops until the ports reach forwarding.

Tested on SWTGW218AS: entries created per VLAN on `stp on` and retargeted on `stp off` (visible in the L2 page with #299's listing fix), enable/disable fully reversible across cold boots.

### DIFF
--- a/rtl837x_port.c
+++ b/rtl837x_port.c
@@ -395,6 +395,48 @@ void port_l2_learned(void) __banked
 
 
 /*
+ * Static L2 multicast entry for the link-local group 01:80:C2:00:00:<mac_last>
+ * in VLAN `vid`, with member portmask `pmask` (bit 9 = CPU port).
+ *
+ * Slow-protocol frames (LACP, STP BPDUs) must reach the CPU without being
+ * flooded to other ports. The RMA "trap" action cannot deliver to the
+ * internal NIC on this hardware (its destination is an external CPU on a
+ * physical port), so the protocol modules keep the RMA action at "forward"
+ * and constrain the egress with this entry instead: the lookup hits the
+ * entry's portmask rather than the VLAN flood mask (hardware-verified with
+ * both the CPU bit cleared - delivery stops - and CPU-only - no egress).
+ *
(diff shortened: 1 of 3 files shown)


vDorst:
Maybe better to close this PR and append it to #303.
Or close both and start a new PR with a bit smaller PR text.

DrDoof:
Closing as suggested. Both commits here are the first two of the STP branch, so
there is nothing to carry over. The whole thing continues in #314 with a shorter
text, rebased onto main.

PR #303, opened 2026-08-04
stp: complete the Spanning Tree implementation

DrDoof:
Completes the STP implementation: the engine currently sets every port to blocking and never counts the listen period down, so `stp on` leaves the switch blocking forever, and the BPDUs it emits are malformed (version/type mismatch, permanent Topology Change flags, a CPU tag that leaks onto the wire as 0x8899).

**Depends on #302** (the BPDU containment and the `port_l2mc_set()` helper it adds). Shares one commit with #299 (`common: define the RTL frame-tag flag bits`) — whichever of the two merges second drops it as empty on rebase.

### What it does

* **Ports actually leave blocking**, and the timers count in real seconds. `stp_timers()` runs at 50 Hz (the main loop idles on the 200 Hz system tick, we are called every fourth pass) - measured on hardware, with hello configured to 2 s the BPDUs left the port 2.560 s apart before the constant was corrected.
* **BPDUs reach the wire correctly**: the CPU tag flags go through `HTONS()` (a raw `0x0020` lands as `0x2000`/EFID, the ASIC then fails to parse the tag and floods the frame with the `0x8899` header still on it), and the management VLAN is suppressed per frame so the tag stays where the ASIC parses it. `RTL_TAG_KEEP` is deliberately not set: hardware-verified that KEEP on an LLC/802.3 frame makes the ASIC drop it, while it is fine on ethertype frames.
* **Correct RST BPDU format**: the 36-byte body including the version-1 length octet, real flag bits derived from the port state instead of a hardcoded `TC|TCA`, 1-based port identifiers, and the message age relayed (incremented per hop) rather than always advertised as zero. Legacy Config BPDUs keep their 35-byte body.
* **RX**: accepts RST, legacy Config and TCN BPDUs (a TCN is acknowledged with TCA on the ingress port), and validates the frame length before touching the body, so truncated or fuzzed BPDUs are dropped rather than parsed.
* **Only designated ports announce periodically** — echoing our own root information back out of the root port made us look like a competing designated bridge on that segment.
* **Topology changes** flush the affected port's dynamic entries (new bounded single-port flush; the existing whole-table helper polls unbounded and must not run from the tick) and set the TC flag for max age + forward delay. Edge ports are exempt.
* **Configuration** per 802.1D-2004/802.1w: bridge priority, hello/max age/forward delay, force-version, tx hold count; per port enable, admin/auto edge, path cost, priority, BPDU guard, root guard, BPDU filter, p2p — CLI, web page (`/stp.json` + a Spanning Tree page) and startup-config persistence.
* **Management failsafe** (commit-confirm): while STP is on, any HTTP request re-arms a countdown; if management goes silent for the configured number of seconds, STP disables itself and restores forwarding. Enabling STP on a switch whose own management rides a blockable port is otherwise a one-way trip to the power switch — this is not hypothetical, it happened repeatedly during development.
* Bugs found on the way, in separate commits: the "do not block the CPU port" mask covered ports 8 **and** 9, so a real front port was pinned to forwarding and could never be blocked; `atoi_byte()` wrapped silently, so `stp failsafe 300` configured 44 seconds; a staged Topology Change Acknowledgment survived a BPDU that was never sent and surfaced on an unrelated port; and the tick rate above.
* The module moves to code bank 2 — the always-mapped common area had 349 bytes left.

### Testing

On a SWTGW218AS, with the switch in a live network (RSTP root elsewhere):

* BPDUs on the wire, captured on a directly attached neighbour: `802.3, length 39, LLC dsap STP ssap STP, STP 802.1w, Rapid STP, Flags [Learn, Forward], bridge-id 8000.06:05:16:1e:f9:24.8008, length 36` — correct RST framing, real flags, no `0x8899` leak.
* RX and the election verified by injecting a superior BPDU from a directly attached host: the switch adopted the announced root (priority and MAC) and set the correct root port.
* Ports leave blocking after the listen period; edge ports go forwarding immediately; the CPU-only containment entries from #302 are written per VLAN on enable and retargeted on disable.
* The failsafe was exercised repeatedly and recovered management every time.

Since then, with the BPDU filter set on the uplink port (the neighbouring switch here reacts to BPDUs by cutting the link, `stp port N filter on` keeps it quiet), the rest was verified too:

* **Hello timing**: with `hello 2`, BPDUs left the port 2.000 s apart (2.560 s before the tick-rate fix), drift under 0.002 %.
* **Election and path cost**: an injected BPDU announcing root `1000.aabbccddeeff` with cost 4 gave `rootPort` = the ingress port and cost `0x4e24` = 20004 = 4 + the port's 20000.
* **Port roles**: measured on both bond members at the same time, the root port emitted 1 BPDU while the designated port emitted 4 — the root port goes quiet as it should.
* **Max age**: with the injection stopped, the root aged out and the switch reclaimed the root role.
* **Topology change**: taking a non-edge port down and up produced blocking → 15 s → forwarding, one topology-change increment, and the L2 entries of *that port* flushed (3 → 1) while the neighbouring port kept its own (3 → 4).
* **Malformed input**: TCN, legacy Config, a truncated RST and two garbage frames — parsed or rejected, switch alive, management and forwarding unaffected.
* An edge-port bug found this way is fixed in the branch: `stp port N edge off` cleared the admin flags but not the operational one, so the port stayed an edge port until the next `stp off`/`stp on`.

Still open: a long soak, and persistence of the per-port settings across a reboot. Reviews and testing in other topologies are very welcome.


doc/stp.md
```diff
@@ -0,0 +1,154 @@
...
+```
+
+## Hardware background
+
+BPDUs are addressed to `01:80:C2:00:00:00`, a reserved link-local group. The
+ASIC's Reserved-Multicast action for that address decides what happens to the
+frame. The "trap to CPU" action does **not** reach the internal 8051 on this
+hardware — its trap destination is an external CPU attached to a physical port
+(`cpuTag_externalCpuPort_set` in the vendor SDK), which these boards do not
+have. Delivery therefore uses the *forward* action, constrained to the CPU port
```
vDorst: Is this tested or is it a assumption?
Because 8051 interface, AFAIK is just a port on the internal switch hardware like every other port.

DrDoof: Tested, not assumed — and you are right about the CPU being an ordinary port, which is exactly why the sentence is worded around *trap* specifically rather than around delivery in general.

Forwarding to the CPU port works normally. That is what this PR relies on: the static L2 multicast entry with a CPU-only member mask delivers BPDUs to the 8051, which is only possible because port 9 behaves like any other member of a forwarding mask. If it did not, none of this would work.

What does not work is the *trap* action, and that is a separate mechanism from forwarding. Three independent measurements, on a SWTGW218AS:

* RMA: setting `RMA_ACT` for `01:80:C2:00:00:00` to trap gives zero frames at the 8051. I widened `CPU_PMSK` and tried the external-CPU destination at both `0xF` and `9` (`EXT_CPU_CTRL 0x6724`) in case it was a routing question — still zero. With the same setup and the *forward* action plus the L2 entry, frames arrive.
* ACL, measured this week while chasing an unrelated question in #299: a rule matching `01:80:C2:00:00:02` with `fwdAct = RTL8373_ACL_FWD_INT_TRAP` freezes the LACP rx counters while the ACL is enabled and they resume the instant it is disabled. So the frames are being pulled out of the normal path and are not arriving at the CPU.
* Same ACL rule with `REDIRECT` and the port mask set to the CPU port: identical result, intercepted but not delivered. I would not lean on this third one too hard — I have not verified whether the action's port mask uses physical or logical port numbering, so a wrong bit there would look the same.

The first two are what the sentence rests on. The common thread is `cpuTag_externalCpuPort_set` in the vendor SDK: the trap destination is a physical port meant for an external CPU, which these boards do not populate.

If the wording reads as a claim about the CPU interface in general, that is my fault and I am happy to narrow it — something like "the trap action's destination is an external CPU port, so trapped frames never reach the internal 8051; forwarding to the CPU port works normally and is what we use". Say the word and I will amend the commit.

DrDoof:
Two more fixes pushed, both found by testing this on hardware rather than by reading the code.

**`stp: react to a port losing carrier`** — the state machine never looked at link state. A non-edge port whose link went down stayed in forwarding, kept being announced, and kept its learned entries; the per-port flush added earlier existed but nothing ever triggered it. Losing a link is the most ordinary topology change there is, so this was a real hole. The carrier bitmap (`LINKS_STS`) is now sampled once per second, alongside the tx-budget refill — the 50 Hz tick has no business doing register reads. On carrier loss the port goes back to blocking and runs the normal topology-change path; on carrier return it re-runs the listen period instead of forwarding straight away (the segment may have been rewired) and drops its operational edge flag so an auto-edged port has to earn it again. `stp_setup()` seeds the bitmap from hardware so enabling STP does not report every already-down port as a fresh change.

Measured on a SWTGW218AS: a non-edge port went forwarding → `tc=0000`, link down → **blocking, `tc=0001`**, link back → blocking for the listen period → **forwarding, `tc=0002`**. An edge port goes to blocking too but produces no topology change, which is what 802.1w asks for.

**`stp: accept BPDUs with a protocol version above 2`** — we only recognised RST BPDUs at version exactly 2, which silently drops every MST BPDU. 802.1s uses version 3 with type 2 and a prefix deliberately identical to RST, and 802.1D-2004 14.4 requires a bridge to accept version 2 *or greater* and treat it as RST. Verified by injecting a v3/type-2 and a v4/type-2 BPDU on a bond port: the switch now adopts the advertised root (`weRoot` goes 1 → 0), where before both were discarded. The receive path already length-checks before touching the body and only reads fields common to both formats, so the longer MST body needs no further handling.

Worth noting the asymmetry that caused this: the Protocol Identifier must be exactly zero — it is a sanity check — while the version field is an extension point that has to tolerate the future. Writing both as `==` looks consistent and is wrong in half the cases.

**Also exercised, no changes needed:** malformed BPDUs (truncated to 10 and to 1 byte, 40 bytes of trailing junk, all-flags-set, bad protocol id, 35 bytes of garbage, legacy Config and TCN) — nothing hung the 8051 and the header validation rejects what it should. Failsafe still trips correctly after the configured silence. LACP unaffected throughout (`aggValid` stays 1, counters keep moving).

**Not tested, environment rather than code:** port roles against an external root, and the unplug/replug re-election. The uplink here carries `bpdufilter` because a neighbouring switch cuts the link when it sees our BPDUs, so this bench cannot hold a foreign root long enough to exercise those paths. If you have a setup that can, those are the two areas I would look at first.

doc/stp.md
vDorst: Based on this sentence `freezes the LACP rx counters`, I think you already checked the counters/port-statistics/mib to see if a packet is send to the 8051 interface?

Are rtl-SDK `rtl8373_aleLatch_set()` and `rtl8373_aleLatch_get()` related to this topic? It is missing documentation. But it seems to able to fetch a packet data with a lot of flags.

I also wonder did you some experiments to see if it works when you are using an external port as "CPU-PORT"?

Are you also making use of the original managed firmware and reverse engineer it, too see how they implemented the `stp`-functionality?

Are you reprogramming the device every time make a code change? Or are you using the `reg{get,set}` or the I2C-SlaveInterface to make quick change?

DrDoof:
Soak on the build carrying both of the above fixes: **59 samples over an hour, zero anomalies**. Port states and roles unchanged throughout, we stayed root, the bond partner remained reachable on every sample, and no sample failed to answer.

The number I was actually watching is the topology-change counter, which stayed at `0003` for the full hour. Now that STP reacts to carrier state, any flicker of the link-status bit on any of the nine ports — or a race between the once-per-second read and the hardware updating it — would have shown up as a stream of spurious topology changes. It did not, on ports carrying live traffic. The functional test could only show that a real link loss is detected; this is the other half, that nothing is detected when nothing happens.

The management failsafe closed the run on its own: when the monitor stopped polling, the switch disabled STP after the configured silence and restored forwarding on all nine ports, with the bond unaffected. That is the intended degradation and it is worth noting it was not staged - the monitor simply ended.

doc/stp.md
DrDoof: Good questions, and two of them point at things I have not done. Taking them in order.

**Counters.** What I watched is the firmware's own LACP receive path — the per-port `rx` counters on `/lacp.json`, which only advance when a frame actually reaches the LACP state machine on the 8051. With the ACL rule matching, those freeze while the ACL is enabled and resume the moment it is disabled, so the frames are definitely being taken out of the normal path and definitely not arriving at the CPU. What I did **not** do is read the MIB/port statistics of the CPU port itself, which would distinguish "never sent towards the 8051" from "sent but dropped before the RX ring". That is a better instrument than mine and I will use it.

**`rtl8373_aleLatch_set/get()`** — no, I have not touched them. I did notice the register block behind what I assume is the same thing (`ACL_LATCH_TRIGGER` 0x4BE0, `ACL_LATCH_ADDR` 0x4BE4, `ACL_LATCH_VAL_L/H` 0x4BE8/0x4BEC) and put it aside because it looked like a debug capture path rather than a delivery path. If it can fetch packet data with flags, it is exactly the instrument I was missing while trying to tell interception from delivery. Thanks — that is the most useful pointer in your list.

**External port as CPU port** — yes. The RMA trap destination is set by `cpuTag_externalCpuPort_set` in the SDK, `EXT_CPU_CTRL` at 0x6724; I tried it pointed at `0xF` and at `9`, with `CPU_PMSK` widened, and got zero frames either way. That is what the sentence in `doc/stp.md` rests on, together with the ACL result below.

**Reverse engineering the OEM firmware** — yes, and this is where it gets interesting. I have a 2 MB dump of the stock image for this exact board (`up-n-atom/SWTG118AS`, the "Horaco SWTGW218AS" entry) and disassembled it with radare2's 8051 backend. Register access there goes through the same SFR mechanism RTLPlayground uses (address raw into 0xa2/0xa3, `0xa0` = go, poll `0xa1`, data in 0xa4-0xa7), so call sites are findable by scanning for the address loaded into r6/r7 — note the operand order is `7F <low> 7E <high>`, which cost me a while.

In **this** image the ACL block is only ever initialised: `ACL_CTRL 0x4810` = 1 (table reset), `ACL_PORT_EN 0x4818`, `ACL_PORT_UNMATCH_PERMIT 0x481C`. There are no writes to the templates (0x4820-0x4844), none to `ACL_ACT_CTRL` (0x4848+), and no rules — I checked by scanning on the high address byte alone, so computed addresses are covered too. @r0bb10 reports that other variants (v100, v1.9, the 5- and 8-port boards) do use acl59 for trap-to-CPU and acl62 for ARP, which I have no reason to doubt; it just means the feature is version-dependent and this particular image is not one of the ones that uses it.

Separately, I did get ACL rules working on this chip, which took embarrassingly long: **`ACL_ACT_CTRL(n)` reads back as `0xFF` on a factory-reset switch**, and I had been writing `0x20` to it (just the FWD bit), clearing the other seven and killing every rule. Also worth knowing, the rule/care pair is a two-mask TCAM — the SDK transforms it as `care=~c|~d, data=~c|d`, so "don't care" is `(1,1)` and `(0,0)` is a rule that can never match; a match-all rule is all-ones in *both* words. With that fixed, a rule on `01:80:c2:00:00:02` does match and does intercept, but `FWD_INT_TRAP` still does not deliver to the 8051, and neither does `REDIRECT` with the port mask set to the CPU port. That is the piece your `aleLatch` question might actually settle.

**Workflow** — no reflashing for experiments. Everything above was done live through `regget`/`regset` over the existing CLI. `regget` writes to the syslog rather than the HTTP response, so `syslog ip <host>` + `syslog on` and a `tcpdump -A udp port 514` on a directly attached machine gives a usable register read-back channel. I only build and flash when actual code changes, and then always a full image from the integration branch plus a cold power cycle.

DrDoof: Amended, as offered — `fd55667`.

The paragraph now says what is actually broken rather than making a claim about the CPU interface as a whole. It states up front that forwarding to the CPU port works normally and that the 8051 is an ordinary member of a forwarding mask (this implementation depends on exactly that), then narrows the failure to the trap action and records the measurements behind it: zero frames with the RMA action set to trap, including with `CPU_PMSK` widened and the external-CPU destination pointed at both `0xF` and `9`, versus frames arriving with the forward action plus the L2 entry.

I also added the ACL result, since it is a second and independent path to the same conclusion: a rule matching the group intercepts the frames — the LACP receive counters stop advancing while it is enabled and resume the moment it is disabled — but they never reach the 8051.

Your `aleLatch` pointer is still the most promising lead I have for closing this properly, and it is on my list. Distinguishing "intercepted" from "delivered" was exactly the instrument I was missing; if it can fetch packet data with flags, it should settle whether the trap ever hands anything to the internal CPU or drops it on the floor.

DrDoof:
I put a patch cord between two free front ports to see the loop path do its thing, and it doesn't hold. Both ports block, both come back to forwarding a forward delay later, the loop reopens, and round it goes. Topology-change counter went 0x00 → 0x51 in 5.5 minutes and kept climbing for as long as the cable was in.

The guard is the problem:

```c
if (port_timers[port] == 0 && !(stp_pflags[port] & STP_PF_TRIPPED)) {
        ...
}
return;   /* a BPDU arriving while the timer runs is ignored */
```

A BPDU that shows up *during* the block does nothing, so the timer expires and the port gets promoted. The comment right above says the opposite ("if the loop persists the BPDUs keep arriving and the port stays blocked") but nothing actually does that. And because both ports block, there's nobody left forwarding to hear the loop in the first place.

That's code this branch adds, not something inherited — `origin/main:rtl837x_stp.c` is 239 lines and has no loop detection at all. It's also less alarming than it sounds: `stp_topology_change()` flushes per port, and I couldn't measure any effect on unrelated traffic. Still, containing the loop is the whole point.

Fix in the branch: let the better Port ID decide for both, and have it act on the *other* port instead of on itself. That port is forwarding by construction, so it keeps hearing the loop and re-arms the held port's timer on every BPDU. The held port only needs to keep transmitting, which the send path already allows in any MSTP state, so none of this rests on a blocked port still receiving — I couldn't establish that either way and didn't want to build on it.

One writer also kills a race I walked into on the first attempt. With both ends deciding for themselves, the winner's re-arm could land in the loser's `port_timers[]` before the loser got to its own frame; the loser then read that as "already blocked", skipped its state change, and the loop stayed open. Whether it worked came down to frame ordering.

Two smaller things came out of it. 802.1D compares the priority before the number and `stp_cnf_send()` already puts `stp_pprio[]` on the wire next to it, so the comparison uses that first — otherwise `stp port N prio` would silently have no bearing on which end of a looped pair survives. And since the port number arrives in a frame, and our bridge MAC is in every BPDU we send, it's bounded to the ports this module manages before it indexes anything. Outside that range nothing would ever release the block anyway: `stp_timers()` only walks `min_port..max_port` and skips ports that aren't STP-enabled.

The helper is `__reentrant` like the two functions above it, so parameters and locals go on the stack. Inlined into `stp_in()` it costs two more bytes of DSEG (that function is `__banked`, so its temporaries can't be overlaid), and that was enough to stop an image carrying LACP from linking at all.

Same setup afterwards, cable in for 1 h 36 min: one port blocked, the other kept forwarding, topology-change counter moved four times total and three of those were the link event and the promotion in the first minute. Then nothing for an hour.

Couple of notes if you re-test this. You may not see `STP: loop detected` at all — if the link supervision already blocked the port when you plugged the cable, the loop path finds a non-zero timer and re-arms without printing. The latch is still visible in the numbers: a blocked port counts down a forward delay, so with nothing re-arming it it would have been promoted something like 380 times over those 95 minutes, each one bumping the counter. It wasn't promoted once.

The other case is weaker. If the frame comes back on the port it left — a loop further out behind an unmanaged switch — the Port IDs are equal, there's no pair to pick from, and the port just holds itself down. That one can only re-arm while it's receiving, so it stays the forward-delay pulse it was before. No worse than now, but not fixed.

Last thing, not really about the fix: the failsafe counts HTTP silence and nothing else, on purpose. So a real loop is only contained while something is still talking to the switch — after `stp_failsafe_s` of quiet, STP turns itself off and the loop is back. Fine given what that watchdog is for, but it does put a bound on "contained".

logicog:
> Last thing, not really about the fix: the failsafe counts HTTP silence and nothing else, on purpose. So a real loop is only contained while something is still talking to the switch — after `stp_failsafe_s` of quiet, STP turns itself off and the loop is back. Fine given what that watchdog is for, but it does put a bound on "contained".

That will be a problem in practice, because the http access may not be used all the time. Maybe failsafe is only entered if after a reboot STP does not converge.

Another point ist this STP or RSTP? And what is the compatibility, here?

DrDoof:
Yes, and I'd go further than you suggest — I think the window is right, just not the one you picked.

The reason it doesn't look at our own port states is the incident it was written for. A neighbouring switch with loop prevention cut our uplink in reaction to our BPDUs, and our own ASIC had every port forwarding at the time. Nothing on our side looked wrong, so a check based on our state would have sat there perfectly content while management was gone. Only shutting up completely — no BPDU transmission at all — got the link back.

That failure happens when you turn STP on, though, not spontaneously three days later. So rather than a boot condition, arm it for a bounded window after `stp on` and after a boot that enables STP from the saved config, then disarm. Protection where the risk actually is, and nobody has to keep a browser open. It also kills a case I hit yesterday, where a long loop test quietly switched STP off underneath me because I wasn't clicking anything for a few minutes.

RST by default — `stp_rstp` starts at 1, and the frame is version 2, type 2, 36-byte body with the version1_length. `stp version stp` drops it to legacy Config BPDUs.

The behaviour underneath is 802.1D, and I'd rather say that plainly than let the frame format imply otherwise. No proposal/agreement, no port-role machine (the module header admits as much), so ports always sit out a full forward delay instead of transitioning rapidly. RST on the wire, legacy bridge in practice.

Receive side takes version 2 or greater with type 2, so MSTP parses fine — 802.1D-2004 14.4 wants a bridge to accept a higher version and treat it as RST — plus legacy Config and TCN. An RSTP neighbour will send us a proposal, never get an agreement, and drop back to timer-based transitions on that link. Slow but correct, which is roughly what you'd expect.

What isn't correct, and I only spotted it writing this up: the flags we send are hardcoded to `0x3c`. Designated, learning, forwarding, for every port, always. Only TC and TCA are dynamic. So a port we've blocked still announces itself as forwarding, and the role is always designated even on our root port. Nothing has gone wrong on my bench because of it, and a neighbour computing its own topology from its own BPDUs isn't going to be misled about the tree — but it's still lying, and that's the kind of thing that surfaces in somebody else's mixed network six months from now rather than mine today.

I'd sooner fix it than write it down as a known limitation. Say the word if you want it in this branch, otherwise I'll do it separately and leave this one about the loop.

DrDoof:
Continued in #314 with the text cut down, rebased onto current main and with
the BPDU flag fix folded in. The discussion here still applies and #314 links
back.

PR #306, opened 2026-08-05
system: show and set the management VLAN in System Settings

DrDoof:
`vlan <id> mgmt` has been available on the CLI for a long time, but nothing reported *which* VLAN currently carries switch management. The setting was invisible from the web UI — the only way to find out was to read the startup config — so this adds the missing half, plus a small alignment fix on the same page.

### Where it lives

`information.json` gains `mgmt_vlan` (0 = untagged), and System Settings gains a picker between **Gateway** and **Language**, filled from the configured VLANs:

```
Gateway:           [10.0.2.1        ]
Management VLAN:   [2 (Lokalna)   v ]
Language:          [English       v ]
```

It belongs with the other switch-wide addressing settings rather than on the VLAN page: it is a property of the switch, not of any one VLAN. When management is untagged there is no VLAN to select, so that state shows as a disabled entry rather than inventing an id the switch would reject.

Changing it asks for confirmation first, and the confirmation says what will happen instead of echoing the action back:

> The switch will start tagging its own traffic with that VLAN. If the port you are connected through does not carry it, this page becomes unreachable and the setting can only be undone over the console. Continue?

That is not decorative caution. Getting this wrong is how a switch drops off the network: the CPU starts tagging its own frames, and if the uplink is not carrying that VLAN the device is simply invisible — link up, forwarding fine, no management. Cancelling puts the picker back where it was.

No parser changes: the command already existed, only the read-back was missing.

### Alignment fix

The System Settings rows all pair a label with a form control, except Model, which is a bare span. The inputs carry `padding: 8px 16px` and a left margin, so their text starts about 32px further in than the model name did — close enough to read as a mistake. Read-only values now get a class with the same metrics, so anything else displayed without a control lines up for free.

### Testing

On a SWTGW218AS, integrated build (this change on top of LACP and STP), after a cold boot, verified against the files the device actually serves:

* `information.json` reports `"mgmt_vlan":2` matching the running configuration
* the picker is populated from `/vlanlist` — `1`, `2 (Lokalna)`, `10 (IoT)`, `20 (Wirtualki)`, `30 (OCFS2)`, `40 (SWARM)` — with the current one preselected
* `system.html`, `system.js`, `style.css` and `i18n.js` on the device carry the select, the call, the CSS rule and the strings in all three language blocks
* the model name lines up with the address fields below it
* no regression: LACP aggregate stays valid, the bond partner stays reachable, pages serve normally

I did not test moving management to a *different* VLAN on the live unit. That path exercises the pre-existing CLI command rather than anything new here, and it is the one that takes the switch off the network. What is new is the read-back, the picker and the confirmation, and those are covered above.

Strings are added to all three language blocks in `i18n.js`, English text in each, so no block ends up with a missing key.


html/style.css
```diff
@@ -82,6 +82,8 @@ object, img {
 .isNOK{ color: #900;}
 .isOK{ color: #090;}
 .ip{padding:8px 16px;margin-bottom: 1em;margin-left: 1em}
+/* read-only values line up with the text inside the input fields */
```
vDorst: Please remove the comments in any `html` or `css`-file. This gets served every time to the client, and is a waste of space.

html/system.js
```diff
@@ -155,3 +166,44 @@ window.addEventListener("load", function() {
   if (langSel) langSel.value = rtlLang;
   systemInterval = setInterval(fetchIP, 1000);
 });
+
+
+var mgmtVlanCurrent = 0;
+
+/* Populate the management-VLAN picker from the configured VLANs. If management
+ * is untagged there is no VLAN to select, so show that as a disabled entry
+ * rather than inventing an id the switch would reject. */
```
vDorst: Same here. No comments

vDorst (changes requested):
I think the mgmtvlan value should be moved to `/vlanlist` endpoint.
`void send_vlanlist(void)` is called and generated a JSON-responce.

I think JSON-responce could be look like this.

```json
{
  "mgmt": 123,
  "vlan": [
    {"id": 1, "name": "LAN"},
    {"id": 10, "name": "Main"}
  ]
}
```

DrDoof:
Done in 4f534f6. /vlanlist answers {"mgmt":N,"vlan":[...]} now, the picker reads both
from that one response, and mgmt_vlan is gone from /information.json. Both
consumers in vlan.js follow the new shape, and the comments in style.css and
system.js are gone.

One thing surfaced while sizing the close: the truncation guard reserved one
byte for the closing bracket and never counted the comma before a non-first
entry, so a worst-case name could already land one byte past outbuf. It
reserves 141 now.

vDorst:
thanks

PR #307, opened 2026-08-05
l2: fix the CPU/SFP port label, add sorting and filtering

DrDoof:
Two changes to the L2 Configuration page: a display bug that mislabels most of the table, and sorting/filtering so the table stays usable once it fills up.

### The SFP port was shown as CPU

`getL2()` mapped the port from logical to physical numbering, keeping logical 9 (the CPU) as 9. `fillL2()` then labelled anything numbered 9 as CPU. On an 8+1 board the SFP port also maps to physical 9, so every entry learned on the SFP came out as CPU.

On the switch here that was 26 of 30 entries — the whole uplink looked like CPU traffic:

```
before:  CPU  dc:a6:32:a3:53:af  002  learned      (this is the SFP port)
after:   9    dc:a6:32:a3:53:af  002  learned
```

Fixed by labelling the CPU while the number is still logical, before the mapping, so the two cannot collide:

```js
e.port = e.port == 9 ? 'CPU' : logToPhysPort[e.port];
```

### Sorting and filtering

The table lists every learned and static entry in one flat block. That is fine with a handful and unusable with a few hundred — finding where one MAC sits, or what a given port has learned, meant reading all of it.

Column headings now sort (click again to reverse), and each has a filter box. Filters are substring matches combined with AND, so narrowing to "port 8, static only" is two keystrokes. A counter above the table shows matched out of total, so a filter that hides everything reads as a filter rather than as an empty table.

Two details worth noting:

* The CPU entry is kept out of the numeric comparison when sorting by port — it sorts last, instead of landing between 8 and 9 where a string-vs-number compare would put it.
* The filter inputs sit inside the existing header cells rather than in a second header row. The paint loop addresses data rows as `rows[i+1]`, so a second header row would have shifted every one of them.

Filtering and sorting are applied at render time from the fetched set, so incremental loading of large tables is unaffected.

### Testing

Built clean; `node --check` passes on the modified `l2.js`. Strings added to all three language blocks in `i18n.js`.

Not yet exercised on hardware — I can flash an integrated build and confirm the port column and the header behaviour against a live forwarding table if that is useful before merging.


html/l2.js
```diff
@@ -64,14 +64,65 @@ function delL2(idx) {
   xhttp.timeout = 1500; xhttp.send();
 }
 
+var l2All = [];
+var l2SortCol = 'port';
+var l2SortDir = 1;
+
+/* Sort keys: ports are numbers except the CPU, which must not compare as one. */
```
vDorst: No comments.
Although they are useful, they are byte wasters on the device.

I am thinking more and more to:
1. use single file minifier-tools to reduce the JS-code while building the image.
2. Pre-compress the JS-file and always serve them in compress format.
   But this needs modification in the http-code.
Atleast it helps with `i18n.js`
```
20969 bytes html/i18n.js
 5680 bytes html/i18n.js.gzip (GZIP)
 4840 bytes html/i18n.js.br   (Brotli)
```

html/l2.js
```diff
@@ -64,14 +64,65 @@ function delL2(idx) {
...
+  renderL2();
+}
+
+function l2FilterChanged() { renderL2(); }
+
+function renderL2() {
+  var tbl = document.getElementById('l2table');
+  if (!tbl) return;
+  var f = {};
+  ['port', 'mac', 'vlan', 'type'].forEach(function(c) {
```
vDorst: Define the array once with
```js
const cols = ['port', 'mac', 'vlan', 'type'];
```

```js
cols.forEach(function(c) { ... };
```

You can ask the AI to improve and reduce the functions.

html/l2.js
```diff
@@ -104,7 +155,10 @@ function getL2() {
         e.vlan = parseInt(e.vlan, 16);
         e.idx = parseInt(e.idx, 16);
         e.type = e.type == "s" ? t('l2_static') : t('l2_learned');
-        e.port = e.port == 9 ? 9 : logToPhysPort[e.port];
+        /* Label the CPU port before mapping to physical numbering: the
+         * SFP port maps to physical 9 as well, and tagging afterwards
+         * relabelled every SFP entry as CPU. */
```
vDorst: Save as above about the comments.

vDorst (changes requested):
See comments

DrDoof:
Done in 22a09bd. The three comments are gone, 357 bytes across l2.js and
style.css, and the column list is a single const now.

Pre-compressing on build sounds right to me, especially with i18n.js shrinking
to a quarter. If the httpd side grows a Content-Encoding path I can pitch in on
the html half.

vDorst:
Thanks.

> Pre-compressing on build sounds right to me, especially with i18n.js shrinking
to a quarter. If the httpd side grows a Content-Encoding path I can pitch in on
the html half.

Lets keep this idea in mind and make a prototype when you have some spare time.

PR #308, opened 2026-08-06
docs: Add Horaco XZ-SG4T2 skeleton documentation and images

plaes:
Horaco XZ-SG4T2 seems to be already supported since 247cb4dd2f7de0734dd6cc3635c6031fb0592f35 but wasn't listed in docs.

Also some general fixes and improvements.


vDorst:
Thanks, great to see that you change doc/supported_devices.md to use relative paths!

PR #309, opened 2026-08-06
httpd: Fix buffer offerflow in scan_header

### LINKED ISSUE #300: Pre-auth buffer overflow in scan_header() multipart boundary parsing (authentication bypass)
### Summary

`scan_header()` in `httpd/httpd.c` copies the `multipart/form-data; boundary=` value from the HTTP `Content-Type` header into the fixed `boundary[72]` buffer with **no length check**. The copy runs on every POST *before* authentication is evaluated, so a single unauthenticated HTTP request with an over-long boundary string overflows the buffer into adjacent `__xdata` globals.

### Location

```
httpd/httpd.c, scan_header()
    while (content_type[i] != '\r' && content_type[i] != '\n') {
        boundary[i + 4] = content_type[i];
        i++;
    }
    ...
    boundary[i + 4] = 0;
```

`boundary` is declared `__xdata uint8_t boundary[72];`. `i` is fully controlled by the attacker via the request header and is never bounded against the buffer size.

### Impact

- **Pre-authentication.** `handle_post()` calls `scan_header()` before the `authenticated` flag is checked, so no credentials are required to reach the loop.
- In the current build the globals that follow `boundary` in `__xdata` include the session/auth state and the stored device password, so the overflow can flip the switch into an authenticated state and/or corrupt the password — i.e. a remote **authentication bypass**, not just a crash.
- Reachable within a single normal-length HTTP header, over the management network.

This affects every device running RTLPlayground, since the web server is shared across all machine targets.

### Suggested fix

Bound the loop so the terminating write stays inside the buffer. The last in-bounds index for the terminator is `boundary[71]`, which corresponds to stopping the copy at `i == 67`:

```c
while (content_type[i] != '\r' && content_type[i] != '\n' && i < 67) {
```

Over-long boundaries are then truncated and the upload simply fails, instead of corrupting adjacent state. Happy to open a PR with this change if useful.

plaes:
Fixes #300

### DIFF
--- a/httpd/httpd.c
+++ b/httpd/httpd.c
@@ -272,7 +272,8 @@ __xdata uint8_t *scan_header(__xdata uint8_t *p)
 		dbg_string("\nFound multipart\n");
 		content_type += 30;
 		uint8_t i = 0;
-		while (content_type[i] != '\r' && content_type[i] != '\n') {
+		while (i < (sizeof(boundary) - 5) &&
+				content_type[i] != '\r' && content_type[i] != '\n') {
 			boundary[i + 4] = content_type[i];
 			i++;
 		}


vDorst:
Thanks for fixing this issue.

PR #313, opened 2026-08-08
Reject VLAN IDs and numeric arguments the hardware cannot take

### LINKED ISSUE #311: VLAN IDs from the CLI aren't range-checked
`vlan 4095 1 2` creates a VLAN the rest of the firmware can't see. `vlan_get()` refuses
anything `>= 0xfff` (rtl837x_port.c:168) and `stp_fdb_update()` only scans `vid < 4095`
(rtl837x_stp.c:683), but `vlan_create()` has no such guard. It writes the ID straight
into the table index register, so the entry ends up in hardware while the CLI, the GUI
and the config dump all stay blind to it.

Above 4094 it gets worse. `vlan 5000` puts 0x1388 into the index register. Depending on
how many bits the chip honours, that either lands on nothing or overwrites a different
VLAN. I haven't measured which one. It's what the register write implies and I didn't
want to find out on a live switch.

`atoi_short()` in cmd_parser.c adds a second layer. There's no overflow check, so
`vlan 65540` wraps to 4 and quietly edits VLAN 4 instead. `atoi_byte()` has the same hole.

Happy to send a patch if you want it that way: overflow detection in `atoi_short()` and
`atoi_byte()`, a range check in `parse_vlan()`, and a guard in `vlan_create()` mirroring
the one in `vlan_get()` so the config-upload path is covered too.

Correcting myself: I first wrote that 056a30a already did this for byte arguments, which
reads as if it were in main. It isn't. That commit sits on the STP branch in #303 and is
still waiting on you, so this is the other half of a change you haven't merged, not a
follow-up to one you did.

### LINKED ISSUE #312: mtu writes a value it failed to parse
`mtu 1 abc` sets the port MTU to 0.

The handler throws away atoi_short()'s return value and leans on its own bound
test instead:

    atoi_short(&mtu, cmd_words_b[2]);
    if (mtu > 0x3fff) {

atoi_short() zeroes its output before it starts parsing, so a non-numeric third
word leaves mtu at 0. That isn't greater than 16383, so the write goes ahead.

The chip takes it. I set a port to 0 and read back 0x0000, no clamping, no
substituted default. On a live 2.5G LAG member it stopped taking frames: LACPDU
receives moved by 4 in fifteen seconds against 21 on the sibling port, and those
4 arrived in the seconds before the write landed. The partner state flipped to
expired. Writing 9216 back brought both the counter and the partner state
straight back. So a typo in the third word is enough to quietly take a port out.

Checking the return covers it:

    if (atoi_short(&mtu, cmd_words_b[2])) {
            print_string("mtu [port] [size]\n");
            return;
    }

Happy to send that as a patch. Separate from #311, which is about VLAN IDs, but
worth doing together: the overflow check I proposed there changes what reaches
this line rather than whether it does. `mtu 1 99999` wraps to 34463 today and
gets caught by the bound test; with that patch it stops at 9999 and goes through.

DrDoof:
Fixes #311. Closes #312.

Seven small commits, all in the same vein: numbers from the CLI reached
registers without anyone checking they fit.

The parser half: atoi_short() and atoi_byte() refuse the digit that would wrap
the value past its type. The command half: vlan_create() and vlan_delete() get
the range check vlan_get() has had all along, parse_vlan() rejects the same
values with the usage message, mtu takes 64 to 16383 and checks the parse result
instead of throwing it away, pvid takes 1 to 4094 and says so when it refuses,
and a management VLAN above 4094 falls to usage rather than parking management
on a VLAN that cannot exist. The last commit validates the port argument of mtu
and pvid: both used to turn the first character into an index with no check, so
"mtu 0 100" read far past the end of phys_to_log_port[] and wrote the size to
whatever register the garbage pointed at. lag and vlan already validated theirs.

Two behaviour notes. A failed parse leaves the partial value alone instead of
tidying it to 0, because 0 is the one number that measurably hurts here: I put
0 in as a port MTU on a live 2.5G LAG member and its LACPDU receives moved by 4
in fifteen seconds against 21 on the sibling port, partner expired, recovered on
restoring the size. That measurement is also where the lower MTU bound comes
from. And "mtu show" gained a return it was missing, so it stops printing a
garbage port byte after the table.

None of it costs internal RAM: cmd_parser.rel stays at DSEG 1, OSEG 2, BSEG 9,
rtl837x_port.rel at DSEG 0, OSEG 5, and the image reports the same 10207 bytes
of XDATA before and after. Every commit builds on its own, so bisect stays
usable.

Boundaries were checked by running the old and new parsing side by side over
65535, 65536, 65540, 255, 256, port words 0, 9, 10, 12, a non-numeric word, and
our own 37-line startup config replays clean against all of the new checks. The
VLAN paths themselves haven't been on hardware: this is built from plain main
and flashing it would take the STP work back off my switch. Say so if you want
that first.

056a30a in #303 fixes atoi_byte by widening the accumulator instead, so it isn't
needed on top of this. Happy to swap to that shape if you prefer it.

### DIFF
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -736,8 +736,7 @@ void parse_mtu(void)
 		print_string("mtu [port] [size]\n");
 		return;
 	}
-	atoi_short(&mtu, cmd_words_b[2]);
-	if (mtu < 64 || mtu > 0x3fff) {
+	if (atoi_short(&mtu, cmd_words_b[2]) || mtu < 64 || mtu > 0x3fff) {
 		print_string("MTU must be 64..16383\n");
 		return;
 	}
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -347,6 +347,8 @@ void parse_vlan(void)
 			return;
 		}
 		if (cmd_compare(2, "mgmt")) {
+			if (vlan_settings.vlan > 4094)
+				goto err;
 			management_vlan = vlan_settings.vlan;
 			if (!vlan_settings.vlan)
 				print_string("Management VLAN disabled\n");
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -191,8 +191,11 @@ uint8_t atoi_byte(__xdata uint8_t *out, uint8_t idx)
 	uint8_t num = 0;
 
 	while (isnumber(cmd_buffer[idx])) {
+		uint8_t val = cmd_buffer[idx] - '0';
 		err = 0;
-		num = (num * 10) + cmd_buffer[idx] - '0';
+		if (num > 25 || (num == 25 && val > 5))
+			return 1;
+		num = (num * 10) + val;
 		idx++;
 	}
 
@@ -209,6 +212,8 @@ uint8_t atoi_short(__xdata uint16_t *vlan, uint8_t idx)
 	while (isnumber(cmd_buffer[idx])) {
 		err = 0;
 		uint8_t val = cmd_buffer[idx] - '0';
+		if (*vlan > 6553 || (*vlan == 6553 && val > 5))
+			return 1;
 		*vlan = (*vlan * 10) + val;
 		idx++;
 	}
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -354,6 +354,8 @@ void parse_vlan(void)
 				print_string("Management VLAN set to "); print_short(management_vlan); write_char('\n');
 			return;
 		}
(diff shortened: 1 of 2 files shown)


vDorst:
Thanks.

> 056a30a in #303 fixes atoi_byte by widening the accumulator instead, so it isn't
needed on top of this. Happy to swap to that shape if you prefer it.

Your current implementati, in this PR, is fine.

PR #314, opened 2026-08-08
stp: Spanning Tree support for the RTL837x switches

DrDoof:
Replaces #303 and #302 in one thread, rebased onto current main, with the PR
text cut down as asked. The review history and the hardware evidence live in
those two threads.

What it is: RST BPDUs on the wire, 802.1D behaviour underneath. No
proposal/agreement machine, so an RSTP neighbour falls back to timed
transitions. Config BPDUs and TCNs are answered, MSTP BPDUs accepted.

Delivery is the part specific to these boards: the RMA trap action cannot reach
the 8051, so BPDUs ride the forward action constrained to the CPU port by a
static L2 entry per VLAN in use. With STP off the entries revert to flooding, so
a surrounding tree spans through the device. doc/stp.md describes it.

Safety around it: a commit-confirm failsafe, reworked the way the old thread
asked. Enabling STP arms a one-shot window; one HTTP request or serial console
command inside it confirms management survived and disarms it, a silent window
drops STP and restores forwarding. The window re-arms whenever something newly
blocks a port, a rejoining port, root guard, the loop latch, so a quiet network
runs unsupervised and a topology change still has a net under it. A latch keeps a
looped port blocked instead of oscillating, measured at 4 topology changes in 96
minutes against 81 in 5.5 minutes without it. The last commit derives the BPDU
flag byte from the port state instead of announcing designated plus forwarding
unconditionally, verified on the wire: a listening port sends role designated
with learn and forward clear.

One commit landed after opening: stp_setup() warns when an STP-enabled port
admits tagged frames only, since such a port can never receive the untagged
BPDUs and silently turns edge. Exercised on hardware both ways, warning present
and absent, with the election coming back on the spot.

Tested on a SWTGW218AS: hour-long soak, BPDU fuzzing, loop latch and the flag
byte all exercised on hardware. stp.rel costs DSEG 5 and no OSEG, image XDATA
10498 bytes.

PR #315, opened 2026-08-10
Feat/webui: compress the embedded assets and switch to a single-page layout

eraiza0816:
## Summary

The Web UI — originally 34 files under `html/` — is embedded into the
flash image and served over HTTP.  This pull request shrinks it to 8
files and roughly a quarter of its original size with four changes:

1. **Minification.**  A new, conservative minifier (`tools/minify.py`)
   removes comments (`<!-- -->`, `//`, `/* */`) and indentation from the
   HTML/JS sources without ever touching string or regex literals, so the
   output is guaranteed to be functionally identical to the input.  The
   minified files go to `output/html_min/`, which becomes the input to
   `fileadder` for both the header generation and the final image.  As
   comments and indentation are pure overhead on the wire, minification
   alone already cuts the transfer size from 109,814 to 97,730 bytes,
   independent of gzip.
2. **gzip compression.**  `fileadder -z` compresses every embedded file
   (zlib, gzip format, `Z_BEST_COMPRESSION`).  The generated `f_data`
   entries carry a `gzip` flag, and the httpd adds `Content-Encoding:
   gzip` for flagged entries.  The 8051 only copies compressed bytes from
   the flash to the socket; decompression happens for free in the
   browser.
3. **One script bundle.**  All 17 page scripts (i18n, shared helpers,
   navigation and the per-page scripts) are merged into a single
   `main.js` that the pages load as their only script.  The browser
   fetches it once, and the gzip dictionary now spans all of the code.
4. **Single-page layout.**  `index.html` holds all pages as sections
   (overview, ports, stat, vlan, l2, mirror, lag, eee, bandwidth,
   system, update); the sidebar switches between them via the URL hash
   (`#/ports`, `#/vlan`, ...).  `login.html` stays separate as the
   authentication gate.  A section's initialisation runs when the
   section is shown, its polling runs only while it is visible, and
   re-entering a section refetches the data.

Comments left in the JS/HTML are stripped at build time anyway, so the
readability that the single-file consolidation costs can be made up
with inline comments going forward.

Only the static file table is affected; the JSON API responses
(`/information.json`, `/vlanlist`, ...) are never compressed.

A full minifier such as terser could shrink the assets further, but it
cannot guarantee that the output stays functionally identical to the
source (removing newlines, in particular, interacts with automatic
semicolon insertion), and a broken Web UI can only be recovered by
reflashing the device.  We therefore chose a conservative, provably
equivalent minifier over a smaller but riskier one.

## Measured effect

The following numbers were measured with `make MACHINE=KP_9000_6XHML_X2`
(the Web UI is machine-independent).  The columns show the raw source,
the minified file and the gzip-compressed file:

| File | Raw | Minified | Gzip | % of raw |
|------|-----:|---------:|-----:|---------:|
| main.js | 81793 | 72043 | 17231 | 21.1% |
| index.html | 15641 | 13744 | 3384 | 21.6% |
| style.css | 4009 | 3855 | 1342 | 33.5% |
| favicon.ico | 1906 | 1906 | 766 | 40.2% |
| switch.svg | 1906 | 1832 | 750 | 39.3% |
| sfp.svg | 1828 | 1750 | 728 | 39.8% |
| port.svg | 1765 | 1709 | 602 | 34.1% |
| login.html | 966 | 891 | 499 | 51.7% |
| **Total** | **109814** | **97730** | **25302** | **23.0%** |

The embedded data block ends at `0x5b977` without compression and at
`0x462d6` with it — roughly 86 KB of flash are freed on a 512 KB image.

The minified assets are also what a browser receives when gzip is
unavailable, so the transfer size drops in both stages: 109,814 →
97,730 bytes from minification, then → 25,302 bytes from gzip.

The consolidation helps at every step: the 17 scripts gzip to 23,302
bytes as separate files but to 17,231 bytes as one bundle, and the
single-page layout removes the repeated downloads entirely — with the
multi-page UI the browser re-fetched `main.js` on every navigation,
while the SPA fetches everything once and switches sections in-page.
Measured on a real device (PCB-K0402WS-V3): cold load ~450 ms, section
switches ~60-160 ms, and only the visible section polls the JSON API.

## Bug fixes included

While wiring this up, a few latent bugs had to be fixed because they
would have made the sizes or the build nondeterministic:

- `fileadder` wrote the NUL terminator one byte past the file content
  (`data_read + 1`).  The `strlen()`-based size computation therefore
  depended on uninitialised buffer content (the image is read into the
  same buffer), and the stray byte was even compressed along with the
  file.  The NUL now lands directly after the content.
- `version.h` was listed in `.PHONY`, so every build rewrote it while
  parallel builds (`make -j`) could read a half-written copy.  It is a
  regular prerequisite now.
- The `find` expression was unscoped and matched only `*.js`, `*.html`
  and `*.svg`.  It is now scoped to `html/` and also covers `*.css` /
  `*.ico`, so the dependency list matches what `fileadder -d` actually
  embeds.
- `minify.py` stripped an escaped slash before a closing `/` as a `//`
  comment because it did not recognise regex literals (e.g. `/^#\//`);
  the tokenizer now copies regex literals verbatim.

## Verification

- `make MACHINE=KP_9000_6XHML_X2` and `make MACHINE=PCB_K0402WS_V3`
  build cleanly (sdcc 4.5.0), both sequentially and in parallel.
- Every embedded entry was decompressed from the image and compared
  byte-for-byte against the minified sources (8/8 files, checked
  automatically with the snippet in `doc/webui-compression.md`).
- On a real device (PCB-K0402WS-V3): all sections render through the
  sidebar navigation without JavaScript errors; re-entering a section
  refetches its data without duplicating rows; the JSON API stays
  uncompressed; and the firmware-update path through the Web UI
  (upload → reset → boot-time image promote) works.
- `tools/httpd_sim` serves the raw files from `html/` (its buffer size
  was raised so the simulator can serve the larger bundle).

## Documentation

`doc/webui-compression.md` describes the pipeline, the measured effect
and how to verify a build.

---

**One question before we go any further:**

I noticed that pull request #220 ("Fastui") has been quiet for some
time.  I would be happy to take that work forward myself if it helps,
but I would not want to take over someone else's task without asking
first.  Would it be all right if I proceeded?


logicog:
Please go ahead, this will be a great improvement.I believe #220  was mostly exploratory in order to see what could be done and anyway does not follow the same ideas as your proposal.

vDorst:
I wonder what if we put every JS-file inside only file so it can be better compressed and served once. But I think that [HTTP_ETag](https://en.wikipedia.org/wiki/HTTP_ETag) is needed for that.

eraiza0816:
Thanks for the suggestion! I'll give the file merging a try, and I'll also verify the effect of HTTP_ETag with and without it.

For now I think ETag can be left out of this PR — the httpd closes the connection after every response anyway, so fewer requests help on their own, and conditional GET (If-None-Match + 304) isn't implemented in the firmware yet. But it's worth measuring, so I'll test that along with the merged bundle.

I'll commit the merged version as soon as it's ready, so please wait a moment — I'll switch this PR to Draft for now.

eraiza0816:
I've committed the consolidated JS bundle and updated the documentation accordingly. I'll start experimenting with HTTP ETag next.

vDorst:
Only support ETAG if it is needed! Is it needed? Does browser redownload the `main.js` everytime?

eraiza0816:
I measured both and came to the same conclusion — I've dropped the ETag. The numbers made it clear that the single-page layout is the better trade-off:

ETag: within the 60 s max-age the browser does not request main.js at all, so ETag only matters after the cache expires. I confirmed on the device that the browser then sends If-None-Match and gets a 304 Not Modified instead of the ~17 KB body — it works, but it merely makes the occasional re-download cheap.
SPA: main.js and the whole UI are downloaded once, and the sidebar switches between sections via the URL hash with no re-downloads at all. Measured on a real device: cold load ~450 ms (the same as one page of the multi-page UI), section switches ~60-160 ms — the ~1.4 s outlier is the VLAN page fetching its data, which is identical in both layouts — and only the visible section polls the JSON API.
In short: the multi-page UI re-fetches main.js (17 KB gzip) on every navigation after the cache expires, while the SPA fetches everything once and then navigates in-page. ETag would have optimised the former; the SPA removes it altogether, so I went with that.

eraiza0816:
I've taken this pull request out of draft status, and it is now ready for your review. I would very much appreciate it if you could take a look when you have a moment. Thank you!

Makefile
```diff
@@ -118,11 +131,11 @@ $(BUILDDIR)/rtlplayground-$(FILENAME_EXTENSION).bin: $(BUILDDIR)/rtlplayground.i
 	tools/output/imagebuilder -i $^ $@
 	tools/output/fileadder -a $(DEFAULT_CONFIG_LOCATION) -s $(IMAGESIZE) -d config.txt $@
 	tools/output/fileadder -a $(CONFIG_LOCATION) -s $(IMAGESIZE) -d config.txt $@
-	tools/output/fileadder -a $(HTML_LOCATION) -s $(IMAGESIZE) -d html -p html_data -b BANK1 $@
+	tools/output/fileadder -a $(HTML_LOCATION) -s $(IMAGESIZE) -z -d $(HTML_MIN) -p html_data -b BANK1 $@
 	tools/output/crc_calculator -u $@
 	ln -sf $(MACHINE)/rtlplayground-$(FILENAME_EXTENSION).bin output/rtlplayground.bin
 
-.PHONY: clean all $(SUBDIRS) $(VERSION_HEADER)
+.PHONY: clean all $(SUBDIRS)
```
feelfree69: Checked out this PR, built it, but version.h was not rebuilt. So VERSION_SW and BUILD_DATE were claiming old versions.

eraiza0816: Sorry about that, I accidentally deleted it during another verification. I'll put it back.

logicog:
Could you have a look at #336 and look into how to make the best out of the ideas by coordinating a bit?

eraiza0816:
I have posted a comment on 336.

eraiza0816:
For now, I will start working on resolving the conflicts.

eraiza0816:
I've resolved the conflicts. I would appreciate it if you could review the pull request once more when you have a moment. Thank you.

Makefile
eraiza0816: [4c4232c](https://github.com/logicog/RTLPlayground/pull/315/commits/4c4232cc07e6251d0e2682fc5b0746080eb7521d)

fix makefile

vDorst:
I hope other people could review it to and give a approval as well.

Remove the space comment.

feelfree69:
> I see some spaces/tabs left over

Not sure what you mean here. 

If I'm not totally mistaken, the actual source (wrote by humans) should keep all it's comments, whitespaces etc in order to keep the readability as good as possible.
Then comes the minifier and strips out whatever is possible.

vDorst:
Hmm, maybe it was my diff, and that is was in the old data/code.

bloqaudio:
Tested on a SWTGW218AS on current main, with one unmerged commit stacked on top: the #317 board support, without which this PCB has no working SFP uplink. Nothing else. The pipeline ran clean on a fresh build tree, nothing beyond python3 needed.

Wire numbers first: the whole UI is now 22.6 KB gzipped on the wire (index 3.4 KB, main.js 17.4 KB, style 1.3 KB, login 0.5 KB) against roughly 64 KB spread over the old pages, with correct Content-Encoding and Content-Type headers, and the served archives decompress byte-clean.

Behavior: I drove all eleven hash routes headlessly. Every section renders and populates, the ports table shows all ports, the L2 table fills through walkL2 and stays stable across refresh cycles, and statistics, EEE, bandwidth, VLAN and the overview all load their data. The login gate works and the removed per-page files answer 404 as they should.

The only console errors I saw are the two my network already produces on the multi-page UI, an unguarded JSON.parse in the status pollers and an occasional connection reset from the one-connection httpd under fast navigation, both present in my baseline capture from before this PR. One such reset briefly left the bandwidth table part-filled during a fast walk; a reload filled it, same as the old UI behaves.

Works on my hardware. Happy to re-test any iteration, and as far as I am concerned this unblocks the sequencing we discussed on #336.

eraiza0816:
Thank you very much for testing on the SWTGW218AS and for the detailed report — it's great to hear that everything works on your hardware.

Regarding #317: I understand it has been suggested that it should be merged, but it is still open at the moment. I have already resolved the differences with main on my side. Also, #317 has not been mentioned anywhere in this PR or in issue #336, so it hasn't been part of our discussion so far. For now, would it be possible for you to take care of the SWTG018AS-V2.1.0 (the SWTGW218AS variant) board-support diff on your end, while I handle the coordination here?

I'm happy to keep iterating on this PR, and please don't hesitate to let me know if you need anything from me.

vDorst:
I wonder if we need a fallback only for the landing page. So if we detect that the browser don't accept `Accept-Encoding: gzip`, we serve them a simple non-compressed html page with the message they need a browser with gzip support. Which is already a [standard for over 11 years](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept-Encoding).

I think it only effects the website it self not the API-endpoints. 

- pros: it provides a better user experience for browser that don't support it.
  But I think this is a very small group of users that hit this issue.
- cons: If we don't do it, it saves a lot of passing of the data.
  Non of the api-calls are not effected by it anyway.

eraiza0816:
I agree that we can skip implementing the fallback. 

Currently the firmware only embeds the gzip-compressed copies of the assets and serves them unconditionally with `Content-Encoding: gzip`, so a fallback would mean adding an extra uncompressed file plus an `Accept-Encoding` check to the httpd — and since gzip has been a baseline for every major browser for well over two decades ([RFC 1952](https://datatracker.ietf.org/doc/html/rfc1952), [RFC 2616](https://datatracker.ietf.org/doc/html/rfc2616), and universally supported across engines per [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept-Encoding)), the number of users who would ever hit it is practically zero. Custom clients that skip gzip would talk to the JSON API directly anyway. 

Keeping the implementation simple and avoiding the extra overhead seems like the best approach.

vDorst:
I would like to see this too be merged.

But here is a conflict, because #326, #343, #314 and maybe more, are touching the JS-file.
I see that at least 'l2.js' and 'main.js' are already changed in `main`-branch.
So a new set of JS-file are needed.
But which PRs we want in before this PR?

vDorst:
I would like to propose to rebase this PR after these PR's are merged, likely are ready in a few days:
- #333
- #329
- #380
So we can go forward on the #336 improvements.

eraiza0816:
thank you for laying out the plan — that makes perfect sense. Apologies for the delayed response; work kept me away for a while.

I agree: let us rebase #315 once #333, #329, and #380 have landed. That should give us a clean baseline to move forward with the #336 improvements. Just ping me when those are in and I will get this PR rebased and up to date.

vDorst:
@eraiza0816 All the issues are merged. So we can move forward with your PR.
Please rebase, so we can all test it again.

eraiza0816:
Ok. I'm starting the rebase process.

eraiza0816:
The rebase is complete. I have done a quick sanity check on real hardware, but I would appreciate it if others could also take a look. Thank you!

vDorst:
Thanks for the rebase.
Sorry that #400 is added.
I would also review and merge #399 and #401 before you rebase again.
I hope that is not to much work?

eraiza0816:
@vDorst, thanks for letting me know — and no problem at all.

I'll wait until #399 and #401 are merged, then rebase #315 again on top, including #400. Just ping me when they're in.

vDorst:
@eraiza0816 All the issues are resolved.
Please rebase, thanks.

httpd/httpd.c
vDorst: this can be replaced with for the main_js. See also #389.

logicog:
Let's not overdo it with the optimization, here. It is best to merge this as soon as possible so that we can also include other things like #409 again, quickly. We can always optimize the code of this PR after it was merged. But that way round we reduce the wait time for the other ideas. It is always best to hold a lock for as little time as possible ;-)

eraiza0816:
I will resolve the current conflicts.

From then on, I would like you to perform a rebase for each pull request—how does that sound?

eraiza0816:
The rebase is complete.
I have done a quick sanity check on real hardware, but I would appreciate it if others could also take a look. Thank you!

vDorst:
Thanks for the rebase.

This PR gets priority now for all the web-related features/issues.
So the other PRs have to adapt/rebase when this is merged.

I would like to see 3 or more approvals from the regulars because of the big change.
So we can merge this in a few days.
I going to review it this evening.

feelfree69:
First 2 issues faced;
- After navigating to 192.168.10.247, I got constant reloads with "wrong password", without any chance to enter the correct password. The magic suddenly slowed down or stopped completely, and after entering the password the UI appeared as before, without this PR.
- I can't enter anything new in System Settings, in the System- as well as in Advanced-Tab: The original content from flash gets constantly reloaded; discarding any manually entered characters.

Tested with Firefox 155.0 for Windows 11.

eraiza0816:
It is possible that repeated rebasing has caused issues in my build environment.
I will clean the environment and try again locally.

eraiza0816:
I was able to reproduce this in Chrome on my end as well. I will fix it.

（I’m tearing my hair out because I don’t know at which rebase step this issue was introduced, but I’ll give it my best shot....）

eraiza0816:
I identified the cause and fixed it.

logicog:
I did a test-drive and it is amazing. Once the overview page is loaded the first time, all further pages are blazingly fast.
I played around with many settings, but not systematically. I did not find any issues, except the following:
Under System Settings, I changed the language, then clicked "Update Settings", but nothing happened. Maybe I made a mistake, or maybe this was broken before, but I would have expected the pages to be in the selected language.

logicog:
Found another issue: On the Port statistics page, clicking on show for port n seems to show the statistics for port n-1. Clicking on the show button for port 1 shows nothing.

vDorst:
- l2.json: it keeps fetching every 3 sec when switch to other tab from "port statistics".
- Bandwitdh limit set, it post 2x the same command "bw in 3 03e8"
- "Port Configuration", port images flicker. I see that the port.svg request starts with "GET /port.svg undefined" instead of "GET /port.svg HTTP/1.1" like the JSON and JS requests.

eraiza0816:
Thank you all for testing — I have fixed the reported issues:

- Language setting: the onchange handlers were lost when the pages were merged into index.html, so changing the language (and the management VLAN) had no effect. Restored both.
- Port statistics: the show button passed a 0-based index, so port n showed port n-1 (and port 1 showed nothing). Now passes i + 1, as before.
- L2 polling: the "connected devices" walk re-scheduled itself unconditionally, so it kept fetching l2.json after leaving the ports section. It now only continues while the section is visible and stops on section switch.
- Bandwidth: applying an ingress limit re-sent the identical bw in command when flow control was enabled. The redundant second POST is removed.
Verified on real hardware (SWGT024 V2.0 Managed): language switches immediately, row 3 requests port=3, no l2.json traffic after leaving the section, and applying a limit sends each command once.

One thing I could not reproduce: the GET /port.svg undefined request line and the image flicker. Each SVG is fetched exactly once (HTTP 200, gzip) with no failures, and nothing in the code constructs request lines by hand — that string may come from the logging setup. If you can still reproduce it, I would appreciate details on how you captured it.

eeeple:
Did a quick test with the latest commit (af4101f0604d914aa7aa552f2f2adcb5d4ac7c8a) on a PCB-K0402WS-V3.0 and it's incredible. Everything seems in order apart from EEE which did not refresh instantly and took a couple of seconds (but that might be due to the underlying logic).

logicog:
I also did a test-drive and it is amazing. No issues whatsoever. For me it is ready to be merged.

feelfree69:
The only issue I found so far: Port Statistics, the All counters "Show" Button doesn't seem to work on Port 1. There's simply no popup at all. The other counters for Port 1 are working.

Regarding the flickering of the port images: I saw these too in the iinitial build; but not with the actual head of this PR.

MACHINE=SWTGW218AS

vDorst:
Same as @feelfree69  http://192.168.10.247/counters.json?port=1 is rejected for a 8x1SFP device.

MACHINE="KP_9000_9XHML_X_V3_1"

But works fine for a 4x2SFP device

This is the issue. https://github.com/logicog/RTLPlayground/blob/main/httpd/
page_impl.c#L305
I assumed that a invalid port is just 0. But that can also be a valid value on a 9-port device.


> One thing I could not reproduce: the GET /port.svg undefined request line and the image flicker.

This happens with Firefox.

The port issue is not a website issue, which can be fixed later.
The rest is looking fine. So you have my approval.

vDorst:
```patch
index b99f0b9..fe6a99b 100644
--- a/httpd/page_impl.c
+++ b/httpd/page_impl.c
@@ -299,11 +299,9 @@ void send_vlan(uint16_t vlan)
 bool send_counters(uint8_t phys_port)
 {
        uint8_t phys_port_idx = phys_port - 1;
-       if (phys_port_idx > 8)
+       if (phys_port_idx < machine.min_port || phys_port_idx > machine.max_port)
                goto err;
        uint8_t log_port = machine.phys_to_log_port[phys_port_idx];
-       if (log_port == 0)
-               goto err;
 
        dbg_string("send_counters called: "); dbg_byte(phys_port_idx); dbg_char('\n');
        slen = strtox(outbuf, HTTP_RESPONCE_JSON);
```

this should fix the http://192.168.10.247/counters.json?port=1 is rejected issue.

Also add a entry in Changelog.md.

eraiza0816:
I tested @vDorst's patch on real hardware (SWGT024 V2.0 Managed) and it breaks this machine in both directions, so I went with a corrected variant.

Test result with the patch as proposed:

| port | before | with patch | expected |
|------|--------|------------|----------|
| 1, 2, 3 (valid) | 200 | **400** | 200 |
| 4, 5, 6 (valid) | 200 | 200 | 200 |
| 7, 8, 9 (don't exist) | 400 | **200** | 400 |

Why: min_port/max_port are logical port numbers (`machine.h`), but the patch compares them against the physical index. On the SWGT024 (`min_port = 3`, mapping `{4,5,6,7,8,3,0,0,0}`) that rejects valid physical ports 1–3 and accepts non-existent ports 7–9.

What I committed instead (`httpd/page_impl.c`): keep the physical bounds check, then validate the mapped logical port against the range:

```c
if (phys_port_idx > 8)
    goto err;
uint8_t log_port = machine.phys_to_log_port[phys_port_idx];
if (log_port < machine.min_port || log_port > machine.max_port)
    goto err;
```

I don't have a 9-port device here, so I could only verify the fix on the SWGT024 (ports 1–6 return 200, non-existent ports 7–9 return 400). 
Could someone with 9-port hardware please check that `counters.json?port=1` now works there? Thank you!

logicog:
I tested it on a 9-port KP_9000_9XH_X_EU and a 4+2-port KP_9000_6XHML_X2 and in both cases the port statistics work as expected.

vDorst:
> Also add a entry in Changelog.md.

My idea was that you added this PR as topic under feature/added. Not my fix.

I am not happy with the final fix.
My patch should bound-check the access to `machine.phys_to_log_port[]`.
Maybe the check was incorrect.
But now we are checking the result of array.

Maybe this is better:
Calculate the max port we have based on `machine.max_port - machine.min_port`.
So that is this the max index value for `machine.phys_to_log_port[]`.
```c
   uint8_t phys_port_idx = phys_port - 1;
   if (phys_port_idx > (machine.max_port - machine.min_port))
                goto err;
   uint8_t log_port = machine.phys_to_log_port[phys_port_idx];
```

eraiza0816:
Apologies for missing the nuance on the CHANGELOG.md — 
I have reworked the entry to cover this PR's topic under Added. I also switched the code to your bound-check variant and verified on the SWGT024 that ports 1–6 return 200 and non-existent ports 7–9 return 400. Does this match what you expected?

vDorst:
> I have reworked the entry to cover this PR's topic under Added. I also switched the code to your bound-check variant and verified on the SWGT024 that ports 1–6 return 200 and non-existent ports 7–9 return 400. Does this match what you expected?

Yes, port 0 should also return 400.

I am happy, so I approve this PR.

vDorst (approved):

logicog:
Happy, too. Really great work and thanks for having consistently followed up on this for such a long time!

PR #316, opened 2026-08-10
rtl837x_igmp: cast &entry to __xdata pointer in memset()

bloqaudio:
Building with SDCC 4.5.0 stops in rtl837x_igmp.c with error 78 (incompatible pointer types): the firmware's own memset() takes __xdata pointers, but &entry is passed as a generic pointer. Cast it explicitly so the call resolves.

No behaviour change; the ipmc/l2mc table entries already live in __xdata.


logicog:
Good catch. Thanks so much for fixing! Will merge.

PR #317, opened 2026-08-10
machine: add PCB SWTG018AS-V2.1.0 variant of the SWTGW218AS

bloqaudio:
Reworked after the report that other boards ship under the SWTGW218AS label and work with the existing definition. `MACHINE_SWTGW218AS` is now left completely untouched, and the changes land as a new machine define named by the PCB silkscreen, `MACHINE_PCB_SWTG018AS_V2_1_0`, following the existing PCB-named defines. Boards with this PCB never detect an SFP module with the old definition (detect is GPIO38 here, and no LOS pin is wired), so an SFP-uplinked switch comes up unreachable.

The LED-block init also moved from 21 inline `REG_SET` calls to a table walked by a loop. The inline version no longer links against current main, bank 0 overflows, and the table is considerably smaller.

Tested against current main on a SWTG018AS-V2.1.0 board: without the change the SFP+ module is never detected; with it the module detects, links at 10G, and the bi-color LED behaves, green at 2.5G and below, blue at 10G.


feelfree69:
I believe there are multiple Devices with Label SWTGW218AS out there....
At least [mine ](https://github.com/logicog/RTLPlayground/blob/main/doc/devices/SWTGW218AS.md) works fine with the existing implementation....

At the moment, I do not have a good idea how to deal with such ambigous Devicenames....

bloqaudio:
That matches what I ran into today from the other side. I flashed a build of current main onto one of my units, which uplinks over SFP+, and the module was never detected, so the switch came up with no uplink and looked dead from the network until I moved the cable to a copper port. Same label, different wiring: detect is GPIO38 on my boards and there is no LOS pin. The PCB silkscreen on mine reads SWTG018AS-V2.1.0.

I have reworked the PR so `MACHINE_SWTGW218AS` stays exactly as it is and the changes are a new define named after the PCB, `MACHINE_PCB_SWTG018AS_V2_1_0`, following the existing PCB-named machines, so your board is unaffected now. Verified on hardware against current main: module detect, 10G link, and the bi-color SFP LED all work on the new define, and the old define is byte-for-byte unchanged. What does the silkscreen on your board say? I would like to record the label-to-PCB mapping in the device doc so the next person knows which define to build.

feelfree69:
Thanks!

PR #318, opened 2026-08-10
machine: optionally read the factory MAC from flash

bloqaudio:
RTLPlayground generates a locally-administered MAC, so the address does not
match the sticker on the case and DHCP reservations cannot follow the device.

Add an optional per-board `machine.mac_flash_offset`. When non-zero, the
firmware reads a 6-byte MAC from that flash address at boot and uses it if it
is a valid globally-administered unicast address; anything else falls back to
the generated LAA as before. Offset 0 - the default for every board -
preserves current behaviour exactly.

Enable it for the SWTGW218AS, whose stock firmware keeps the factory MAC in
nvcfg at 0x1FC000.

### DIFF
--- a/machine.c
+++ b/machine.c
@@ -415,6 +415,7 @@ void machine_custom_init(void) { }
 __code const struct machine machine = {
 	.machine_name = "SWTGW218AS 8+1 Managed Switch",
 	.isRTL8373 = 1,
+	.mac_flash_offset = 0x1FC000,
 	.min_port = 0,
 	.max_port = 8,
 	.n_sfp = 1,
--- a/machine.h
+++ b/machine.h
@@ -85,6 +85,7 @@ typedef struct machine {
 	uint32_t led_sets[4][4];
 	uint8_t led_mux_custom;
 	uint8_t led_mux[28];
+	uint32_t mac_flash_offset;
 };
 
 typedef struct machine_runtime
--- a/rtlplayground.c
+++ b/rtlplayground.c
@@ -2112,20 +2112,28 @@ void main(void)
 	uip_ipaddr(&uip_hostaddr, ownIP[0], ownIP[1], ownIP[2], ownIP[3]);
 	uip_ipaddr(&uip_draddr, gatewayIP[0], gatewayIP[1], gatewayIP[2], gatewayIP[3]);
 	uip_ipaddr(&uip_netmask, netmask[0], netmask[1], netmask[2], netmask[3]);
-	reg_read_m(RTL837X_REG_CHIP_UUID);
-#ifdef DEBUG
-	print_string("SoC UUID: "); print_sfr_data();
-#endif
-	uip_ethaddr.addr[0] = 0x06;  // LAA prefix
-	uip_ethaddr.addr[3] = sfr_data[0] ^ sfr_data[3];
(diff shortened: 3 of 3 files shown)


machine.c
```diff
@@ -415,6 +415,7 @@ void machine_custom_init(void) { }
 __code const struct machine machine = {
 	.machine_name = "SWTGW218AS 8+1 Managed Switch",
 	.isRTL8373 = 1,
+	.mac_flash_offset = 0x1FC000,   // stock keeps the factory MAC here; read it instead of LAA
```
logicog: Comment is not needed.

machine.h
```diff
@@ -85,6 +85,9 @@ typedef struct machine {
 	uint32_t led_sets[4][4];
 	uint8_t led_mux_custom;
 	uint8_t led_mux[28];
+	// Flash offset of a factory-programmed 6-byte MAC (stock nvcfg). 0 = none: generate a
```
logicog: The comment is not needed, the function becomes clear from its use.

rtlplayground.c
```diff
@@ -2112,20 +2112,33 @@ void main(void)
...
-	uip_ethaddr.addr[4] = sfr_data[1] ^ sfr_data[3];
-	uip_ethaddr.addr[5] = sfr_data[2] ^ sfr_data[3];
-	reg_read_m(RTL837X_REG_CHIP_LOT_NO);
-#ifdef DEBUG
-	print_string(", LOT: "); print_sfr_data(); write_char(' ');
-#endif
-	uip_ethaddr.addr[1] = sfr_data[0] ^ sfr_data[2];
-	uip_ethaddr.addr[2] = sfr_data[1] ^ sfr_data[3];
+	// Factory MAC from flash if this board stores one (machine.mac_flash_offset != 0).
+	// Stock keeps 6 raw MAC bytes at 0x1FC000; RTLPlayground only uses 0-0x100000 so that
```
logicog: Please reduce the comment, or better remove it entirely, what the code does is quite clear. The "Stock keeps 6 raw MAC bytes at 0x1FC000" is not necessarily correct for other boards as it is configurable, similarly "so generic builds on other hardware are unaffected" is out of context, because the reader does not know which "other hardware" is meant. Looks like an AI comment to the programmer of the patch, not so much intended for the subsequent reader of the code.

rtlplayground.c
```diff
@@ -2112,20 +2112,33 @@ void main(void)
...
+	// region is free and survives web upgrades (they rewrite only 0-CONFIG_START). Fall back
+	// to a generated locally-administered MAC when the region is blank/invalid, or when the
+	// board defines no offset — so generic builds on other hardware are unaffected.
+	uip_ethaddr.addr[0] = 0xff;
+	if (machine.mac_flash_offset) {
+		flash_region.addr = machine.mac_flash_offset;
+		flash_region.len = FLASH_BUF_SIZE;
+		flash_read_bulk(flash_buf);
+		// accept only a real unicast, globally-administered address (reject blank/LAA/multicast/all-zero OUI)
```
logicog, line 2120: In contrast, this comment is quite useful, as it explains what the intention of the code below is.

logicog (commented):
Really great idea and improvement. Code looks good, but some comments should be removed.

logicog:
Thanks a lot for the quick fixes!

PR #319, opened 2026-08-10
cmd_editor: fix cmd_buffer overflow and CLI hang on long input

bloqaudio:
Typing or pasting a line of 128 characters or more into the CLI hangs the editor and overruns `cmd_buffer[128]`: the length check did not reserve room for the terminating NUL written on Enter, and on a full buffer the character was retried via `continue` without advancing the serial-ring read pointer, so input processing never caught up again.

Cap the line at CMD_BUF_SIZE-1 and, when the line is full, drop the character but still fall through so the ring byte is consumed.

Note on diff size: the functional change is three lines; the rest is re-indentation of the insert-and-echo block. `git diff -w` shows the minimal form.

The host-side unit tests originally bundled here are split out to #324.

### DIFF
--- a/cmd_editor.c
+++ b/cmd_editor.c
@@ -40,21 +40,24 @@ void cmd_edit(void) __banked
 {
 	while (l != sbuf_ptr) {
 		if (sbuf[l] >= ' ' && sbuf[l] < 127) { // A printable character, copy to command line
-			if (cmd_line_len >= CMD_BUF_SIZE)
-				continue;
-			write_char(sbuf[l]);
-			// Shift buffer to right
-			for (uint8_t i = cmd_line_len; i > cursor; i--)
-				cmd_buffer[i] = cmd_buffer[i-1];
-			// Insert char in comand buffer
-			cmd_buffer[cursor++] = sbuf[l];
-			cmd_line_len++;
-			// Print rest of line
-			for (uint8_t i = cursor; i < cmd_line_len; i++)
-				write_char(cmd_buffer[i]);
-			// Move backwards
-			for (uint8_t i = cursor; i < cmd_line_len; i++)
-				write_char('\010'); // BS works like cursor-left
+			// Reserve one byte for the terminating NUL written on Enter. When the
+			// line is full, drop the character but still fall through to advance the
+			// serial-ring read pointer below; a 'continue' here would spin forever.
+			if (cmd_line_len < CMD_BUF_SIZE - 1) {
+				write_char(sbuf[l]);
+				// Shift buffer to right
+				for (uint8_t i = cmd_line_len; i > cursor; i--)
+					cmd_buffer[i] = cmd_buffer[i-1];
+				// Insert char in comand buffer
+				cmd_buffer[cursor++] = sbuf[l];
+				cmd_line_len++;
+				// Print rest of line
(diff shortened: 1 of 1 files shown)


test/sdcc_shim.h
```diff
@@ -0,0 +1,36 @@
...
+#define SDCC_SHIM_H
+
+/* Memory-space qualifiers → nothing (host has one flat address space). */
+#define __xdata
+#define __code
+#define __pdata
+#define __idata
+#define __data
+#define __bit         unsigned char
+#define __sbit        unsigned char
```
vDorst: I am not sure if it helps but `bool` describes the `__bit` and `__sbit` better.

cmd_editor.c
```diff
@@ -40,21 +40,24 @@ void cmd_edit(void) __banked
 {
 	while (l != sbuf_ptr) {
 		if (sbuf[l] >= ' ' && sbuf[l] < 127) { // A printable character, copy to command line
-			if (cmd_line_len >= CMD_BUF_SIZE)
-				continue;
```
vDorst, line 44: Can't the change just be?
```c
if (cmd_line_len >= CMD_BUF_SIZE - 1)
   continue;
```

rtl837x_common.h
```diff
@@ -145,7 +145,9 @@ void reg_write_m(uint16_t reg_addr);
 void sds_read(uint8_t sds_id, uint8_t page, uint8_t reg);
 void sds_write_v(uint8_t sds_id, uint8_t page, uint8_t reg, uint16_t v);
 void delay(uint16_t t);
-void sleep(uint16_t t);
+#ifndef RTLP_HOST_TEST
+void sleep(uint16_t t);         /* clashes with <unistd.h> in host test builds */
+#endif
 void write_char_no_syslog(char c);
```
vDorst: Is it a option to group all the `#ifndef RTLP_HOST_TEST` together so it is more clear?

vDorst (commented):
I really like this idea and would love to see more of these tests.
But is a great start!

This is not a full review, but some question/comments.

logicog:
Can you split off the tests for now, please and put them into a different commit? They are not covered by the description of the PR, so maybe make a second PR?

logicog:
As @vDorst  mentions, I do not understand why there are so many changes in cmd_parser.c.

bloqaudio:
Done — the tests are split out to #324 and this PR is now only the cmd_editor fix.

@logicog on the amount of change in cmd_editor.c: the functional change is three lines; the rest is re-indentation of the insert-and-echo block that moved inside the new `if`. `git diff -w` shows the minimal form.

@vDorst on `if (cmd_line_len >= CMD_BUF_SIZE - 1) continue;` — that keeps the overflow fix but reintroduces the hang: the ring read pointer `l` is only advanced at the bottom of the `while (l != sbuf_ptr)` loop, so on a full line the `continue` spins forever on the same character. That stuck loop is the second half of the bug, which is why the full character has to fall through so the ring byte is consumed. Your other two suggestions (`bool` for `__bit`/`__sbit`, grouping the `RTLP_HOST_TEST` guards into one block) are both incorporated in #324.

vDorst:
A yes I see, thanks for pointing it out.

Thanks!

PR #320, opened 2026-08-10
makefile: derive BUILD_DATE deterministically

bloqaudio:
BUILD_DATE currently comes from wall-clock time, so two builds of the same
commit differ. Honor SOURCE_DATE_EPOCH, falling back to the HEAD commit date
and then to wall-clock when git is absent. Same-commit builds become
byte-identical (BUILD_DATE is baked into the image and covered by the
trailing CRC), which makes it possible to verify that a binary matches a
source tree.

Verified: two clean builds of the same commit produce identical sha256 sums.


logicog:
Perfect! Tested and works great. Will merge.

logicog:
Thanks so much for contributing this!

PR #321, opened 2026-08-10
dhcp: announce the hostname via option 12

bloqaudio:
Now that the device has a configurable hostname, send it as DHCP option 12 on
discover and request so the DHCP server can register the device in local DNS.
Skipped while the hostname is still empty, i.e. before
set_hostname_default() has run.

### DIFF
--- a/dhcp.c
+++ b/dhcp.c
@@ -41,6 +41,7 @@ __xdata uip_ipaddr_t server;
 #define DHCP_REBIND_LEN		4
 #define DHCP_CLIENT_ID		61
 #define DHCP_CLIENT_ID_LEN	7
+#define DHCP_HOSTNAME		12
 #define DHCP_REQUEST_IP		50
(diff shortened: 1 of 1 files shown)


logicog:
Thanks for this contribution! Great idea and code looks good.
Could you remove the comment in line 120, though, it is not necessary. The comment actually looks more like an AI instruction to the programmer than being helpful to the reader of the code. It is actually a bit confusing because none of the other similar functions have a comment, so the reader expects that the function actually does something unexpected, which is not the case. The "Option 12" first looks suspicious because there is no 12 being obviously used in the function, and if one were to change the 12 above, then the comment would be contradictory. In general an internal function used only from within the file does not need comments unless it does something unexpected.

bloqaudio:
Removed the comment, thank you — I will do better in the future. I utilize AI to help with my notes to keep things consistent on my end, and that one slipped through into the code.

logicog:
Perfect!

PR #322, opened 2026-08-10
httpd: send real verdicts for firmware and config uploads

bloqaudio:
Firmware uploads end in a silent connection close, so the client cannot
distinguish a verified upload from a failed one - the browser just sees the
connection drop while the device resets (or doesn't).

Send an explicit verdict instead: 200 with "OK: checksum verified, rebooting"
or 400 with "NO: checksum failed, not applied", with Content-Length set so
the browser completes the response without waiting for a close that the reset
would swallow. The chip now resets only once the verdict has been fully
ACKed.

Config-sector uploads get the same treatment: a 200 response instead of a
silent close.

The existing web UI keeps working unchanged - it previously treated the close
as completion and now simply receives a real response - and UIs that want to
can show the actual outcome. httpd_sim mirrors both behaviours.

### DIFF
--- a/httpd/httpd.c
+++ b/httpd/httpd.c
@@ -51,6 +51,9 @@ __xdata uint32_t max_upload;
 __xdata uint16_t short_parsed;
 
 __xdata char passwd[21];
+// Set when a verified firmware upload awaits its response ACK, after
+// which the chip resets to apply the staged image
+__xdata uint8_t fw_reset_pending;
 __xdata char session_id[SESSION_ID_LENGTH + 1];
 __xdata uint8_t authenticated;
 __xdata uint32_t now;
@@ -81,6 +84,7 @@ void httpd_init(void) __banked
 	// Start listening to port 80
 	uip_listen(HTONS(80));
 	s->tstate = TSTATE_CLOSED;
+	fw_reset_pending = 0; // xdata is not zeroed by the startup code
 }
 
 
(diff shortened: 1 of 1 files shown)


httpd/httpd.c
```diff
@@ -342,18 +346,32 @@ uint8_t stream_upload(uint16_t bptr)
...
+						"NO: checksum failed, not applied\n");
 				}
+			} else {
+				// Config upload: answer instead of closing the
+				// connection unacknowledged (a zero-byte close makes
+				// fetch() fail with NetworkError even though the
+				// sector was written)
+				slen = strtox(outbuf, "HTTP/1.1 200 OK\r\nContent-Length: 19\r\n"
+					"Content-Type: text/plain\r\n\r\n"
+					"OK: config written\n");
```
vDorst, line 374: I wonder if it better to already switch to a json responce, so the future website can show a nice message/popup?

bloqaudio: Do you mean just the body format, or moving update.html to a fetch()-based upload with a popup now? Format alone would not help yet: the page today is a plain form POST, so the browser renders the body directly and the user would see raw JSON instead of a readable verdict. The machine-readable part is already the status code (200/400). If you mean the full fetch()-based updater, I like that idea but would rather keep this PR to the transport fix and do the updater as a follow-up PR, where the JSON verdict, the JS that parses it, and the popup land together (the html and httpd ship in the same image, so nothing can get out of sync in between).

vDorst: Yes I mean the body of the response.

But you do have a valid point, that the text is now shown directly to the user.
I was not aware of that, I never use the webupdate.
So keep it as is.

vDorst:
This needs a rebase.
Can this maybe merged into #359 and close this PR?

bloqaudio:
Agreed. The config rework merged in #356 made the config half of this PR redundant, since buffered config uploads already get a real response. The firmware verdict half is reworked on top of #359 as its own commit, so this PR is superseded. Closing in favor of #359.

PR #324, opened 2026-08-11
test: add host-side unit-test harness, covering cmd_editor

bloqaudio:
Follow-up to #319, split out as requested there. The branch is stacked on the #319 fix commit so the overflow regression tests pass; once #319 merges only the harness commit remains.

Adds `make -C test`: selected firmware translation units are compiled unmodified under host gcc with AddressSanitizer/UBSan, force-including a shim header that erases the SDCC 8051 keywords, and linked against a mocked serial ring and console. The first suite covers the CLI line editor: line entry, backspace editing, and both long-input overflow cases (length cap and ring-pointer stall).

The only firmware-source change is one `#ifndef RTLP_HOST_TEST` block in rtl837x_common.h grouping the libc-named prototypes (sleep/memcpy/memset/strlen/strcpy/strcmp) so host builds take them from the C library instead; a full SDCC target build compiles unchanged with the grouped block.

Review suggestions from #319 are incorporated: `__bit`/`__sbit` map to `bool` (matches SDCC's 0/1 bit normalization), and the host-test guards sit in a single grouped block.


DrDoof:
Gave this a spin and it does what it says on the tin.

`make -C test` built and passed 5/5 first try, on macOS with Apple clang rather than gcc, so it isn't welded to the machine it was written on. Then I merged it onto current main, 55 commits past the branch point: clean merge, tests still green, and a full SDCC build for SWTGW218AS comes out fine. So the grouped prototypes really are invisible to the target build, as advertised.

Two small things I noticed along the way.

`-Wno-builtin-declaration-mismatch` is gcc-only, so clang greets every build with `unknown warning option`. Purely cosmetic, but it's the first thing you see on a Mac.

The other one is more fun, because the harness has already paid for itself. Running the firmware headers through a second compiler turned this up in machine.h:

```c
typedef struct machine { ... };
typedef struct machine_runtime { ... };
```

Two typedefs with no name, so the keyword isn't doing anything there. Nothing is broken: everything says `struct machine` and `struct machine_runtime`, no alias is ever used, and sdcc doesn't say a word about it. It predates this PR and doesn't belong in it, but catching that sort of thing is exactly what a host build is for, which feels like a decent argument for merging this one.

vDorst:
This is my result with the current main.
```
make -C test
make: Entering directory '/home/r/dev/c/RTLPlayground/test'
mkdir -p build
gcc -std=gnu11 -g -O1 -Wall -Wextra -Wno-unused-parameter -Wno-unknown-pragmas -Wno-builtin-declaration-mismatch -DRTLP_HOST_TEST -fsanitize=address,undefined -fno-omit-frame-pointer -include sdcc_shim.h -I. -I.. -I../httpd -I../uip test_cmd_editor.c support.c ../cmd_editor.c -o build/test_cmd_editor -fsanitize=address,undefined
In file included from ../cmd_editor.c:2:
../machine.h:103:1: warning: useless storage class specifier in empty declaration
  103 | };
      | ^
../machine.h:109:1: warning: useless storage class specifier in empty declaration
  109 | };
      | ^
=== build/test_cmd_editor ===
== cmd_editor host tests ==
[test] basic line entry + Enter
  ok:   Enter marks a command available
  ok:   cmd_buffer holds the typed line
[test] backspace edits the line
  ok:   backspace then retype yields 'hello'
[test] C4: overlong line does not hang the editor
  ok:   line length caps at CMD_BUF_SIZE-1 (room reserved for NUL)
[test] C4: full line + Enter does not overflow cmd_buffer
  ok:   128-char line + Enter completes without ASan overflow

5 checks, 0 failed

ALL TESTS PASSED
```
Gives two warnings, can these be fixed?

bloqaudio:
Done, thanks both. `machine.h` loses the two `typedef` keywords that declared no alias (every user says `struct machine` and `struct machine_runtime`), which is what your two warnings were, and `-Wno-builtin-declaration-mismatch` is now added only when the compiler is not clang, so a Mac build starts quiet too. Rebased onto current main while at it. `make -C test` is 5/5 with no warnings here, `make machine_check` is clean and a full SDCC build for the SWTGW218AS links.

vDorst:
That is better.
Is it useful to run this on every PR?

bloqaudio:
Yes, and cheaply: it builds with the gcc the workflow already installs and runs in under a second, so I added it as a step before `machine_check`, so a host-visible regression fails the build first. Pushed as a third commit on this branch. A run of the workflow in the `debian:trixie` container passes with the tests executed, sanitizers included.

vDorst:
Tested on my own repo and it did work, also pushed a commit that broke the test which return an CI error.

Thanks!

PR #325, opened 2026-08-11
stp: treat a link aggregation group as one entity

DrDoof:
Rebased onto #314, which is where it should have sat from the start. It was carrying its own copy of that branch's history: 32 of its 40 commits were byte identical patches to commits there, which is the real reason the two conflicted in nine files. It is now eight commits on top of #314 and nothing else, so those eight are the whole of what this PR asks you to read.

```
stp: make a link aggregation group a port of its own
stp: factor the entity tests out
stp: report a trunk's state from one of its members
stp: refill the transmit budget for every entity
stp: derive a trunk's BPDU flags from a member's hardware state
stp: reapply trunk state when membership changes
httpd: leave a trunk member out of the STP port list
html: keep the trunk form of the STP commands in the saved config
```

In order: a group becomes an entity standing beside the physical ports, the entity tests move out of the places that were spelling them out, the state shown for a trunk is taken from one of its members because the hardware holds none for the group itself, the transmit budget is refilled per entity rather than per port, the BPDU flags for a trunk come from a member's hardware state, a change of membership re-applies the entity's state to whatever joined or left, the port list leaves members out, and the saved configuration keeps the trunk form of the commands instead of expanding them to members.

Dropping the duplicated history also dropped what those commits had inherited from before the management failsafe went away. Every single conflict in the rebase was that failsafe trying to come back, and I took the current side each time, so none of it survives here.

One thing this still does not do, and you will see it the moment you run it: the status table added in #314 walks physical ports, so it lists the members of a trunk with their own role and counters and has no row for the entity at all. What that row should say is a design question rather than a mechanical one, so I would rather ask than guess. My inclination is a row per entity with the members hidden, the way the port list already handles them, but say if you would rather see both.

Costs on top of #314: 451 bytes of BANK1, 1472 of BANK2 and 139 of xdata. Built for SWTGW218AS and KP_9000_6XHML_X2 on sdcc 4.5.0. Not run on hardware in this form yet, and the earlier hardware results on the old lineage do not carry over, since the base underneath has moved a long way.


DrDoof:
Draft on purpose, because of what it sits on.

This branches off #314 rather than off main, so the 40 commits and 19 files
GitHub shows here are mostly that PR. Only the last eight are this work, and they
are the ones touching the entity model, which is what a trunk becomes a port of.
None of it means anything until #314 lands, so there's nothing to review here
yet and I'd rather it didn't compete for your attention while the other one is
open.

What it does not depend on is LACP. Membership comes from the trunk membership
registers the `lag` command already writes on main, read once a second, so a
static group behaves the same as a negotiated one and #299 is not in the way
either.

Once #314 merges I'll rebase onto main and take it out of draft. Eight commits
against a clean base.

DrDoof:
Now that #347 is in, the map here calls `port_lag_members_get()` instead of decoding the register itself, so the copy this PR used to carry is gone and `rtl837x_stp.c` no longer touches the membership register at all.

DrDoof:
Superseded by #397, which is the same idea rebuilt as one commit on current main.

This branch predates the trunk/lag rename and 45 of its 61 commits have landed here by other routes since, so rebasing it would have meant re-doing work that is already in. Closing rather than leaving it to rot, but happy to reopen if you would rather review it in this shape.

PR #326, opened 2026-08-11
Add/refactor helpers to parse IP, numbers and port arguments

vDorst:
While hunting for SRAM, I saw a lot of places that do manual parsing of the `cmd_buffer`.
Mostly the same parsing code, but a bit changed, in multiple parts of the code.
So I made helpers to parse:
- u8/byte `atoi_byte()`
- u16/short `atoi_short()`
- port partially `cmd_parse_port()`, so a `1t` can be parsed
- port complete `cmd_parse_port_separator()` so number must end with a space or NUL.
- Add and use helper `cmd_is_space_or_nul()`.

The parsers also return the number of bytes they consumed.
So the overall parser can keep track the index into `cmd_buffer`.

All this saves only 1 SRAM but almost +1300 of code space.

```patch
diff ../rtlplayground.mem  output/SWGT024_V2_0_MANAGED/rtlplayground.mem
3,10c3,10
< 0x00:|0|0|0|0|0|0|0|0|a|b|b|b|b|b|b|b|
< 0x20:|B|B|B|T|d|d|e|e|e|e|e|e|e|e|e|e|
< 0x40:|e|e|e|f|f|f|f|f|g|g|g|g|g|g|g|g|
< 0x50:|g|g|g|g|g|g|g|h|h|h|h|h|h|h|h|h|
< 0x60:|h|i|i|i|i|i|i|i|i|i|j|j|j|j|j|j|
< 0x70:|j|j|j|j|j|j|Q|Q|Q|Q|Q|Q|Q|Q|S|S|
---
> 0x00:|0|0|0|0|0|0|0|0|a|a|a|a|a|a|a|a|
> 0x10:|a|a|a|a|a|a|a|a|a|a|b|b|b|b|c|c|
> 0x30:|d|d|d|d|d|d|d|d|d|d|d|d|d|d|d|d|
> 0x40:|e|e|e|e|e|f|f|f|f|f|f|f|f|f|f|f|
> 0x50:|f|f|f|f|g|g|g|g|g|g|g|g|g|g|h|h|
> 0x60:|h|h|h|h|h|h|h|i|i|i|i|i|i|i|i|i|
> 0x70:|i|i|i|j|Q|Q|Q|Q|Q|Q|Q|Q|S|S|S|S|
21c21
< 16 bit mode initial stack starts at: 0x7e (sp set to 0x7d) with 130 bytes available.
---
> 16 bit mode initial stack starts at: 0x7c (sp set to 0x7b) with 132 bytes available.
28,29c28,29
<    EXTERNAL RAM     0x0001   0x255b    9563    16777216
<    ROM/EPROM/FLASH  0x0000   0x2d5d5  96090    16777216
---
>    EXTERNAL RAM     0x0001   0x2556    9558    16777216
>    ROM/EPROM/FLASH  0x0000   0x2d0c2  94749    16777216
```

Focus points:
- Port parser. I think we can remove the CPU part, because non of the commands need to include the CPU as a port.
- parse_vlan and parse_mirror, mainly the part to parse 't' on the end.

### DIFF
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -381,7 +381,7 @@ void parse_lag(void)
 			print_string(" member ports: ");
 			for (uint8_t j = 0; j < 10; j++) {
 				if (members & 1) {
-					print_port(j);
+					print_phys_port(j);
 					write_char(' ');
 				}
 				members >>= 1;
@@ -569,7 +569,7 @@ void parse_isolate(void)
 		members = port_isolation_get(port_configured);
 		for (uint8_t i = 0; i < 10; i++) {
 			if (members & 1) {
-				print_port(i);
+				print_phys_port(i);
 				write_char(' ');
 			}
 			members >>= 1;
@@ -634,7 +634,7 @@ void parse_ingress(void)
 		// Setting mode for all ports at once
 		for (log_port = machine.min_port; log_port <= machine.max_port; log_port++) {
 			if (!port_ingress_filter(log_port, mode)) {
-				print_string("Error setting ingress filter for port "); print_port(log_port); write_char('\n');
+				print_string("Error setting ingress filter for port "); print_phys_port(log_port); write_char('\n');
 				return;
 			}
 			print_string("All ports ingress filter set to: ");
@@ -652,14 +652,14 @@ void parse_ingress(void)
 			idx += ret;
 
 			if (!vlan_ingress_mode_parse(cmd_buffer[idx++], &mode) || !cmd_is_space_or_nul(idx)) {
-				print_string("Invalid ingress mode for port "); print_port(log_port); print_string(" in ingress command\n");
+				print_string("Invalid ingress mode for port "); print_phys_port(log_port); print_string(" in ingress command\n");
 				goto err;
 			}
 			if (!port_ingress_filter(log_port, mode)) {
-				print_string("Error setting ingress filter for port "); print_port(log_port); write_char('\n');
+				print_string("Error setting ingress filter for port "); print_phys_port(log_port); write_char('\n');
 				return;
 			}
-			print_string("Port "); print_port(log_port);
+			print_string("Port "); print_phys_port(log_port);
 			print_string(" ingress filter set to: ");
 			print_port_ingress_filter_mode(mode); write_char('\n');
 		}
@@ -684,7 +684,7 @@ void parse_mirror(void)
 			print_string("NOT Enabled: ");
 		}
 		print_string("Mirroring port: ");
-		print_port(mirroring_port);
+		print_phys_port(mirroring_port);
 		reg_read_m(RTL837x_MIRROR_CONF);
 		uint16_t m = sfr_data[0];
 		m = (m << 8) | sfr_data[1];
@@ -838,7 +838,7 @@ void parse_mtu(void)
 		for (p = machine.min_port; p <= machine.max_port; p++) {
 			reg_read_m(RTL8373_REG_MAC_L2_PORT_MAX_LEN + ((uint16_t) p << 8));
 			uint16_t mtu = SFR_DATA_U16 & 0x3fff;
-			print_string("Port "); print_port(p);
+			print_string("Port "); print_phys_port(p);
 			write_char(' '); print_short(mtu); write_char('\n');
 		}
 		return;
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -27,7 +27,6 @@
 
 extern __code struct machine machine;
 extern __xdata uint8_t stpEnabled;
-extern __code uint8_t log_to_phys_port[9];
 
 extern volatile __xdata uint32_t ticks;
 extern volatile __xdata uint8_t sfr_data[4];
@@ -356,7 +355,7 @@ void parse_lag(void)
 			print_string(" member ports: ");
 			for (uint8_t j = 0; j < 10; j++) {
 				if (members & 1) {
-					write_char('0' + machine.log_to_phys_port[j]);
+					print_port(j);
 					write_char(' ');
 				}
 				members >>= 1;
@@ -545,10 +544,7 @@ void parse_isolate(void)
 		members = port_isolation_get(port_configured);
 		for (uint8_t i = 0; i < 10; i++) {
 			if (members & 1) {
-				if (i < 9)
-					write_char(machine.log_to_phys_port[i] + '0');
-				else
-					print_string("CPU");
+				print_port(i);
 				write_char(' ');
 			}
 			members >>= 1;
@@ -613,7 +609,7 @@ void parse_ingress(void)
 		// Setting mode for all ports at once
 		for (log_port = machine.min_port; log_port <= machine.max_port; log_port++) {
 			if (!port_ingress_filter(log_port, mode)) {
-				print_string("Error setting ingress filter for port "); print_byte(machine.log_to_phys_port[log_port]); write_char('\n');
+				print_string("Error setting ingress filter for port "); print_port(log_port); write_char('\n');
 				return;
 			}
 			print_string("All ports ingress filter set to: ");
@@ -622,7 +618,6 @@ void parse_ingress(void)
 	} else {
 		for(uint8_t w = 1; w < cmd_words_len; w++) {
 			idx = cmd_words_b[w];
-			char p = cmd_buffer[idx];
 			uint8_t ret = cmd_parse_port(idx);
 			if (ret != 1) {
 				print_string("Invalid physical port number\n");
@@ -632,14 +627,14 @@ void parse_ingress(void)
 			idx += ret;
 
 			if (!vlan_ingress_mode_parse(cmd_buffer[idx++], &mode) || !cmd_is_space_or_null(idx)) {
-				print_string("Invalid ingress mode for port "); write_char(p); print_string(" in ingress command\n");
+				print_string("Invalid ingress mode for port "); print_port(log_port); print_string(" in ingress command\n");
 				goto err;
 			}
 			if (!port_ingress_filter(log_port, mode)) {
-				print_string("Error setting ingress filter for port "); write_char(p); write_char('\n');
+				print_string("Error setting ingress filter for port "); print_port(log_port); write_char('\n');
 				return;
 			}
-			print_string("Port "); write_char(p);
+			print_string("Port "); print_port(log_port);
 			print_string(" ingress filter set to: ");
 			print_port_ingress_filter_mode(mode); write_char('\n');
 		}
@@ -664,7 +659,7 @@ void parse_mirror(void)
 			print_string("NOT Enabled: ");
 		}
 		print_string("Mirroring port: ");
-		write_char('0' + machine.log_to_phys_port[mPort >> 1]);
+		print_port(mirroring_port);
 		reg_read_m(RTL837x_MIRROR_CONF);
 		uint16_t m = sfr_data[0];
 		m = (m << 8) | sfr_data[1];
@@ -818,7 +813,7 @@ void parse_mtu(void)
 		for (p = machine.min_port; p <= machine.max_port; p++) {
 			reg_read_m(RTL8373_REG_MAC_L2_PORT_MAX_LEN + ((uint16_t) p << 8));
 			uint16_t mtu = SFR_DATA_U16 & 0x3fff;
-			print_string("Port "); print_byte(machine.log_to_phys_port[p]);
+			print_string("Port "); print_port(p);
 			write_char(' '); print_short(mtu); write_char('\n');
 		}
 		return;
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -81,6 +81,7 @@ __xdata uint16_t cmd_history_ptr;
 __xdata uint8_t err_status;
 
 __xdata uint16_t atoi_results_short;
+__xdata uint8_t atoi_results_u8;
 
 inline uint8_t isletter(uint8_t l)
 {
@@ -187,23 +188,29 @@ uint8_t atoi_hex(uint8_t idx)
 }
 
 
-uint8_t atoi_byte(__xdata uint8_t *out, uint8_t idx)
+// return 0 on error or non-zero number of number-byte taken for the conversion.
+// Stops at any non-digit '0'-'9' char or bytes is more then 3.
+uint8_t atoi_byte(uint8_t idx)
 {
-	uint8_t err = 1;
+	uint8_t cnt = 0;
 	uint8_t num = 0;
 
-	while (isnumber(cmd_buffer[idx])) {
-		uint8_t val = cmd_buffer[idx] - '0';
-		err = 0;
-		if (num > 25 || (num == 25 && val > 5))
-			return 1;
+	uint8_t * ptr = &cmd_buffer[idx];
+
+	while (1) {
+		uint8_t val = *ptr++ - '0';
+		if (val > 9)
+			break;
+		if (num > 25 || (num == 25 && val > 5) || cnt >= 3)
+			return 0;
 		num = (num * 10) + val;
-		idx++;
+		cnt++;
 	}
 
-	*out = num;
-	return err;
+	atoi_results_u8 = num;
+	return cnt;
 }
+
 
 // return 0 on error or non-zero number of number-byte taken for the conversion.
 // Stops at any non-digit '0'-'9' char or bytes is more then 5.
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -80,6 +80,8 @@ __xdata uint16_t cmd_history_ptr;
 // Error set by commands
 __xdata uint8_t err_status;
 
+__xdata uint16_t atoi_results_short;
+
 inline uint8_t isletter(uint8_t l)
 {
 	// return (l >= 'a' && l <= 'z') || (l >= 'A' && l <= 'Z');
@@ -203,22 +205,26 @@ uint8_t atoi_byte(__xdata uint8_t *out, uint8_t idx)
 	return err;
 }
 
-
-uint8_t atoi_short(__xdata uint16_t *vlan, uint8_t idx)
+// return 0 on error or non-zero number of number-byte taken for the conversion.
+// Stops at any non-digit '0'-'9' char or bytes is more then 5.
+uint8_t atoi_short(uint8_t idx)
 {
-	uint8_t err = 1;
-	*vlan = 0;
+	uint8_t cnt = 0;
+	atoi_results_short = 0;
 
-	while (isnumber(cmd_buffer[idx])) {
-		err = 0;
-		uint8_t val = cmd_buffer[idx] - '0';
-		if (*vlan > 6553 || (*vlan == 6553 && val > 5))
-			return 1;
-		*vlan = (*vlan * 10) + val;
-		idx++;
+	uint8_t *ptr = &cmd_buffer[idx];
+
+	while (1) {
+		uint8_t val = *ptr++ - '0';
+		if (val > 9)
+			break;
+		if (atoi_results_short > 6553 || (atoi_results_short == 6553 && val > 5) || cnt >= 5)
+			return 0;
+		atoi_results_short = (atoi_results_short * 10) + val;
+		cnt++;
 	}
 
-	return err;
+	return cnt;
 }
 
 
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -269,6 +269,34 @@ uint8_t cmd_parse_port_separator(uint8_t idx) {
 	return ret;
 }
 
+// Same as cmd_parse_port_separator() addition to allow the CPU-port.
+// returns 0 when on parser error or invalid value or no space or no NUL.
+// return non-zero number of bytes consumed including the space.
+uint8_t cmd_parse_port_cpu_separator(uint8_t idx) {
+	uint8_t ret = atoi_byte(idx);
+	// Check valid con
+	if (ret != 0) {
+		// port number 1-10 -> 0-9
+		atoi_results_u8 -= 1;
+		if(atoi_results_u8 < CPU_PORT)
+			// Phy port, translate it.
+			ret = cmd_parse_port(idx);
+		// Check is not
+		else if (atoi_results_u8 > CPU_PORT)
+			ret = 0;
+	}
+
+	if (ret != 0) {
+		idx += ret;
+		uint8_t c = cmd_buffer[idx];
+		if (c == ' ') {
+			ret++;
+		} else if (c != NUL)
+			ret = 0;
+	}
+	return ret;
+}
+
 
 // check if the cmd_buffer[idx] is a space.
 __bit cmd_is_space(uint8_t idx) {
@@ -562,7 +590,7 @@ void parse_isolate(void)
 
 	uint8_t w = 2;
 	while (w < cmd_words_len) {
-		if (cmd_parse_port_separator(cmd_words_b[w++]) == 0)
+		if (cmd_parse_port_cpu_separator(cmd_words_b[w++]) == 0)
 			goto err;
 		uint8_t port = atoi_results_u8;
 		members |= ((uint16_t)1) << port;
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -283,6 +283,12 @@ __bit cmd_is_space(uint8_t idx) {
 	return cmd_buffer[idx] == ' ';
 }
 
+// check if the cmd_buffer[idx] is a space or null.
+__bit cmd_is_space_or_null(uint8_t idx) {
+	uint8_t c = cmd_buffer[idx];
+	return c == ' ' || c == '\0';
+}
+
 // returns 0 when on parser error or invalid value or no space.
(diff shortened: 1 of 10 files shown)


DrDoof:
Happy to. I ran the new helpers rather than only reading them, compiled straight out of this branch, so the numbers below are the code's and not mine.

The part I would lead with is the one the diff hides. Today the same port argument is turned into a logical port at about a dozen places, each writing its own check, and two of them get it wrong: `parse_isolate` indexes `phys_to_log_port[]` with `cmd_buffer[...] - '1'` before checking anything at all, so `isolate 0` reads at index -1, and `parse_ingress` guards with `p - '1' > 9`, which promotes to int and so catches the high side but not the low one, giving the same read for `ingress 0u`. `parse_mtu` and the `pvid` branch do it properly with a range test on the character. Routing all of them through `cmd_parse_port` leaves exactly one indexing site, reached only after `atoi_byte` succeeded and under `port < 9`. That is worth more than the line count suggests.

`atoi_byte` and `atoi_short` come out clean. I walked `atoi_short` over 0 to 69999 and `atoi_byte` over 0 to 999: every value inside the type round-trips, everything past it is rejected, and the edges behave (`""`, `"abc"`, `"+5"`, `" 12"` rejected, `"007"` gives 7 over three bytes, `"12x"` gives 12 over two).

Two of the other helpers return something other than what their comment promises. Neither breaks a caller today, which is why I would rather raise them now than let the next caller find them.

`parse_ip` ends on `return idx_start - idx`, but `idx` has already moved past what it parsed, so the value is 256 minus the byte count instead of the count: `1.2.3.4` returns 249 and `192.168.100.200` returns 241. All four call sites only test it against zero, and the validation itself is right, with `256.1.1.1`, `1.2.3` and `1.2.3.4.5` all rejected. It is the sign that is wrong.

`cmd_parse_port_space` checks the trailing space only when the port parse has already failed, since the body sits under `if (!ret)`. On success the space is never looked at and the count does not include it; on failure a space at that index turns the failure into a success. Measured here on a six port machine: `"1 "` returns 1 rather than 2, `"1"` with no space is accepted, and `" "` alone is accepted with `atoi_results_u8` left at 0.

The callers are safe from the second half of that, because all eight pass a word start, which cannot be a space. The first half is load bearing, and this is the bit I would not want you to hit later: of the eight, only `parse_ingress` uses the returned count. It takes `[p]<u/t/a>` as one word and does `idx += ret` before reading the mode letter, so it works precisely because the space is not consumed. Correct the condition on its own and `ingress 1u` starts answering "invalid ingress mode". The two changes want to land together.

One behaviour change worth calling out in the description, since a refactor is not where a reader looks for it: every call passes `cpu_is_valid` as false, so `isolate` stops accepting the CPU port. Main accepts it there, which is the subject of #334. I think refusing it is right, it just deserves a line.

Something that looks like a bug and is not: the digit counter in both atoi helpers is a `uint8_t` and wraps at 256 digits, but `CMD_BUF_SIZE` is 128, so it cannot be reached.

For the syslog PR, yes, `atoi_short` is doing exactly what that hand rolled loop does, and I will move #333 onto it once this lands. I would rather not build on an unmerged PR before then.

Scope, so you know what this is worth: I ran `atoi_byte`, `atoi_short`, `parse_ip`, `cmd_parse_port` and `cmd_parse_port_space`, read `print_ip`, and went through every call site of the two port helpers. I have not re-derived the unrelated conversions one by one. `atoi_hex` I left alone as it is unchanged here, though `c - '0' < 10` on line 158 promotes to int and so accepts anything below '0' as a digit, and `h_idx & 1 == 1` on line 173 parses as `h_idx & (1 == 1)` and works by luck. Both are in main already, so a separate matter.

vDorst:
> atoi_byte and atoi_short come out clean. I walked atoi_short over 0 to 69999 and atoi_byte over 0 to 999: every value inside the type round-trips, everything past it is rejected, and the edges behave ("", "abc", "+5", " 12" rejected, "007" gives 7 over three bytes, "12x" gives 12 over two).

I added extra length check so the function stops at 3 or 5 bytes. So `0000000000000000000000` is not valid anymore.

> parse_ip ends on return idx_start - idx, but idx has already moved past what it parsed, so the value is 256 minus the byte count instead of the count: 1.2.3.4 returns 249 and 192.168.100.200 returns 241. All four call sites only test it against zero, and the validation itself is right, with 256.1.1.1, 1.2.3 and 1.2.3.4.5 all rejected. It is the sign that is wrong.

This is fixed, value have to be swapped.

> cmd_parse_port_space checks the trailing space only when the port parse has already failed, since the body sits under if (!ret). On success the space is never looked at and the count does not include it; on failure a space at that index turns the failure into a success. Measured here on a six port machine: "1 " returns 1 rather than 2, "1" with no space is accepted, and " " alone is accepted with atoi_results_u8 left at 0.

Fixed, `!ret` should be `ret == 0`.

> The callers are safe from the second half of that, because all eight pass a word start, which cannot be a space. The first half is load bearing, and this is the bit I would not want you to hit later: of the eight, only parse_ingress uses the returned count. It takes [p]<u/t/a> as one word and does idx += ret before reading the mode letter, so it works precisely because the space is not consumed. Correct the condition on its own and ingress 1u starts answering "invalid ingress mode". The two changes want to land together.

ingress is fixed.

DrDoof (commented):
The three you listed all check out, and `parse_ingress` came out better than it was: taking `cmd_parse_port` and testing for exactly one digit, then requiring a space or end of word after the mode letter, means `ingress 1uX` is refused now, which it was not before.

`parse_ip` returns the count, `1.2.3.4` gives 7 and `192.168.100.200` gives 15, each equal to the string length, with the rejects unchanged. `cmd_parse_port_space` behaves: `"1 "` accepted, `"1"` without the space not, a bare space not, `"1u"` not. The length caps work, `"12345"` is fine and `"123456"` is not, and the long run of zeros is gone. One edge they also cover: `atoi_byte("0000")` is now refused, four digits being past the cap even though the value fits. I think that is the right trade, just noting it in case a config file somewhere pads a port number.

The commit on top of those breaks `isolate` though, and I left the details on the line. Everything I say below I ran out of the branch rather than read.

cmd_parser.c
```diff
@@ -452,22 +557,14 @@ void parse_isolate(void)
...
-				port = machine.phys_to_log_port[port];
-				if (port < machine.min_port || port > machine.max_port)
-					goto err;
-			}
-			members |= ((uint16_t)1) << port;
-		}
-		w++;
+		uint8_t idx = cmd_words_b[w++];
+		uint8_t ret = cmd_parse_port(idx, false);
+		if (ret == 0 || !cmd_is_space_or_null(idx))
```
DrDoof: This always jumps to `err`. `cmd_is_space_or_null(idx)` looks at the first character of the word, which for a valid port is the digit itself, so it is never a space or a NUL.

Running the loop as it stands, `isolate 1 2 3`, `isolate 1 5` and `isolate 1 2 3 4` are all refused; with `idx + ret` all three are accepted again. The `show` and `off` forms are handled before the loop, so only the member list is affected.

`parse_ingress` gets this right and is worth comparing against: there the `idx++` happens while reading the mode letter, so by the time the check runs it is already looking past it. I tested that one the same way, `1u`, `6t` and `1a` pass while `1uX` and a bare `1` are refused.

```suggestion
		if (ret == 0 || !cmd_is_space_or_null(idx + ret))
```

cmd_parser.c
```diff
@@ -185,66 +188,167 @@ uint8_t atoi_hex(uint8_t idx)
...
 
+// Same as cmd_parse_port() addition to check the trailing space.
+// returns 0 when on parser error or invalid value or no space.
+// return non-zero number of bytes consumed including the space.
+uint8_t cmd_parse_port_space(uint8_t idx, __bit cpu_is_valid) {
+	uint8_t ret = cmd_parse_port(idx, cpu_is_valid);
+	if (ret != 0) {
+		ret += idx;
+		if (cmd_buffer[ret++] != ' ')
+			ret = 0;
```
DrDoof: Left over from the previous round and harmless today, but the code and the comment above still disagree. `ret += idx` makes this return an absolute offset past the space rather than the number of bytes consumed. It only coincides at index zero, which is why it looks right in isolation: with the word at index 8 it returns 10, at 16 it returns 18. All seven call sites test it against zero and nothing else, so nothing is wrong right now.

I ran the suggestion below against the current form over ten inputs at three offsets, covering a good port with and without the trailing space, an out of range one, a zero, a non digit, `1u` and a bare space. The accept or reject decision is identical everywhere and the returned value stops moving with the offset.

```suggestion
		if (cmd_buffer[idx + ret] != ' ')
			return 0;
		ret++;
```

cmd_parser.c
```diff
@@ -411,19 +519,16 @@ void parse_vlan(void)
 
 void parse_isolate(void)
 {
-	__xdata uint16_t members = 0;
+	uint16_t members = 0;
```
DrDoof, line 550: A question rather than a finding: this used to be an `__xdata` local. Internal RAM is the thing this image runs out of first and the rest of the file keeps scratch in xdata for that reason, so I wanted to flag it in case dropping it here was incidental rather than deliberate.

cmd_parser.c
vDorst: Well spotted!

cmd_parser.c
vDorst: Storing temporary data in xdata is only useful in case the compiler has no room left and puts it permanently in sram.
But that is not the case there, I haven't the the SRAM layout changed, after this change.
And load 16-bit value from xdata is also not cheap. It needs to be store somewhere while manipulating the variable.

cmd_parser.c
vDorst: Well spotted, is fixed.

cmd_parser.c
vDorst: After #330 is merged, it is easier to write code that first puts in non-xdata memory.
And see if the compiler needs some help to put some variables in xdata.

vDorst: > The three you listed all check out, and parse_ingress came out better than it was: taking cmd_parse_port and testing for exactly one digit, then requiring a space or end of word after the mode letter, means ingress 1uX is refused now, which it was not before.

I see more placed that we can improve this kind of checking, see if there is no extra data after the wanted input.

DrDoof:
Rebase looks right and having the bounds check live in one helper is the good part. Four things from reading it.

`pvid` now rejects the first port. The condition is `cmd_parse_port_separator(cmd_words_b[1]) != 0 && atoi_results_u8 && ...`, and `atoi_results_u8` is the logical port, so the truthiness test drops logical port 0. On every board with `min_port` 0 the first physical port maps to logical 0: SWTGW218AS, SWTG018AS-A, the three KP-9000-9X variants and the 8+1 SFP board. `pvid 1 10` answers with the usage line instead of setting the PVID. main had no such test, it passed the logical port straight to `port_pvid_set()`.

`parse_lag_hash()` passes a character where the helper wants an index:

```c
if (cmd_parse_port_separator(cmd_buffer[cmd_words_b[1]]) == 0)
```

The other eleven call sites pass `cmd_words_b[...]`. This one reads `cmd_buffer[(char)1]`, that is `cmd_buffer[49]`, which is zero on a fresh buffer, so `laghash` just errors out. Both sides are `uint8_t`, so nothing warns about it. A byte above 127 in that position indexes past the 128 byte buffer; the console filters input to printable ASCII, but `execute_commands()` does not, so the HTTP path and a replayed config can reach it.

The same two lines also run a LAG group number through the port helper, which translates it through `phys_to_log_port[]`. `parse_lag()` next door gets this right with `atoi_byte(cmd_words_b[1])`. Once the line above is fixed, on an 8+1 board `laghash 1` yields logical port 0 and then group 255 and is rejected, while `laghash 3` lands on group 1.

The last one is a question rather than a defect. With the manual CPU form gone, nothing sets bit 9 on the explicit member path, while `off` still does `members |= 0x200`. So `isolate 3 4` takes the management path away from port 3, and there is no longer any way to keep it. Is that the automatic handling you described in #334, still to come, or is it the intended policy? If it is policy, the usage line is the place to say so.

Not from this PR, but you are rewriting the line anyway: `parse_bw()` guards on `cmd_words_len < 2` and then reads `cmd_words_b[2]`, which needs three words.

logicog:
I went over it at a broader scale and I really like the concept, this will be a big clean-up. Need to go through it in detail, though.

vDorst:
Rebase to latest main.

> pvid now rejects the first port...

Fixed, also improve the pvid parsing, so error show when the line start with `pvid `.

> Fix parse_lag_hash()

We are pasing lag numbers not ports.
Also improved the usage message.

> Fix and improved parse_bw()

Fix your suggestion and also refactor it

Also added a `print_port` helper to print the phy port of a log port number.

Update main port about code size.


> The last one is a question rather than a defect. With the manual CPU form gone, nothing sets bit 9 on the explicit member path, while off still does members |= 0x200. So isolate 3 4 takes the management path away from port 3, and there is no longer any way to keep it. Is that the automatic handling you described in #334, still to come, or is it the intended policy? If it is policy, the usage line is the place to say so.

I think this is a good example were we can't tell if the cpu-port is needed or not without additional information/rules we currently don't have.
Rules like port type eg. 'USER', 'UPSTREAM', 'ADMIN'.
So `isolate` may need also able to add the CPU-port as a argument.

vDorst:
Ready for a new review.
`isolate` can add CPU-port as member again.

DrDoof:
All four are fixed, and I checked them in the code rather than taking the commit messages for it.

`pvid` accepts the first port again. I also went back over `atoi_short()`, since you changed it in this series, to be sure it still only writes `atoi_results_short` and cannot clobber the port between the check and `port_pvid_set()`. It does not.

`parse_lag_hash()` now goes through `atoi_byte(cmd_words_b[1])`, the same way `parse_lag()` next door does, so both the index and the semantics are right. `laghash 1` lands on group 0, `laghash 4` on group 3, and `0` and `5` are rejected.

`parse_bw()` I read end to end this time rather than from the hunk. The `cmd_words_len < 4` check before `cmd_words_b[3]` is there, so the fix is complete and not just moved one word along, which is what I was expecting to find.

The CPU port split is the right shape: the configured port still goes through `cmd_parse_port_separator()` and stays front panel only, while the member list uses the new helper. I ran that helper through its edge cases: `1` to `9` translate through `phys_to_log_port[]`, `10` yields logical 9 so the mask bit is exactly the `0x200` the `off` branch sets, and `0`, `11`, `100`, `07` and a trailing non separator are all rejected. Your reasoning is better than my question was, incidentally. Without port types the firmware has no way to decide when the CPU port belongs in the mask, so leaving it to the operator is the honest answer for now.

On the size claim, I built current main and this branch with the same toolchain for SWTGW218AS. It is better than you wrote: BANK1 41575 to 41474, BANK2 38354 to 37055, so 1341 bytes of code and 5 bytes of xdata.

One nit left, in the new helper: `// Check valid con` and `// Check is not` look like they were meant to say more.

vDorst:
- Fix the comment
- rename `print_port()` to `print_phys_port()` to make it more clear that it is printing the physical port.
- Refactor some code.

cmd_parser.c
```diff
@@ -111,9 +113,9 @@ uint8_t cmd_compare(uint8_t start, __code uint8_t * cmd)
 		uint8_t c = cmd[j];
 		uint8_t b = cmd_buffer[i];
 
-		// cmd is garanteerd to be NULL-terminated.
-        if (c == '\0') {
-            if ((b == ' ') || (b == '\0')) {
+		// cmd is garanteerd to be NUL-terminated.
```
logicog: typo: guaranteed.

cmd_parser.c
```diff
@@ -185,61 +187,182 @@ uint8_t atoi_hex(uint8_t idx)
 }
 
 
-uint8_t atoi_byte(__xdata uint8_t *out, uint8_t idx)
+// return 0 on error or non-zero number of number-byte taken for the conversion.
+// Stops at any non-digit '0'-'9' char or bytes is more then 3.
```
logicog: "bytes is more then 3" -> "or more than 3 bytes."

cmd_parser.c
```diff
@@ -185,61 +187,182 @@ uint8_t atoi_hex(uint8_t idx)
...
+uint8_t atoi_byte(uint8_t idx)
 {
-	uint8_t err = 1;
+	uint8_t cnt = 0;
 	uint8_t num = 0;
```
logicog, line 195: Do you still need num, if you return the result anyway in atoi_results_u8? Could you not directly work with that xdata byte?

cmd_parser.c
```diff
@@ -185,61 +187,182 @@ uint8_t atoi_hex(uint8_t idx)
 }
 
 
-uint8_t atoi_byte(__xdata uint8_t *out, uint8_t idx)
+// return 0 on error or non-zero number of number-byte taken for the conversion.
```
logicog: returns 0 or the number of digits taken into account for conversion.

cmd_parser.c
```diff
@@ -185,61 +187,182 @@ uint8_t atoi_hex(uint8_t idx)
...
 
-	*out = num;
-	return err;
+	atoi_results_u8 = num;
+	return cnt;
 }
 
 
-uint8_t atoi_short(__xdata uint16_t *vlan, uint8_t idx)
+// return 0 on error or non-zero number of number-byte taken for the conversion.
```
logicog: See comment above.

cmd_parser.c
```diff
@@ -185,61 +187,182 @@ uint8_t atoi_hex(uint8_t idx)
...
 	}
 
-	return err;
+	return cnt;
+}
+
+/* Parse, validate and translate phys_to_log_port physical port argument.
+ * So CPU-port 10 is not valid.
+ * returns 0 when on parser error or invalid value.
+ * return non-zero number of bytes consumed.
```
logicog: "bytes consumed" -> "characters consumed" is clearer

cmd_parser.c
```diff
@@ -185,61 +187,182 @@ uint8_t atoi_hex(uint8_t idx)
...
+		atoi_results_short = (atoi_results_short * 10) + val;
+		cnt++;
 	}
 
-	return err;
+	return cnt;
+}
+
+/* Parse, validate and translate phys_to_log_port physical port argument.
+ * So CPU-port 10 is not valid.
```
logicog: "The CPU-port, i.e. port 10, is not a valid argument."

cmd_parser.c
```diff
@@ -185,61 +187,182 @@ uint8_t atoi_hex(uint8_t idx)
...
+		return 0;
+
+	atoi_results_u8 = port;
+	return 1;
+}
+
+
+// Same as cmd_parse_port() addition to check the trailing SPACE or NUL.
+// returns 0 when on parser error or invalid value or no SPACE or no NUL.
+// return non-zero number of bytes consumed including the SPACE.
```
logicog: number of characters consumed

cmd_parser.c
```diff
@@ -185,61 +187,182 @@ uint8_t atoi_hex(uint8_t idx)
...
+		uint8_t c = cmd_buffer[idx];
+		if (c == ' ') {
+			ret++;
+		} else if (c != NUL)
+			ret = 0;
+	}
+	return ret;
+}
+
+// Same as cmd_parse_port_separator() addition to allow the CPU-port.
```
logicog: addition to allow the CPU-port -> but additionally allow CPU-port.

cmd_parser.c
```diff
@@ -185,61 +187,182 @@ uint8_t atoi_hex(uint8_t idx)
...
+// returns 0 when on parser error or invalid value or no SPACE or no NUL.
+// return non-zero number of bytes consumed including the SPACE.
```
logicog: number of characters

cmd_parser.c
```diff
@@ -185,61 +187,182 @@ uint8_t atoi_hex(uint8_t idx)
...
+	port = machine.phys_to_log_port[port];
+	if (port < machine.min_port || port > machine.max_port)
+		return 0;
+
+	atoi_results_u8 = port;
+	return 1;
+}
+
+
+// Same as cmd_parse_port() addition to check the trailing SPACE or NUL.
```
logicog: addition to check the trailing SPACE or NUL. -> but additionally check for trailing SPACE or NUL.

cmd_parser.c
```diff
@@ -185,61 +187,182 @@ uint8_t atoi_hex(uint8_t idx)
...
+	return cmd_buffer[idx] == ' ';
 }
 
+// check if the cmd_buffer[idx] is a SPACE or NUL.
+__bit cmd_is_space_or_nul(uint8_t idx) {
+	uint8_t c = cmd_buffer[idx];
+	return c == ' ' || c == NUL;
+}
 
+// returns 0 when on parser error or invalid value or no SPACE.
```
logicog: Start with
// Parses and IP address
....

cmd_parser.c
```diff
@@ -185,61 +187,182 @@ uint8_t atoi_hex(uint8_t idx)
...
+			goto err;
 	}
+	return idx - idx_start;
+
+err:
+	print_string("Error in IP format\n");
 	return 0;
 }
 
+// Print a IPv4 adress.
```
logicog: a -> an

logicog (commented):
Really nice code clean-up! Looks much more readable, now.

cmd_parser.c
vDorst: I tested both and saw no code-size difference.
Idea was, first process it in the register than push it to xdata.
So you don't have a lot of xdata reads and writes.
But I think the compiler is smart enough to already keep the data in a register and do only one xdata write.

So I keep it as is.

vDorst:
I think I have fixes all the comments.

I am still thinking what if we use physical port `0` instead of `10` for the CPU-port.
This makes it easier to remembers and also easier to parse.

logicog:
Using 0 for the CPU port is a really good idea. I had been thinking about c or C, or CPU, but it would have complicated the code a lot. Could you make 0 the CPU-port?

vDorst:
Change seems easy, I have to look in to other parts of the code, I think mainly http-part, which may need to change too.

On Sunday, I'll have time to work on it again.

vDorst:
Now the CLI code treats port zero as the CPU port.

But I think we need to define what a `port` is in some context.
`send_counter()` also expects a port number. Port value is index into the `machine.phys_to_log_port` not the physical port number.

feelfree69:
I really like this refactoring! 

What is still confusing me is the "doubled" definition of ports, with logical and physical view. 
It would be much clearer if the vars and args are named according to their actual function, e.g. `log_port`or `phys_port` instead of just `port`. But such changes can be added anytime if considered useful, they don't need to be part of this PR.

logicog:
> I really like this refactoring!
> 
> What is still confusing me is the "doubled" definition of ports, with logical and physical view. It would be much clearer if the vars and args are named according to their actual function, e.g. `log_port`or `phys_port` instead of just `port`. But such changes can be added anytime if considered useful, they don't need to be part of this PR.

It is not as bad as it sounds: all internal workings are with the logical ports, the physical ones are the ones that happen on the CLI and Web-Interface. That's because the switch manufacturer can place the ports in any sequence they want at any place they want on the device. It is the same with, say OpenWRT, where the port numbers of the switch have nothing to do with where the actual ports are located on the device. However, there they use a device tree to make the mapping and the user only sees the labels given in the DTS, like "lan3". The original reason to use port for both types of ports was to save registers. I do not know if the compiler is clever enough to understand that once a physical port has been converted to a logical one, the variable space or register can be re-used.

vDorst:
> It is not as bad as it sounds: all internal workings are with the logical ports, the physical ones are the ones that happen on the CLI and Web-Interface.

In case of `send_counter()` physical port 1 is requested as `port=0`.
Which is used as the index into `machine.phys_to_log_port`.
So there is already a discrepancy.

> I do not know if the compiler is clever enough to understand that once a physical port has been converted to a logical one, the variable space or register can be re-used.

AFAIK that is happening.

feelfree69:
> In case of `send_counter()` physical port 1 is requested as `port=0`.
> Which is used as the index into `machine.phys_to_log_port`.
> So there is already a discrepancy.

Then why not change the logic to transfer really the **physical** port, and change the prototype to `void send_counters(char phys_port)`. It is called at only once from `page_impl.c`. 
Ok, it seems a bit crazy to add a `+1` at the calling function and add a `-1` at the called function, but at least it would restore consistency.

vDorst:
Good idea, I changed it.

Please review.

httpd/page_impl.c
```diff
@@ -310,11 +310,12 @@ void send_vlan(uint16_t vlan)
 }
 
 /* Send counters
- * Only accepts physical port idx to 0-8.
+ * Only accepts physical port 1..8.
```
feelfree69: 1..9

feelfree69 (commented):
only one small comment issue

vDorst:
Fixed.

logicog:
Oops, wanted to merge this, but because of the merge of the SFP -fixes, there are now merge-conflicts. Can you check?

vDorst:
Rebased.

logicog:
Merged. Really great cleanup!

PR #327, opened 2026-08-12
Bound the ASIC table waits and act on the result

DrDoof:
Seventeen spots across three modules poll an ASIC engine for a table operation to
finish, and if the engine never clears its execute bit they spin for ever. Nothing
else runs on the 8051 to notice, so management disappears while the ASIC keeps
forwarding, which is a miserable thing to diagnose from the outside.

The review on #314 asked for these to be rethought, and that objection was right. A
bound on its own is the dangerous half. The code gives up and then carries on as
if the operation had finished, reading a data register that was never filled or
reporting a delete that may never have happened. That is how I bricked this switch
twice in August, and the way back was an SPI clip.

So fifteen of them now report whether the engine actually went idle, and every
caller does something honest with that. vlan_get returns the -1 it already has.
vlan_create and port_l2_learned return without printing a result. port_l2_forget
skips its done line. The walks over the L2 table break out rather than read an
entry the engine never produced. In the web server the response is half written by
the time the wait fails, so send_l2 closes its array and drops the separator it had
already emitted, l2_delete answers 0 exactly as it does for an entry it cannot
find, and the SFP read leaves its hex field empty instead of filling it from an
output register the controller never wrote, which would come back looking like a
plausible temperature.

One caller does not benefit yet, and it is worth naming here rather than leaving it
to turn up in review: send_vlan ignores the -1 vlan_get hands back, as it already ignores
the one for an out of range id, so it renders whatever the last read left behind.
That fault is older than this and belongs with that function.

The remaining two waits are deliberately left unbounded, both in vlan_setup. That
one runs from the boot sequence before anything else is up, and a switch that comes up with half
a VLAN table written is worse than one that visibly never finishes starting. There
is nothing there to fall back to, so hanging where you can see it is the better
failure.

I measured the worst case on hardware rather than guessing at it. Real silicon will
not wedge its table engine on request, so what I timed was a loop of the same shape
driven to exhaustion on the switch: 65536 register reads cost 0.23 s, about 3.5 us
each, four runs with a spread under 7 ms. So the pathological case blocks management for a quarter of a
second instead of for ever. The same number says why 8 bits was not enough. It
would have given up after 0.9 ms, which sits inside ordinary contention for the
table, and giving up early is exactly what hurts.

The premise turned up on its own while I was testing. Management died on me for an
unrelated reason, the one in #298, and the switch carried on switching without a
flicker: traffic through it was untouched the whole time the management CPU was
gone and the switch would not answer a ping.

The healthy path I exercised on the switch as well, since a guard that broke
ordinary operation would be worse than no guard. The L2 and VLAN pages answer
exactly what they answered before, vlan_create and port_l2_forget both print their
done line, the flush empties the table, and port_l2_learned walks it and prints
what is there.

The giving up itself I could only test off the switch. There are host tests for
every one of these fifteen spots, using the harness in #324, with an engine that
can be told to wedge after a chosen number of successful polls. Run against the code as it is today all three suites trip that harness's
watchdog, which is what makes them worth anything. They are not in this PR because
#324 has not landed yet; I will send them separately once it does.

Three commits, one per module, separable if you want to take them at different
speeds. Sizes against main with both trees built from the same version string:
BANK1 grows by 309 bytes, BANK2 is unchanged, and image xdata goes from 10207 to
10213. Each helper's result costs one bit of BSEG, which fits inside the two bytes
of bit space the image already reserves, so DSEG, OSEG and the stack base are
untouched.

### DIFF
--- a/httpd/page_impl.c
+++ b/httpd/page_impl.c
@@ -46,6 +46,18 @@ extern __xdata char sfp_module_model[2][17];
 extern __xdata char sfp_module_serial[2][17];
 extern __xdata uint8_t sfp_options[2];
 
+static __xdata uint16_t hw_guard;
+
+static uint8_t page_tbl_idle(void)
+{
+	hw_guard = 0;
+	do {
+		reg_read_m(RTL837X_TBL_CTRL);
+	} while ((sfr_data[3] & TBL_EXECUTE) && ++hw_guard);
+
+	return !(sfr_data[3] & TBL_EXECUTE);
+}
+
 __code uint8_t * __code HTTP_RESPONCE_JSON = "HTTP/1.1 200 OK\r\nConnection: close\r\nContent-Type: application/json\r\n\r\n";
 __code uint8_t * __code HTTP_RESPONCE_TXT = "HTTP/1.1 200 OK\r\nContent-Type: text/plain\r\n\r\n";
 
@@ -210,9 +222,12 @@ void sfp_send_data(uint8_t slot, uint8_t reg, uint8_t len)
 	reg_bit_set(RTL837X_REG_I2C_CTRL, 0);
 
 	// Wait for execution to finish
+	hw_guard = 0;
 	do {
 		reg_read_m(RTL837X_REG_I2C_CTRL);
-	} while (sfr_data[3] & 0x1);
+	} while ((sfr_data[3] & 0x1) && ++hw_guard);
+	if (sfr_data[3] & 0x1)
+		return;
 
 	for (uint8_t i = 0; i < len; i++) {
 		if (!(i & 0x3))
@@ -338,9 +353,11 @@ void send_l2(uint16_t idx)
 	dbg_short(idx);
 	__xdata uint8_t entries_left = L2_MAX_TRANSFER;
 
-	do {
-		reg_read_m(RTL837X_TBL_CTRL);
-	} while (sfr_data[3] & TBL_EXECUTE);
+	if (!page_tbl_idle()) {
+		char_to_html('[');
+		char_to_html(']');
+		return;
+	}
 
 	/* The L2 table in the ASIC can hold up to 4096 (0x1000) entries, which
 	 * are accessed using an index. The index is the hash of the MAC address
@@ -364,9 +381,12 @@ void send_l2(uint16_t idx)
 		REG_WRITE(RTL837x_TBL_DATA_0, sfr_data[0], sfr_data[1] & 0xfc, sfr_data[2] | (TBL_LUTREAD_NEXT_L2UC << 6), sfr_data[3]);
 
 		REG_WRITE(RTL837X_TBL_CTRL, entry >> 8, entry, TBL_L2_UNICAST, TBL_EXECUTE);
-		do {
-			reg_read_m(RTL837X_TBL_CTRL);
-		} while (sfr_data[3] & TBL_EXECUTE);
+		if (!page_tbl_idle()) {
+			if (first_entry != 0xffff)
+				slen--;
+			char_to_html(']');
+			break;
+		}
 
 		reg_read_m(RTL837x_L2_DATA_OUT_B);
 		if ((sfr_data[0] & 0x20)) {	// Check entry is valid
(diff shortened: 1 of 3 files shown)


vDorst:
`page_tbl_idle()`, `igmp_tbl_idle()` and `port_tbl_idle()` are the same.
I propose to just use one and move it to `rtlplayground.c`.
Only only one globe countdown-variable for all the location that we need a `timeout`-feature for accessing the SOC.

Return value can be a `bool` or `__bit`.
Add comment what the return value is representing.

httpd/page_impl.c
```diff
@@ -464,11 +487,7 @@ void l2_delete(uint16_t idx)
 		REG_WRITE(RTL837x_TBL_DATA_0, sfr_data[0], sfr_data[1], TBL_L2_UNICAST, sfr_data[3]);
 
 		REG_WRITE(RTL837X_TBL_CTRL, idx >> 8, idx, TBL_L2_UNICAST, TBL_WRITE | TBL_EXECUTE);
-		do {
-			reg_read_m(RTL837X_TBL_CTRL);
-		} while (sfr_data[3] & TBL_EXECUTE);
-
-		char_to_html('1');
+		char_to_html(page_tbl_idle() ? '1' : '0');
```
vDorst, line 490: bool_to_html() can be used.

vDorst:
Maybe also useful to inside `page_tbl_idle()` print a warning in case of a timeout.
So we can atleast spot when it happens.

logicog:
Please investigate this further. This sounds like a very dangerous idea. There is no guarantee that the table operations have any kind of time constraint, so that this can be done even in 65535 re-tries is just an assumption. What would we do if the the wait times out? Probably the ASIC part of the switch is dead anyway at that point and would need a reboot. I have not encountered a single case where a table operation did not succeed. So trying to solve a problem that probably does not exist by something that could be very dangerous to do, does not sound like a good idea to me. 

The first thing would be to identify a reproducible situation in which table writes do not finish. If this does not exist or can be avoided otherwise, then we leave it at that, or avoid it. If they can not be avoided, the next thing is to understand how to recover. There are options to reset the ASIC. Not sure that this works. Also not sure how to reconfigure the ASIC. We could reset the switch, but maybe this does not work properly on the ASIC. Before we go down this road, please give an example of when reproducibly a table operation gets stuck that cannot be avoided by proper programming in the first place.

DrDoof:
I went looking for one and I can't produce it.

A control first, otherwise a negative result means nothing. Driving the engine by hand really does reach it: `regset 5cac 00800401` asks for L2 entry 0x0080 and the data registers come back with b8:27:eb:4d:27:e9 on VLAN 2 with the valid bit set, which is a real learned entry. So the commands below did run.

Then everything I could think of to jam it, reading `TBL_CTRL` back each time:

- reads with table type 0x0f which doesn't exist, with type 0, with entry 0xffff on a 4096 entry table, with entry 0x0fff and read method 5
- twelve reads issued back to back inside one command batch, microseconds apart with nothing waiting between them
- a table write, a write with a bogus table type, a write with an out of range entry
- the flush engine at 0x53d4 with a port mask, with an empty mask, and with all ones
- that twelve read burst again while the switch was busy learning MACs from a 254 address ping sweep

Execute was already clear every time I looked. The hardware quietly masks what it doesn't implement, too: 0xffff comes back as 0x1fff, 0x0f as 0x07, the flush mask 0xffff as 0x03ff. That is what an engine that always finishes looks like.

The structural part bothers me more than any of that. Every one of these guards is written as

```c
do {
	reg_read_m(RTL837X_TBL_CTRL);
} while ((sfr_data[3] & TBL_EXECUTE) && ++tbl_guard);
```

and `reg_read_m()` ends in `do { } while (SFR_EXEC_STATUS != 0);` with no bound at all. There are 14 of those in the tree and every guarded loop sits on one. So I'm bounding the outer loop while the inner one can still hang forever. It buys something only in the narrow case where TBL_EXECUTE sticks and the register bus keeps answering, and if the ASIC is really gone we hang a level below every guard I added, which is roughly where you said we would end up.

So it doesn't do what the title says, and I'm closing it. Thanks for pushing back rather than merging it. If anyone does want bounded ASIC access later, it belongs in `reg_read_m` and `reg_write_m`, not in their callers.

Separately, there is one thing in the branch I would still like to offer, and it sits closer to the proper programming you asked for than to guards. `handle_rx()` dispatches to `igmp_packet_handler()` on the destination address alone, and the handler checks only the IP protocol and the message type before it writes a table entry. Two lines above, the LACP branch reads `lacpEnabled && uip_buf[5] == 0x02`. There is no `igmpEnabled` anywhere, so an ordinary IGMPv3 report off the wire writes to the table for a feature that might well be switched off. Say the word and I will send that on its own.

logicog:
> Separately, there is one thing in the branch I would still like to offer, and it sits closer to the proper programming you asked for than to guards. `handle_rx()` dispatches to `igmp_packet_handler()` on the destination address alone, and the handler checks only the IP protocol and the message type before it writes a table entry. Two lines above, the LACP branch reads `lacpEnabled && uip_buf[5] == 0x02`. There is no `igmpEnabled` anywhere, so an ordinary IGMPv3 report off the wire writes to the table for a feature that might well be switched off. Say the word and I will send that on its own.

That sounds like a good idea. Go ahead, please.

DrDoof:
Sent as #332, with the note about the parse being off four bytes included there rather than repeated here.

PR #329, opened 2026-08-12
httpd: answer /cmd with what the command printed, and show it in the console tab

DrDoof:
Two small commits that belong together. Either one on its own doesn't get you anything.

`POST /cmd` runs the command and answers with an empty 200, so the caller can't tell whether it worked, was refused, or was just misspelt, and whatever the command printed only reaches the serial console. Reading a register over the network means finding a second channel to see the answer, which is how I ended up watching syslog to check a value I had written a moment earlier.

`write_char()` is the one funnel every console helper goes through, and it already tees into the syslog ring. A third sink, live only while the httpd is running a command that arrived over the network, drops the same bytes into the response behind the header. Serial and syslog keep their copy on purpose: I nearly suppressed them, then remembered I had been leaning on syslog for exactly this an hour before. The capture stops at the end of the response buffer, so a long listing is truncated rather than running past it.

The System page already has a Console tab, but `cmdSub()` threw the response away, and until now there was nothing in it to keep. It holds a transcript now: the command after a prompt, then whatever it printed. The box is a `pre` filled through `textContent`, so nothing the device prints gets taken for markup.

39 bytes of BANK1 and 52 of the common area, one byte of xdata, BANK2 unchanged.

Tried on a SWTGW218AS: `regget` comes back with the value, `version` with its three lines, a misspelt command with `Unknown command`, and `vlan show` with the whole table.

### DIFF
--- a/httpd/httpd.c
+++ b/httpd/httpd.c
@@ -16,11 +16,6 @@
 
 #define CMARK_S 6
 
-/* Answer to POST /cmd. Kept as a macro so that its length is known at compile
- * time: an unchanged slen afterwards means the command printed nothing. */
-#define CMD_RESPONSE_HEADER "HTTP/1.1 200 OK\r\nConnection: close\r\n" \
-			    "Content-Type: text/plain\r\n\r\n"
-
 #pragma codeseg BANK1
 #pragma constseg BANK1
 
@@ -631,7 +626,9 @@ static void handle_firmware_fragment(__xdata uint8_t *p)
 
 static void run_cmd_body(__xdata uint8_t *body)
 {
-	slen = strtox(outbuf, CMD_RESPONSE_HEADER);
+	uint16_t hdr_len = strtox(outbuf, HTTP_RESPONCE_TXT);
+
+	slen = hdr_len;
 	cmd_capture = 1;
 	execute_commands(body);
 	if (cmd_capture == 2)
@@ -640,7 +637,7 @@ static void run_cmd_body(__xdata uint8_t *body)
 	/* Commands that configure something print nothing at all. Saying
 	 * so beats an empty body, which reads the same as "nothing ran".
 	 * Only on success: a silent failure must not answer with "OK". */
-	if (err_status == ERR_OK && slen == sizeof(CMD_RESPONSE_HEADER) - 1)
+	if (err_status == ERR_OK && slen == hdr_len)
 		slen += strtox(outbuf + slen, "OK\n");
 	/* On a parse error keep what the parser printed, because that text
 	 * is the explanation, and only restate the status. "400 NO" is as
--- a/httpd/httpd.c
+++ b/httpd/httpd.c
@@ -16,12 +16,17 @@
 
 #define CMARK_S 6
 
+/* Answer to POST /cmd. Kept as a macro so that its length is known at compile
+ * time: an unchanged slen afterwards means the command printed nothing. */
+#define CMD_RESPONSE_HEADER "HTTP/1.1 200 OK\r\nConnection: close\r\n" \
+			    "Content-Type: text/plain\r\n\r\n"
+
 #pragma codeseg BANK1
 #pragma constseg BANK1
 
 extern volatile __xdata uint8_t sfr_data[4];
 extern volatile __xdata uint32_t ticks;
-extern __xdata uint8_t cmd_capture;	/* owned by rtlplayground.c, see write_char() */
+extern __xdata uint8_t cmd_capture;	/* owned by rtlplayground.c, see write_char_no_syslog() */
 extern __code uint8_t * __code hex;
 extern __code struct f_data f_data[];
 extern __code char * __code mime_strings[];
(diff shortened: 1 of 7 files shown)


bloqaudio:
Nice, I built almost exactly this for an internal UI rework a while back, so two field notes and a suggestion.

`cmd_capture` needs an explicit init. The startup code on these devices does not zero xdata; I lost an evening to a keepalive flag that came up as garbage and hung boot. As written, `write_char()` reads `cmd_capture` from the first boot message onward, and it keeps its power-on garbage until the first `POST /cmd` assigns it. If that garbage is nonzero, every console print does `outbuf[slen++]`. Single-segment responses survive because handlers reassign `slen`, but responses larger than one MSS carry their remaining count in `slen` across appcalls, so a link-change print in the middle of a big transfer shifts it and corrupts the download. One line in `httpd_init()` (`cmd_capture = 0;`) closes it.

Two things from my version that might be worth taking:
- Reserve a few bytes at the cap and append `[output truncated]` so a clipped listing is visibly clipped rather than silently short.
- On `err_status != ERR_OK`, keep the captured text and patch the status line in place instead of `send_bad_request()`. `200 OK` and `400 NO` are the same length, so it is a 6-byte copy at offset 9, and the caller sees the parser's own `ERROR: ...` line instead of the 400 HTML page, which the new `cmdSub()` would otherwise dump into the transcript.

DrDoof:
Thank you, that was worth more than the review I would have given it myself. Two of your three landed, and the third sent me looking at something I had assumed rather than checked.

The truncation marker and the status line are both in, along with a third one I found while acting on yours.

`syslog off` used to answer with nothing at all and `syslog on` with the bare address, because `print_string_no_syslog()` goes straight to `write_char_no_syslog()` and so walked past the capture. Those are exactly the replies of the commands that must not generate a syslog packet. Moving the capture one level down catches both paths and costs nothing, since `write_char()` reaches it anyway. On hardware now: `Stopped syslog`, and the start message comes back as a whole sentence rather than the address on its own.

On the error path you were right about the shape of the fix and I had underrated how bad the current behaviour was. `send_bad_request()` rewrites the buffer from index 0, so the parser's own explanation is thrown away and the 400 page takes its place, and the console window in the new page appends whatever comes back without looking at the status. Typing a bad command put `<!DOCTYPE HTML PUBLIC>` in the transcript. It is patched in place now, as you suggested. A command longer than the buffer answers `400 NO` with `ERROR: Command too long: ...` in the body.

The truncation cap is reserved for the marker rather than added on top, so the worst case is one byte inside the buffer. Verified by running the arithmetic under a sanitizer against every output length from 2400 to 100000, and then on the device: three table dumps in one request come back as exactly 2435 bytes ending in `[output truncated]`, which is the header subtracted from the buffer.

The uninitialised flag I could not reproduce, and I think the premise does not hold for this tree. sdcc's own startup is linked: `___sdcc_external_startup` compiles to `MOV DPL,#0` and `RET`, so the branch that would skip the clear is not taken, and the clear loop counts out `l_XSEG`, which the flag sits inside. There is a simpler argument too: not one xdata global in `rtlplayground.c` carries an initialiser, and `authenticated` in `httpd.c` is among them, so a tree that came up with dirty xdata would serve pages without a login after every cold boot. I built and read this on sdcc 4.5.0 for SWTGW218AS, and I would rather hear which version you have than assume the difference is elsewhere, because the failure you describe is a real one and I have seen its shape before.

Two more went in from the same audit, both on the page rather than the firmware. The transcript caps itself rather than growing without bound, and a 401 sends the browser to the login form instead of appending nothing and clearing the input, which used to make an expired session look like a command that ran and printed nothing.

DrDoof:
A nudge on this one as well. The notes from bloqaudio landed a while back and it has been quiet since.

Rechecked against current main today: it merges cleanly and costs 79 bytes of the common segment, which is comfortable now that the per machine init hooks moved to BANK2. I measured 1111 bytes free before and 1032 after, identical on every board I built.

Happy to rebase it, split it, or drop the console tab part if only the /cmd change is wanted.

Thanks.

bloqaudio (approved):
Reviewed the series and ran an independent verification against current main (fc3c686).

Build, sdcc 4.5.0: merges clean, machine_check compiles every machine, and both boards I link full images for came out healthy, including PCB_SWTG018AS_V2_1_0. Internal RAM is untouched on both (stack byte for byte identical to main), xdata grows by the one flag byte, and the common segment cost is the 79 bytes you measured, leaving 1012 bytes of bank 0 headroom after the #359 merge.

On a SWTG018AS V2.1.0 board running main plus this series: version and vlan show come back in the body, a config command that prints nothing answers OK, a misspelt command answers 200 with Unknown command, an overlong command answers 400 NO with the parser's ERROR line in the body, syslog off and on answer with their console replies (the write_char_no_syslog() placement doing its job), no cookie answers 401, and six table dumps in one request come back as exactly 2435 bytes ending in [output truncated].

One observation, pre existing and not this PR's problem: a POST /cmd whose body arrives in a separate TCP segment from the headers executes nothing, and with this series it answers OK rather than an empty 200. Browsers send these small POSTs in one segment, so the console tab is unaffected, and the multipart paths got their equivalent fix in #356 and #359. Worth an issue at some point; it should not hold this up.

DrDoof:
Thank you, that is far more verification than I would have asked for, and the internal RAM comparison is something I had not thought to check myself.

You are right about the split POST, and I went and read it again after your note. `execute_commands()` runs on a pointer into the segment that carried the headers, and nothing accumulates a body arriving later, so there is simply nothing there to parse. The OK makes that silence visible rather than causing it.

I would rather fix it than file it. The config and upload paths already buffer, so `/cmd` can follow the same shape, and I will put that up as its own PR once this one is in so it stays out of the way here.

Thanks for the board time.

bloqaudio:
> Thank you, that is far more verification than I would have asked for, and the internal RAM comparison is something I had not thought to check myself.
> 
> You are right about the split POST, and I went and read it again after your note. `execute_commands()` runs on a pointer into the segment that carried the headers, and nothing accumulates a body arriving later, so there is simply nothing there to parse. The OK makes that silence visible rather than causing it.
> 
> I would rather fix it than file it. The config and upload paths already buffer, so `/cmd` can follow the same shape, and I will put that up as its own PR once this one is in so it stays out of the way here.
> 
> Thanks for the board time.

I am working on a fix right now.

httpd/httpd.c
```diff
@@ -16,10 +16,16 @@
 
 #define CMARK_S 6
 
+/* Answer to POST /cmd. Kept as a macro so that its length is known at compile
+ * time: an unchanged slen afterwards means the command printed nothing. */
+#define CMD_RESPONSE_HEADER "HTTP/1.1 200 OK\r\nConnection: close\r\n" \
+			    "Content-Type: text/plain\r\n\r\n"
```
vDorst: This is also defined in page_impl.c as `HTTP_RESPONCE_TXT`, but missing "Connection: close" part. Update and use one or the other.

DrDoof: Good catch, and it turned out to be more than a duplicate.

The two strings weren't quite the same: the shared `HTTP_RESPONCE_TXT` is the one missing `Connection: close`. `HTTP_RESPONCE_JSON` right above it has always had it, so the text one was the odd man out, and `send_config()` and `send_cmd_log()` have been answering without it. On a server that closes after every response that's the case the comment further down in httpd.c warns about, the browser keeps the socket in its pool and the next request lands on a connection that's already gone.

So I did it the way round you suggested: the header went into the shared string and my copy is gone. The "command printed nothing" check compares against the length `strtox()` returned now, rather than `sizeof()` on the macro.

Neither of the other two users can overrun the buffer with the extra 19 bytes. `send_config()` bounds against `TCP_OUTBUF_SIZE - slen` and pushes the rest into the continuation, and `send_cmd_log()` tops out at CMD_HISTORY_SIZE, so 1024 plus the header against a 2500 byte buffer.

BANK1 comes out 23 bytes smaller, BANK2, xdata and the stack are unchanged, and machine_check is clean.

vDorst:
thanks

PR #330, opened 2026-08-13
Sram easy changes

vDorst:
These are the easy changes to free-up SRAM usage by the code/compiler.

Looking in the build directory, generated asm-files.
See which functions passed arguments via SRAM or put local variable also SRAM because compiler has no space.

Example from `httpd.asm`.
```
; internal ram data
	.area DSEG    (DATA)
_content_type::
	.ds 2
_session::
	.ds 2
_timeptr::
	.ds 2
_scan_header_p_10000_152:
	.ds 2
_gen_random_bytes_PARM_2:
	.ds 1
```

You can see that `content_type`, `session` and `timeptr` are placed in SRAM.
Fix was simple, add `__xdata` to the declaration so the compiler put it in XDATA.

Sometimes you have to look into the code, to see that compiler stores the passed argument direcly in SRAM at the begin of the function to make room for other things.
Like `scan_header_p_10000_152`, many thing this can be fixed by passing that argument via xdata.

```patch
4,10c4,10
< 0x20:|B|B|T|d|d|e|e|e|e|e|e|e|e|e|e|e|
< 0x40:|e|e|f|f|f|f|f|g|g|g|g|g|g|g|g|g|
< 0x50:|g|g|g|g|g|g|h|h|h|h|h|h|h|h|h|h|
< 0x60:|i|i|i|i|i|i|i|i|i|j|j|j|j|j|j|j|
< 0x70:|j|j|j|j|j|Q|Q|Q|Q|Q|Q|Q|Q|S|S|S|
---
> 0x10:|b|b|b|b|b|b|b|b|b|b|b|c|c|g|g|g|
> 0x30:|d|d|d|d|d|d|d|d|d|d|d|e|e|e|e|e|
> 0x40:|f|f|f|f|f|f|f|f|f|h|h|h|h|h|h|h|
> 0x50:|h|h|i|i|i|i|i|i|i|i|i|i|i|i|j|Q|
> 0x60:|Q|Q|Q|Q|Q|Q|Q|S|S|S|S|S|S|S|S|S|
> 0x70:|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|
21c21
< 16 bit mode initial stack starts at: 0x7d (sp set to 0x7c) with 131 bytes available.
---
> 16 bit mode initial stack starts at: 0x67 (sp set to 0x66) with 153 bytes available.
28,29c28,29
<    EXTERNAL RAM     0x0001   0x27df   10207    16777216
<    ROM/EPROM/FLASH  0x0000   0x2d59a  96027    16777216
---
>    EXTERNAL RAM     0x0001   0x27f7   10231    16777216
>    ROM/EPROM/FLASH  0x0000   0x2d5c2  96330    16777216
```

Free up 22 bytes of SRAM-bytes

### DIFF
--- a/httpd/httpd.c
+++ b/httpd/httpd.c
@@ -50,8 +50,8 @@ __xdata uint8_t config_upload;
 __xdata uint8_t config_buf[CONFIG_UPLOAD_BUF];
 __xdata uint16_t cfg_pos, cfg_hdr, cfg_body, cfg_end, cfg_last;
 __xdata uint8_t cfg_bl;
-__xdata uint8_t *content_type = 0;
-__xdata uint8_t *session = 0;
+__xdata uint8_t * __xdata content_type = 0;
+__xdata uint8_t * __xdata session = 0;
 
 // Global variables holding POST state
 __xdata uint16_t bindex; // Current index into the boundary
@@ -63,7 +63,7 @@ __xdata char passwd[21];
 __xdata char session_id[SESSION_ID_LENGTH + 1];
 __xdata uint8_t authenticated;
 __xdata uint32_t now;
-__xdata uint8_t *timeptr;
+__xdata uint8_t * __xdata timeptr;
 __xdata uint32_t last_session_use;
 
 #define TSTATE_NONE		0
@@ -311,7 +311,7 @@ __xdata uint8_t *scan_header(__xdata uint8_t *p)
 }
 
 
-void gen_random_bytes(__xdata uint8_t *b, uint8_t bytes)
+void gen_random_bytes(__xdata uint8_t *b, register uint8_t bytes)
 {
 	__xdata uint8_t i = 0;
 	while (bytes) {
--- a/httpd/httpd.c
+++ b/httpd/httpd.c
@@ -310,10 +310,12 @@ __xdata uint8_t *scan_header(__xdata uint8_t *p)
 	return p;
 }
 
-
-void gen_random_bytes(__xdata uint8_t *b, register uint8_t bytes)
+/*
+ * Generate random HEX-chars at the buffer location.
+ */
+void gen_random_hex_chars(__xdata uint8_t * b, __xdata uint8_t bytes)
 {
-	__xdata uint8_t i = 0;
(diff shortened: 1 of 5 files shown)


DrDoof:
Some numbers from pushing this a bit further on `httpd/page_impl.c`, in case they save you a build.

I took the twelve scalar locals in that file that carry `__xdata`, the ones in `send_vlan`, `send_l2`, `l2_delete`, `send_config`, `send_cmd_log` and `send_vlanlist`, and dropped the qualifier from all of them at once. The 256 byte `flash_buf` in `send_config` I left where it is. It does not link:

```
?ASlink-Error-Could not get 7 consecutive bytes in internal RAM for area OSEG.
?ASlink-Error-Could not get 8 consecutive bytes in internal RAM for area OSEG.
```

Which is what the map says should happen. On main today the modules hold 55 of the 128 bytes of DSEG, the overlay sits at 0x71 and runs eight bytes to 0x79, and the stack starts at 0x79. There is nothing in between, so the overlay cannot grow without taking stack away.

So in that file the qualifier is already doing the job you describe rather than being habit: the compiler has run out and moving those back would make them permanent. `page_impl`'s own DSEG is five bytes as it stands.

The same experiment on an image where `page_impl` is close to the only heavy user does link, and comes out the way you would expect when there is slack: less code and less xdata for a few more bytes of DSEG. The difference between the two builds is only how much room was left, which suggests the useful question here is which functions can give the overlay room back, rather than which locals can come out of xdata. The two ends of that meet quite quickly.

vDorst:
Also rebase this PR. Please review.

vDorst:
Rebase again. Please review.

vDorst:
@DrDoof  @pleas @feelfree69 @bloqaudio I would like to ask you all to helping us with reviewing patches not only this one but others too. I feel the project is stalling a bit on important patches. So if you could review some patches a just give a "Files changes -> submit review -> Approve" otherwise a comment that it is fine, that would be helpful.

feelfree69 (approved):

PR #331, opened 2026-08-13
regs: wrap REG_SET and REG_WRITE, and fix what that uncovers

DrDoof:
`REG_SET` and `REG_WRITE` expand to a run of statements with no do/while around them, so as the unbraced body of an `if` or a `for` only the first assignment belongs to that body. The other three and the `reg_write()` call sit after it and run once, unconditionally, with whatever the loop counter ended on.

Five call sites are written that way. Wrapping the four definitions fixes them without touching a single one, and the two follow-up commits deal with what the restored guards then expose.

What it looks like on an SWTGW218AS, all read back with `regget` after a cold boot.

The IGMP one is the clearest. `igmp_setup()` loops over the ports to install the per-port configuration, and the write landed once, at `machine.max_port + 1`, so the ports themselves never got it. Before: every port read `0x00ff7c00`, the reset state, while index 9 held the `0x00ff7c15` the loop meant for them, and `igmp off` left a preceding `igmp on` in place until the next reboot. After: ports read `0x00ff7c15` and `igmp off` restores it. `igmp_enable()` has the same loop with braces around it and was always correct, which is why turning snooping on worked and turning it off did not.

I marked three unused ports with `0x00ff7c3f` and called `igmp off` to find where the write was going: the marks survived and `0x52c4` picked up the value, which is index 9 rather than any port.

Two sites change behaviour once the guard works again, and both wanted fixing rather than leaving.

`port_lag_members_set()` installs `LAG_HASH_DEFAULT` when the hash register reads all zeros. That test never passes, because the register comes out of reset holding `0x3f`, so with the guard restored the default would never be installed where before it was installed on every call. The same line also wrote to the base address while the read that decides it used the group offset, so a group other than zero was tested and group zero was written. Both ends use the offset now and the test is against the reset value, which the header names. Measured across three images on the same switch: the hash for the configured group went `0x7e`, then `0x3f` when the guard started working, then `0x7e` again with the test corrected, and the unconfigured groups stayed at `0x3f` throughout, which is what says the offset now reaches the right register.

`port_isolate()` guards on `machine.max_port`, so it now declines the CPU port, which is one past that. `parse_isolate()` does accept the CPU port and `port_isolation_get()` has always returned 0 for it, so reading and writing disagreed and the write only landed by accident, with the top byte of the mask left over from an earlier register operation. This makes the two agree on refusing. Whether the CPU port ought to be isolatable is a separate question and I have not answered it here.

The remaining two are small. `rtl837x_leds.c:187` wrote an eight-port LED configuration on every machine and now writes it where the condition intended. The duplicate loop in `igmp_setup()` wrote the same value twice, which cost one redundant register write while the macro swallowed the body and would cost one per port now, so it goes.

Twenty nine bytes smaller than before on SWTGW218AS. No change to internal RAM or xdata. Built for KP_9000_6XHML_X2 as well.

### DIFF
--- a/rtl837x_regs.h
+++ b/rtl837x_regs.h
@@ -305,32 +305,40 @@
 
 #ifdef REGDBG
 
-#define REG_SET(r, v) SFR_DATA_24 = (((uint32_t)v) >> 24) & 0xff; \
+#define REG_SET(r, v) do { \
+	SFR_DATA_24 = (((uint32_t)v) >> 24) & 0xff; \
 	SFR_DATA_16 = (((uint32_t)v) >> 16) & 0xff; \
 	SFR_DATA_8 = (((uint16_t)v) >> 8 & 0xff); \
 	SFR_DATA_0 = (v) & 0xff; \
 	reg_write(r); \
 	write_char('R'); print_byte(r >> 8); print_byte(r); write_char('-'); \
-	print_byte(((v) >> 24) & 0xff); print_byte((v) >> 16 & 0xff); print_byte((v) >> 8 & 0xff); print_byte( (v) & 0xff); write_char(' ');
+	print_byte(((v) >> 24) & 0xff); print_byte((v) >> 16 & 0xff); print_byte((v) >> 8 & 0xff); print_byte( (v) & 0xff); write_char(' '); \
+} while (0)
 
-#define	REG_WRITE(r, v24, v16, v8, v0) SFR_DATA_24 = (v24); \
+#define	REG_WRITE(r, v24, v16, v8, v0) do { \
+	SFR_DATA_24 = (v24); \
 	SFR_DATA_16 = (v16); \
 	SFR_DATA_8 = (v8); \
 	SFR_DATA_0 = (v0); \
 	reg_write(r); \
-	write_char('R'); print_byte(r>>8); print_byte(r); write_char('-'); print_byte(v24); print_byte(v16); print_byte(v8); print_byte(v0); write_char(' ');
+	write_char('R'); print_byte(r>>8); print_byte(r); write_char('-'); print_byte(v24); print_byte(v16); print_byte(v8); print_byte(v0); write_char(' '); \
+} while (0)
 #else
-#define REG_SET(r, v) SFR_DATA_24 = (((uint32_t)v) >> 24) & 0xff; \
+#define REG_SET(r, v) do { \
+	SFR_DATA_24 = (((uint32_t)v) >> 24) & 0xff; \
 	SFR_DATA_16 = (((uint32_t)v) >> 16) & 0xff; \
 	SFR_DATA_8 = (((uint16_t)v) >> 8 & 0xff); \
 	SFR_DATA_0 = (v) & 0xff; \
-	reg_write(r);
+	reg_write(r); \
+} while (0)
 
-#define	REG_WRITE(r, v24, v16, v8, v0) SFR_DATA_24 = (v24); \
+#define	REG_WRITE(r, v24, v16, v8, v0) do { \
+	SFR_DATA_24 = (v24); \
 	SFR_DATA_16 = (v16); \
 	SFR_DATA_8 = (v8); \
 	SFR_DATA_0 = (v0); \
-	reg_write(r);
+	reg_write(r); \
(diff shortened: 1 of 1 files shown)


DrDoof:
One overlap worth flagging before you look at these in sequence: #299 touches the same line in `port_lag_members_set()`. It adds the group offset to the hash write, which this branch also does, so whichever lands second will want a small rebase there.

The part that does not overlap is the test in front of it. #299 keeps `if (!(sfr_data[0] | sfr_data[1] | sfr_data[2] | sfr_data[3]))`, which passes only when the register reads zero, and it never does. That did not matter while the macro swallowed the guard, because the write happened regardless. It starts mattering the moment the macro is wrapped, so if you take #299 without this branch's third commit the default hash would stop being installed rather than start being installed in the right place.

Same for the missing `return` after the range check, which is here and not there.

rtl837x_regs.h
```diff
@@ -232,6 +232,11 @@
 #define LAG_HASH_L4_SPORT		0x20
 #define LAG_HASH_L4_DPORT		0x40
 #define LAG_HASH_DEFAULT (LAG_HASH_L2_SMAC | LAG_HASH_L2_DMAC | LAG_HASH_L3_SIP | LAG_HASH_L3_DIP | LAG_HASH_L4_SPORT | LAG_HASH_L4_DPORT)
+/* What the hash registers hold after a reset, measured on an SWTGW218AS:
```
logicog: The comment is not necessary. The definition speaks for itself and the meaning is given by the name of the variable. If you put name and comment, readers are confused.

rtl837x_port.c
```diff
@@ -744,11 +744,18 @@ void port_lag_members_set(__xdata uint8_t lag, __xdata uint16_t members) __banke
 {
 	print_string("port_lag_members_set, lag: "); print_byte(lag); print_string(", members: "); print_short(members);
 	write_char('\n');
-	if (lag > 3)
+	if (lag > 3) {
 		print_string("Link aggregation group must be 0-3!\n");
+		return;
+	}
+	/* Install the default hash only where nobody has chosen one. The test
```
logicog: Please remove the comment. A comment is for a reader of the code, but this is a comment for the user of the AI writing the code. "used to be against" will utterly confuse a subsequent reader, because he/she will only be able to see the result, not the process. Also, the code speaks for itself.

logicog:
Can this be split into a PR for the guards around the macros and then 1-2 PRs for the different fixes. The commit message for the guards does not make sense as it talks about the other changes. The commit message for the guards should just be something like "Fix macro definitions for register access by wrapping them in do {} while(0)" or something along the lines.

The original definitions are historic, btw, as it was not clear what the code would look like that the compiler generated. Too much optimization of the SFR access leads to very strange results (opcodes directly manipulating a register do not work, a value has to be read by mov, manipulated, and re-written with mov), so the original macros were written to break easily when there was the danger for the compiler to do something clever.

DrDoof:
Both comments are in, and the branch is now the macro change on its own with a message that only talks about the macro.

You were right about the second one in a way I hadn't seen. The comment described the change rather than the state, and a reader a year from now has only the result in front of them, so a sentence about how it used to be points at nothing they can see. That belongs in the commit message, and that's where the measurement lives now.

The two fixes are out separately as #337 for the duplicate igmp loop and #338 for the trunk hash. #338 sits on top of this one rather than on main, because the test it repairs only means anything once the guard works. That does create an ordering gap worth knowing about: this PR alone restores the guard in front of the hash default, and the test it guards is written against zero, which the register never reads, so between this landing and #338 landing the default isn't installed at all where it used to be installed on every call.

Thank you for the history, it explains a shape I'd read as an oversight. It also sent me to check whether wrapping hands the compiler any new room around the SFR writes, since that's what the loose form was defending. It doesn't. Building the image before and after and comparing the generated assembly module by module, with label numbering, block scope suffixes and the version string normalised away, three modules differ, and they're exactly the three holding an unbraced use. Ten modules call these macros, so the other seven come out identical, and so does the rest of the image.

vDorst (approved):
Great that you spotted this error.
Thanks

vDorst (approved):

PR #332, opened 2026-08-14
igmp: only hand reports to the CPU while snooping is on

DrDoof:
Following up on the offer in #327.

`handle_rx()` picked the IGMP branch on the destination address alone, so a report off the wire reached `igmp_packet_handler()` whether or not anyone had switched snooping on, and the handler goes straight to writing a table entry. The STP branch immediately above it is gated on `stpEnabled`, so this only brings IGMP into line with what is already there.

One thing I should have placed properly in #327: the `lacpEnabled && uip_buf[5] == 0x02` line I compared against lives in #299, not in main. On main the neighbouring gate is the STP one. Either way the point is the same, the branch next to it asks whether the feature is switched on and the IGMP branch does not.

Snooping state lives only in the per-port registers, and the receive path cannot afford a register read per packet, so the flag shadows it. `igmp_enable()` sets it and `igmp_setup()` clears it. Since `igmp_setup()` is both the "igmp off" path and the boot path, the flag starts out clear and follows the registers wherever they get set from.

Six bytes of BANK1 and one of xdata, no internal RAM. Built for SWTGW218AS on sdcc 4.5.0 and measured on one: `igmp on` sets the trap action on every port and the flag with it, `igmp off` clears both.

That last part only became true with #331, which is worth knowing if you take these in either order. Until then `igmp off` reached the per-port registers of nothing except one slot past the last port, so the trap action survived it and the flag was the only thing that actually stopped. The two changes are independent and do not touch the same lines, but they answer halves of the same question.

There is one thing I ran into while checking this, and you may want it before deciding how much the gate is worth: I do not think the handler can parse a report at all as it stands. `struct igmp_pkt` has no `struct vlan_tag`, while the other three receive structs in the tree do have one (`eth_in`, `arp_hdr_i`, `stp_pkt_in`). Everything past the RTL tag therefore sits four bytes early: `protocol` reads the low byte of the IP identification field, and `igmp_type` reads the first byte of the Router Alert option, which is never 0x12, 0x16 or 0x22. I took the offsets out of the generated assembly rather than working them out on paper.

That would also explain why the missing gate never turned into a symptom. I have not sent a patch for it, because so far I can only show that the parse is off, not that correcting the struct makes snooping work end to end.

### DIFF
--- a/rtl837x_igmp.c
+++ b/rtl837x_igmp.c
@@ -16,6 +16,7 @@
 #include "machine.h"
 
 extern __code struct machine machine;
+extern __xdata uint8_t igmpEnabled;
 
 #include "uip.h"
 
@@ -85,6 +86,7 @@ void igmp_setup(void) __banked
 {
 	uint8_t i;
 	print_string("igmp_setup called\n");
+	igmpEnabled = 0;
 	// For now, forward all unkown IP-MC pkts (2 bits per port. 00: flood via floodmask, 01: drop, 10: trap, 11: to rport)
 	REG_SET(RTL837X_IPV4_PORT_MC_LM_ACT, LOOKUP_MISS_FLOOD);
 	REG_SET(RTL837X_IPV6_PORT_MC_LM_ACT, LOOKUP_MISS_FLOOD);
@@ -130,6 +132,7 @@ void igmp_setup(void) __banked
 void igmp_enable(void) __banked
 {
 	print_string("igmp_enable called\n");
+	igmpEnabled = 1;
 	// Configure trapping of unhandled IGMP protocol packets to CPU
 	REG_SET(RTL837X_IGMP_TRAP_CFG, IGMP_CPU_PORT | IGMP_TRAP_PRIORITY);
 
--- a/rtlplayground.c
+++ b/rtlplayground.c
@@ -121,6 +121,7 @@ __xdata uint16_t management_vlan;
 __xdata uint8_t tx_seq;
 
 __xdata uint8_t stpEnabled;
+__xdata uint8_t igmpEnabled;
 __xdata char hostname[24];	/* device hostname, default set at boot, see rtl837x_common.h */
 
 __code uint16_t bit_mask[16] = {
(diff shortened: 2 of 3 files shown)


rtl837x_igmp.c
```diff
@@ -85,6 +86,8 @@ void igmp_setup(void) __banked
 {
 	uint8_t i;
 	print_string("igmp_setup called\n");
+	/* Also the "igmp off" path: stop handing reports to the CPU. */
```
logicog: Don't comment a variable that is used for logic flows, as the variable may be used differently when the code elsewhere is changed later. Then the comment becomes contradictory. The name anyway speaks for itself.
It may be better to add an "igmp off" command, so that the logic becomes clear. That this does not exist appears to be an oversight.

rtlplayground.c
```diff
@@ -121,6 +121,10 @@ __xdata uint16_t management_vlan;
 __xdata uint8_t tx_seq;
 
 __xdata uint8_t stpEnabled;
+/* Set by igmp_enable(), cleared by igmp_setup(), which is what "igmp off"
```
logicog: Again, don't comment state or logic variables, as the action happens elsewhere and may subsequently change while forgetting to update the comment. It is very easy to figure out where a variable is used in code via either the IDE or just "rg" on the command line, so one does not need the comment for understanding. I just looked at a couple of Linux network drivers and could not find a single example of a variable being commented. What is commonly commented are data structures and the meaning of the fields.

logicog (commented):
Great catch! If you could remove/shorten the comments and maybe add the "igmp off" command, then it is perfect.

rtl837x_igmp.c
DrDoof: @logicog Both done. igmp off is now an explicit subcommand that calls igmp_setup(), and anything the parser does not recognise prints the usage line instead of quietly turning snooping off. The comment here is gone. Folded into the one commit (56fa96c49420e832afc542e7c16a3ef0c647770f) rather than added on top.

rtlplayground.c
DrDoof: @logicog Comment removed, the name stands on its own. Point taken on state variables in general, I will keep comments to structures and their fields from here on.

logicog:
Thanks for taking this on!

PR #333, opened 2026-08-14
syslog: send a hostname, allow a port other than 514, and stop the save from dropping the address

DrDoof:
Three things about the syslog client, found in the order they stopped me from using it.

Lines arrive without a hostname. RFC 3164 puts one between the priority and the text, and we send nothing there, so a receiver fills the field from the first word of the message instead. Every line ends up attributed to `STP:`, or `REGGET:`, or whatever the log happens to begin with, and the sender cannot be selected on at all because the name changes with the content. The first commit puts our hostname where the field belongs.

The destination port was frozen at 514 in the call that opens the connection. Anything listening elsewhere was out of reach, and plenty do listen elsewhere, because 514 is privileged and a collector in a container usually gets a high port instead. The way around it is a proxy in front, which then rewrites the source address and hides which device sent the line, which is the same problem as the missing hostname wearing a different hat. `syslog port <n>` is now a command, defaulting to 514 so an existing configuration behaves as before.

The port is parsed in place rather than through `atoi_short()`, which reports "no digits" but not overflow, so 65540 would wrap to 4 and open a connection nobody asked for. Checking before each multiply keeps it inside 16 bits, and zero then covers all three ways to be wrong: no digits, past 65535, or an explicit zero, which is not a port either.

The third one is why the other two were hard to test. Saving the configuration drops the syslog server address whenever the command log also contains `syslog on`, which it nearly always does, since you enable the thing after pointing it somewhere. `parseConf()` deduplicates by taking the matched text of a `conf_overwrite` entry and removing every stored line starting with it; for `syslog on` the entry that matches is the catch-all `^syslog\b`, so the matched text is just `syslog` and the filter eats the `syslog ip ...` line saved a moment earlier. The address falls back to 0.0.0.0 on the next boot, quietly: nothing is reported, the running switch keeps logging until it restarts, and the loss reads like the firmware forgetting a setting rather than the page dropping it. Narrowing the entry to the commands it means to collapse leaves the address and the port with entries of their own.

`itoa_short()` is new, because `itoa()` handles a byte and `print_short()` prints hex, so there was no way to show a port as a decimal number. It is separate from `itoa()` rather than a widening of it, since every existing caller passes a byte and would start paying for 16 bit divisions it does not need.

The parser scratch variables are in xdata on purpose. Locals would take internal RAM and this image has none to spare, with data ending at 0x77, the overlay at 0x78 to 0x7f and the stack starting at 0x80. Measured: no module's DSEG grew. The three commits together cost 656 bytes of BANK2 and nine of xdata, nothing in BANK1 or internal RAM. Built for SWTGW218AS and KP_9000_6XHML_X2 on sdcc 4.5.0.


DrDoof:
Updated the third commit after testing it rather than reasoning about it.

The narrowed entry keeps the address and the port safe, which was the point, but it also stopped `syslog on` and `syslog off` from replacing each other, because the list drops lines starting with the text it matched and `syslog off` does not start with `syslog on`. The old catch-all collapsed the pair by accident, so narrowing it lost that for free. I had claimed the opposite in the commit message.

Running the real `parseConf()` over a saved configuration showed both lines surviving side by side, so a separate list now names the stem of such a pair and collapses the two. Same configuration through it: stable across repeated saves, and two lines that the current code drops are now kept.

cmd_parser.c
```diff
@@ -1312,11 +1320,66 @@ void parse_syslog(void)
...
+			}
+			if (syslog_port_scratch == 6553) {
+				if (syslog_port_digit > 5) {
+					syslog_port_scratch = 0;
+					break;
+				}
+			}
+			syslog_port_scratch = syslog_port_scratch * 10 + syslog_port_digit;
+			syslog_port_idx++;
+		}
```
vDorst: Parsing u8/u16 number and overflow detection is fixed in #326, maybe you can help review that PR.
Because is makes handing of parsing cmd_buffer more cleaner.
Also made a helper `print_ip()`

DrDoof:
Thanks for the pointer, that was the right one. I went through #326 and left the details there rather than here: the short of it is that `atoi_short` is doing exactly what this hand rolled loop does, correctly, and I will move this PR onto it once #326 lands. I would rather not build on an unmerged PR before then, so the loop stays for now.

Your comment did make me look at it properly, though, and the loop was carrying a justification that does not hold. It tested the digit range and the overflow bound with separate ifs because of a comment claiming the joined form makes sdcc park a boolean in the bit area, and that an image with LACP and STP in it is one bit short of needing a third byte. Measured on this branch that is simply not the case: the joined form links, the bit segment stays at fourteen bits overall and nine in cmd_parser, and BANK2 comes out ten bytes smaller. The claim was about an image this branch does not build, and no branch currently carries LACP, STP and this parser at the same time, so there is nothing left to check it against either.

So the loop is now the joined form and the comment is gone. The two versions agree on every input from 0 to 99999 and on the edges, so nothing about what the command accepts has changed. What stays is the note about `atoi_short()` reporting "no digits" without reporting overflow, which is the thing #326 fixes.

vDorst:
Please rebase

DrDoof:
Rebased, and sorry you had to ask.

Your pointer to #326 was the right one. It has landed, and the new `atoi_short()` carries the same overflow bound the hand rolled loop was written for, so the loop is gone and three xdata scratch variables with it. Deleting code because someone else did it better is a good way to spend an evening.

Two things needed resolving against main: in `syslog.c` the hostname header met the old payload copy, and the status line now uses your `print_ip()` with the port appended. Builds for SWTGW218AS and KP_9000_6XHML_X2.

I saw your note about the project stalling on reviews as well. That is a fair thing to ask for, and I will take a proper look at the open ones rather than only my own.

feelfree69:
Gave it a quick try; works as expected. 
We should consider to rename the `syslog_server_ip` in the overview of the WebUI to `syslog_server`and add the portnumber.

syslog.c
```diff
@@ -12,6 +12,11 @@ __xdata char logbuf[LOGBUF_SIZE];
 __xdata struct syslog_state syslog_state;
 __xdata uip_ipaddr_t server_ip;
 
+/* Header length, in xdata rather than on the stack: syslog_callback() is
+ * __banked, so a local here takes a byte of internal RAM nobody can overlay,
+ * and the two it costs are enough to stop the image from linking. */
+__xdata uint16_t syslog_hdr;
+
```
vDorst: This can be just a local variable.
After moving I think the comment is not useful anymore.

syslog.h
```diff
@@ -5,12 +5,15 @@
 
 #define LOGBUF_SIZE 512
 
+#define SYSLOG_PORT_DEFAULT 514
+
 struct syslog_state {
     uint8_t enabled;
```
vDorst: make is a bool

cmd_parser.c
```diff
@@ -65,6 +65,10 @@ __xdata	char save_cmd;
 
 __xdata uint8_t ip[4];
 
+/* Scratch for parse_syslog(), in xdata: a local here would take internal RAM
+ * the linker has none of. */
+__xdata uint8_t  syslog_port_was_on;
+
```
vDorst, line 71: This variable can be removed?
Just use `syslog_state.enabled` directly.

syslog.c
DrDoof: You're right, it fits now. That comment was written when a local wouldn't link, but that was on the branch that also carries LACP, where the overlay area is gone. Here OSEG is still 8 bytes and it links fine, so the local is back and the comment with it.

syslog.h
DrDoof: Done. Needed `stdbool.h` in the header too, since syslog.h gets included before rtl837x_common.h.

cmd_parser.c
DrDoof: This one I'd rather keep, if you don't mind. `syslog_stop()` clears `state.enabled` on its first line, so by the time the second test runs the flag is already zero and syslog would never come back up after a port change. It isn't a copy of the state, it's the state from before we cleared it.

Happy to rename it to something that says that more clearly if the name is what's bothering you.

DrDoof:
@feelfree69 thanks for testing it. Renamed: the overview row is `syslog_server` now and shows the port with the address, since main_info.js uses the JSON key as the label so the key is what people read.

feelfree69 (approved):

syslog.c
```diff
@@ -22,20 +22,22 @@ void syslog_init(void) __banked
...
-		state.syslog_conn = uip_udp_new(&server_ip, HTONS(514));
+		state.syslog_conn = uip_udp_new(&server_ip, HTONS(state.server_port));
 		if (state.syslog_conn == 0) {
 			print_string_newline_no_syslog("Failed to create a new UDP client");
 			return;
 		}
 		print_string_newline_no_syslog("Started syslog to IP ");
 		itoa(state.server_ip[0]); write_char('.'); itoa(state.server_ip[1]); write_char('.');
-		itoa(state.server_ip[2]); write_char('.'); itoa(state.server_ip[3]); write_char('\n');
+		itoa(state.server_ip[2]); write_char('.'); itoa(state.server_ip[3]);
```
vDorst: `print_ip()` can be used here.

cmd_parser.c
vDorst: A yes I see what you was trying to do in the code.
Keeping the variable is right.
Name is better now, it reflex better was it is doing.

vDorst:
Only the `print_ip()` and the rest is looking good.

syslog.c
DrDoof: Done. It came out better than a straight swap, the hand rolled quad was seven calls against one, so BANK2 ends up 48 bytes lighter than it was.

One thing I'd rather flag than bury: the prototype isn't `__banked`. cmd_parser.c and syslog.c are both BANK2, so a plain call is what fits, and I measured the other way round, banked costs 58 bytes across the call sites you already have. The catch is that it has to change if either file ever moves bank, so I put that in the commit message rather than leave it sitting there quietly. Say the word if you'd sooner pay the 58 and not have to remember.

DrDoof:
`print_ip()` is in, so that should be the last of it.

Builds for SWTGW218AS and KP_9000_6XHML_X2_V1_1, and `machine_check` is clean across the lot. BANK2 came out 48 bytes smaller than before, xdata and stack unchanged, which was a nicer result than I expected from a tidy-up.

vDorst:
Thanks

PR #335, opened 2026-08-14
ports: show the devices connected behind each port

bloqaudio:
Adds a Connected devices column to the Port Configuration table. A port with a single MAC behind it shows the MAC directly; several are shown as a count with the full list in the cell tooltip, since picking one of many to display would suggest it was "the" device. Translations for all three languages included.

The data comes from the same paged /l2.json walk with wrap detection that the L2 page already uses, so there is no firmware change. The walk restarts every 15 seconds rather than polling continuously, and it starts 3 seconds after page load to stay clear of the initial status and MTU requests, since the httpd serves one connection at a time and concurrent requests at load are the flaky window.

Tested on a SWTGW218AS on current main: single-MAC ports show the MAC, down ports stay empty, and the existing speed, disable and MTU controls are unaffected.


DrDoof:
Thanks for the PR — the writeup is careful (the 3s startup offset and the 15s walk restart show you thought about the single-connection httpd), and your copy of the walk actually fixes two bugs the original `getL2()` in l2.js still has (`l2Entries >= 4096` compares the array itself, and an empty response makes `l2CurrentEntry` NaN). Credit where due.

That said, I don't think the feature should go in as-is, for two reasons that are about direction rather than polish:

**1. This duplicates the L2 page, in a weaker form.** The L2 table already shows MAC↔port with VLAN, entry type, per-column filtering and sorting. This column reduces that to "one MAC, or a count behind a tooltip" — no VLAN, and tooltips are unreachable on touch devices. If the goal is convenience while configuring a port, a link from each row to the L2 page pre-filtered on that port would give the full data for near-zero cost, instead of adding a third poller to a page that already runs `update` every 2s against an httpd that serves one connection at a time — the same flakiness window your description mentions.

**2. It hard-codes port↔MAC attribution right as trunking lands.** With LAGs (#299, #325, #338 in flight), an L2 entry for a MAC learned on a trunk points at the trunk, not a physical port — `logToPhysPort[e.port]` will return garbage/`undefined` and those entries silently vanish from the column. Conceptually the MAC is "behind the LAG", and the hash can move flows between member ports, so a per-physical-port column will show something misleading for trunked ports. Decoding the port/trunk field of `/l2.json` is about to change, and after this PR it would live in two diverged copies.

So my suggestion: if @logicog wants this convenience on the ports page at all, it should wait until the trunk representation in `/l2.json` has settled, and be built on a shared `walkL2()` helper in main.js (used by l2.js too, which would fix the bugs above in one place) so the port/trunk decoding exists exactly once. Happy to help with the shared-walk refactor — it interacts with the LAG work I'm doing anyway.

DrDoof:
To make the shared-helper suggestion concrete, this is the shape I have in mind — a single `walkL2(onDone)` in main.js (which both pages already load), doing one paged request per second with the wrap/4096/empty termination, and handing the decoded entries to a per-page callback:

```js
// main.js
function walkL2(onDone) {
  // pages through /l2.json once; idx/vlan parsed, port -> physical or 'CPU';
  // stops on wrap, empty page or 4096 entries, then onDone(entries)
}
```

- `l2.js`: `walkL2(entries => { /* map type s/l to labels */ fillL2(entries); restart walk })` — drops `getL2()` entirely, which also fixes the `l2Entries >= 4096` comparison and the empty-response `idx = NaN` case in one place.
- `ports.js` (this PR): `walkL2(entries => devRender(entries))` with its 15s restart — the per-port grouping and cell rendering stay here, only the transport/decoding moves out.

The main payoff is that the port/trunk decoding of `/l2.json` then exists exactly once: when trunk entries start showing up (#299/#325/#338), only `walkL2()` needs to learn about them, and both pages inherit it. I'd hold off on landing the column itself until that representation has settled, but if @logicog wants the feature, rebasing this PR onto such a helper would be a small change. I can put up the main.js + l2.js part as a separate PR if that helps.

bloqaudio:
You are right about the trunk problem, and I would rather have the decoding exist once than land a column that starts lying the day LAGs merge. Happy to take you up on the shared walkL2(): if you put up the main.js and l2.js part alongside your LAG work, I will rebase this PR onto it once the trunk representation has settled, and the per-port grouping stays here as you sketched. Good catch on the two getL2 bugs, that was accidental honesty on my part. I will mark this a draft until then; ping me when the helper lands.

logicog:
> So my suggestion: if @logicog wants this convenience on the ports page at all

Let me comment on that: I consider this a community project. There are a handful of contributors who can merge. If you can convince anyone among those that it should be merged, so will it be and I will never question that. Anyone who contributes, especially as a reviewer, for a couple of months gets merge privileges (there will be a discussion behind the scenes). If a reviewer says  something needs to be merged, there will always be someone among those who have merge rights who will do it, if they agree.

DrDoof:
@bloqaudio the helper is up as #344. It stacks on #340 for now, so only the last commit there is the helper itself.

One difference from what I sketched: the callback takes a second argument, `onDone(entries, ok)`. Your `entries => devRender(entries)` still works unchanged, but you probably want to skip the redraw when `ok` is clear. A failed page is retried at the same index three times before the walk gives up that way, so a clear `ok` means the table really is incomplete rather than merely late, and rendering it would blank ports that do have something behind them. I only noticed this because my first version ended the walk on the first bad reply, which quietly turned every timeout into a truncated table.

The part you were waiting on is the port decoding: it lives in `walkL2()` alone now, so when trunk entries start showing up in that field only that one function has to learn about them, and both pages follow.

The trunk representation itself still has not settled, so there is nothing to rebase onto in that sense yet. The transport half is there whenever you want to move your walk across.

DrDoof:
A correction to what I wrote above, since I got the mechanism wrong.

I said an L2 entry for a MAC behind a trunk points at the trunk, so `logToPhysPort[e.port]` returns garbage and those entries silently vanish from the column. That is not what happens. The entry holds a physical port and always has: the vendor SDK stores `l2Table.spa = rtk_switch_port_L2P_get(port)` and reads it back through `rtk_switch_port_P2L_get()`, and there is no trunk id, no `isTrunk` flag and no branch selecting between the two anywhere in the entry. The field is even split the same way we split it, two low bits in one word and the rest beside them.

So nothing disappears. What happens instead is worse for a per port column: the address is learned on whichever member the frame arrived on and moves between members as the far side's hash sends it, so the same device shows up behind port 7, then behind port 8, and the column never says the two are one link.

That makes the case for decoding this in one place stronger rather than weaker, so my suggestion stands, just for the right reason.

bloqaudio:
Rebased onto the merged walkL2(): the transport is gone from ports.js, the walk result is used only when ok is set, and the 15 second restart lives in the callback. Tested on a SWTGW218AS on current main: the column populates through the helper, a port with 55 devices behind it shows the count with the full list in the tooltip, and the L2 page on the same build fills completely and stays stable across refresh cycles. During one page load a request failed transiently and the helper retried it without disturbing either table, which is the ok flag doing exactly what it was added for. Marking this ready for review.

vDorst:
thanks

PR #338, opened 2026-08-14
port: make the trunk hash default reachable again, and per group

DrDoof:
Split out of #331, as @logicog you asked. It sits on top of the macro change rather than on main, because it only means anything once the guard in front of the hash default actually works. The first commit shown here is #331 and will drop out of this PR when that one lands.

One thing about the order, since the split creates it. #331 on its own restores the guard, and the test it guards is written against zero, which this register never reads: it comes out of reset holding source port number plus both MAC fields, both IP fields and the L4 source port. So between #331 landing and this landing, the default hash stops being installed anywhere, where before it was installed on every call. Taken together they are a fix; taken one at a time in that order there is a gap.

Three things about the line the guard sits in front of:

The test was against zero, and the register does not read zero. Measured on an SWTGW218AS, where all four groups read `0x3f` after a cold boot and a value written before a power cycle is gone afterwards. Testing against the reset value restores what the line meant, and zero is still accepted in case another device does reset that way.

The write went to the base address while the read that decides it used the group offset, so a group other than zero was tested and group zero was written. Both ends use the offset now.

The range check printed a complaint and carried on. It returns, which matters more now that the hash write also builds an address out of the group number.

Thirty bytes of BANK1 on top of #331, nothing anywhere else. Built for SWTGW218AS and KP_9000_6XHML_X2 on sdcc 4.5.0.


vDorst:
thanks

PR #339, opened 2026-08-14
Fix parallel build

plaes:
Bunch of changes so `make -j20` works now :)

When this gets merged, I would like to tackle the "always-rebuild" issue.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -110,10 +110,10 @@ distclean:
 	-rm -f html_data.c html_data.h $(VERSION_HEADER)
 	-rm -rf $(BUILDDIR)
 
-$(BUILDDIR)/%.rel: %.c
+$(BUILDDIR)/%.rel: %.c | create_build_dir html_data.h
 	$(CC) -MMD $(CC_FLAGS) -o $@ -c $<
 
-$(BUILDDIR)/%.rel: %.asm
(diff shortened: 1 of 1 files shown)


Makefile
vDorst, line 1: I am a Makefile n00b, but the description of 9f2d6b7
trigger me, because I have the feeling that the underlying issue was not fixed.
How I interpret it based on your description, is that the task `$(VERSION_HEADER):` is called multiple times in parallel. Not just one task that is blocking the other tasks that depends on this task.

Seeing that the command writes the file once, is way better.

plaes: It might be that this is not the correct solution, but If you run current main branch in parallel, then you might spot the "missing endif" warning.
If think that there's also another dependency issue in other targets.

vDorst: According to AI, the problem is that `$(VERSION_HEADER)` is used in `.PHONY`.
So Make is regenerating this file every time for every build-step.
AI had no real fix what feels like `that is the right way to do it`.

logicog:
I tested my favorite `make -j12`, which always failed, and it works in deed. Aslo that the makefile now only runs once for html is amazing! Thanks so much for fixing this! Looking at your Makefile-tricks was an eye-opener! I verified that the generated image is byte-identical.

PR #340, opened 2026-08-15
httpd: emit valid JSON from send_l2

DrDoof:
The MAC table listing writes its separator once per iteration instead of once per object. When the table engine reports an entry as invalid no object gets written, but the comma still does, so JSON.parse chokes on what comes back and the whole table fails to load rather than just the row that was missing.

I didn't rewrite the function to test it. I pulled send_l2 out of both this branch and main verbatim, drove both with the same scripted table engine and put the output through JSON.parse:

| stimulus | before | after |
|---|---|---|
| empty table | `[,]`, SyntaxError | `[]`, 0 objects |
| every third entry invalid | SyntaxError | 20 objects |
| all entries valid | 2302 bytes, 30 objects | 2302 bytes, 30 objects |

That last row is the one worth checking: identical to the byte, so the restructure leaves the path that already worked alone.

The second commit touches the page and it isn't optional. Once the comma is fixed an empty table answers `[]`, and l2.js reads `s[s.length-1].idx` without checking the length, so the failure would just move from JSON.parse to a TypeError instead of going away. The same file also compares the entry array itself against a number rather than its length, so the 4096 cap never fires; a 5000 element array compares false.

There's a precedence bug in the index of an invalid entry as well, `h | low + 1`, where the addition binds first. I enumerated all 4096 combinations and it differs from `(h | low) + 1` in eight of them, only when the low byte reads 0xff and bit 8 of the index is already set. In those the walk goes back to the start of the current block of 256 instead of moving on to the next one.

I added a bound check on the output buffer for consistency with send_vlanlist further down the file. Thirty entries of at most 74 bytes plus the brackets fit the 2500 byte buffer with 179 to spare, so nothing changes today, but that margin wasn't written down anywhere and L2_MAX_TRANSFER is a tunable.

Costs 58 bytes of BANK1, nothing in BANK2, xdata or internal RAM. Built for SWTGW218AS and KP_9000_6XHML_X2 on sdcc 4.5.0. I haven't tested it on hardware: with a populated table the invalid branch never runs, which is probably why nobody has hit this yet.

### DIFF
--- a/httpd/page_impl.c
+++ b/httpd/page_impl.c
@@ -356,6 +356,7 @@ void send_l2(uint16_t idx)
 	 */
 	__xdata uint16_t entry = idx & 0xfff;
 	__xdata uint16_t first_entry = 0xffff; // An illegal entry index
+	bool first = true;
 	char_to_html('[');
 	while (1) {
 		entries_left--;
@@ -369,9 +370,22 @@ void send_l2(uint16_t idx)
 		} while (sfr_data[3] & TBL_EXECUTE);
 
 		reg_read_m(RTL837x_L2_DATA_OUT_B);
-		if ((sfr_data[0] & 0x20)) {	// Check entry is valid
+		bool valid = (sfr_data[0] & 0x20) != 0;
+		if (valid) {
+			/* separator + 74-byte worst-case entry + closing "]" */
+			if (slen + 76 > TCP_OUTBUF_SIZE)
+				break;
+			if (!first)
+				char_to_html(',');
+			first = false;
+
+			// VLAN, taken from the read above instead of reading the register twice
+			slen += strtox(outbuf + slen, "{\"vlan\":\"");
+			charhex_to_html(sfr_data[0] & 0x0f);
+			byte_to_html(sfr_data[1]);
+
 			// MAC
-			slen += strtox(outbuf + slen, "{\"mac\":\"");
+			slen += strtox(outbuf + slen, "\",\"mac\":\"");
 			byte_to_html(sfr_data[2]); char_to_html(':');
 			byte_to_html(sfr_data[3]); char_to_html(':');
 			port = (sfr_data[0] >> 6) & 0x3;
@@ -380,12 +394,6 @@ void send_l2(uint16_t idx)
 			byte_to_html(sfr_data[1]); char_to_html(':');
 			byte_to_html(sfr_data[2]); char_to_html(':');
 			byte_to_html(sfr_data[3]);
-
-			// VLAN
-			slen += strtox(outbuf + slen, "\",\"vlan\":\"");
-			reg_read_m(RTL837x_L2_DATA_OUT_B);
-			charhex_to_html(sfr_data[0] & 0x0f);
-			byte_to_html(sfr_data[1]);
 
 			// type
 			reg_read_m(RTL837x_L2_DATA_OUT_C);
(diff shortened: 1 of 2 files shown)


httpd/page_impl.c
```diff
@@ -356,6 +356,7 @@ void send_l2(uint16_t idx)
 	 */
 	__xdata uint16_t entry = idx & 0xfff;
 	__xdata uint16_t first_entry = 0xffff; // An illegal entry index
+	uint8_t first = 1;
```
vDorst: This can be just a `bool` or `__bit`

httpd/page_impl.c
vDorst: Because the json is position indepentend and this function have to interact the with SOC a lot, we can reduce a extra `reg_read_m(RTL837x_L2_DATA_OUT_B);`.

So we can start the `vlan` first, because this data is already in `sfr_data`.

```c
			// VLAN
			slen += strtox(outbuf + slen, "{\"vlan\":\"");
			charhex_to_html(sfr_data[0] & 0x0f);
			byte_to_html(sfr_data[1]);
	
			// MAC
			slen += strtox(outbuf + slen, "\",\"mac\":\"");
            <rest of the code>
```

httpd/page_impl.c
```diff
@@ -408,20 +415,14 @@ void send_l2(uint16_t idx)
 			entry += 1; // We want the next entry following after the current entry
 		} else {
 			reg_read_m(RTL837x_TBL_DATA_0);
-			entry = (((uint16_t)sfr_data[2] & 0x0f) << 8) | sfr_data[3] + 1;
+			entry = ((((uint16_t)sfr_data[2] & 0x0f) << 8) | sfr_data[3]) + 1;
 		}
```
vDorst: Can we put the entry handleing in one place
```c
		// Index
		reg_read_m(RTL837x_TBL_DATA_0);
		entry = (((uint16_t)sfr_data[2] & 0x0f) << 8) | sfr_data[3];
		if (is_valid) {
			slen += strtox(outbuf + slen, ",\"idx\":\"");
			byte_to_html(entry >> 8);
			byte_to_html(entry);
			char_to_html('"');
			char_to_html('}');
		}
		entry += 1; // We want the next entry following after the current entry
```

httpd/page_impl.c
vDorst, line 365: I think this can be simplified with
```c
		sfr_data[2] |= (TBL_LUTREAD_NEXT_L2UC << 6);
		sfr_data[3] &= 0xfc;
		reg_write_m(RTL837x_TBL_DATA_0);
```

httpd/page_impl.c
```diff
@@ -371,6 +372,12 @@ void send_l2(uint16_t idx)
 		reg_read_m(RTL837x_L2_DATA_OUT_B);
 		if ((sfr_data[0] & 0x20)) {	// Check entry is valid
 			// MAC
```
vDorst: This comment also have to move after `first = 0`.

DrDoof:
Thanks, that was worth more than the byte count suggests. Four of the five are in.

`first` is a `bool` now, the comment moved, and the VLAN goes first and takes its data from the same `L2_DATA_OUT_B` read that decides whether the entry is valid, so the second read of that register is gone. The entry handling is in one place too: the index is read once after the branch, `entry += 1` happens once, and only the `,"idx":"` part stays inside the valid case.

That last one turned out to be more than a tidy up. The index expression existed in both arms and the precedence bug I was fixing lived in the second copy, so with one copy there is nowhere left for them to disagree.

Together they take the patch from 58 bytes of BANK1 down to 5, with BANK2, xdata and internal RAM still untouched, same on SWTGW218AS and KP_9000_6XHML_X2.

I reran the harness on all three versions, main, the branch as you reviewed it, and the branch now. Empty table 53 bytes and `[]`. Every third entry invalid, 20 objects in 1552 bytes. All valid, 30 objects in 2302 bytes. The 0xff case still asks for 0x200 rather than 0x100. Identical to what you reviewed apart from the key order, which the page does not care about since it addresses the fields by name.

The one I left out is the `REG_WRITE` on line 365, and I want to check I am not misreading you. `REG_WRITE(r, v24, v16, v8, v0)` assigns in that order down from `SFR_DATA_24` to `SFR_DATA_0`, and `sfr_data[0]` is the most significant byte, so what is there today clears the low two bits of the 23:16 byte, while

```c
sfr_data[3] &= 0xfc;
```

clears the low two bits of the 7:0 byte. Those are different fields, so I did not want to swap one for the other inside a patch about JSON. It is also not a line this PR introduces: the same write appears again in `l2_delete` further down.

If you do mean the mask belongs on the low byte, then that is a bug in main rather than a simplification, and I will send it as its own patch covering both places. Just say which and I will do it.

vDorst:
> The one I left out is the REG_WRITE on line 365, and I want to check I am not misreading you. REG_WRITE(r, v24, v16, v8, v0) assigns in that order down from SFR_DATA_24 to SFR_DATA_0, and sfr_data[0] is the most significant byte, so what is there today clears the low two bits of the 23:16 byte, while

> sfr_data[3] &= 0xfc;

I made a typo, it should be 3 instead of 1.

I do like what you are doing and fixing up good issues. But can you please reduce the length of your responses, I feel I have to read a lot of text before get to the point.

I would like to suggest only reporting the code size when it matters, like big or structural changes. Most time it is not. Maybe github action can help up reporting to codesize and memory-usage inside a runner, so if someone is interested, they can look it up.

DrDoof:
Thanks, both taken. #348 has the mask on `sfr_data[3]` in `send_l2()` and `l2_delete()`.

And noted on length. I will get to the point faster and drop the size figures unless something structural moves.

vDorst:
Seeing #348, I belief I made a typo again, original code was fine. Sorry.

original code:
```c
reg_read_m(RTL837x_TBL_DATA_0);
REG_WRITE(RTL837x_TBL_DATA_0, sfr_data[0], sfr_data[1] & 0xfc, sfr_data[2] | (TBL_LUTREAD_NEXT_L2UC << 6), sfr_data[3]);
```

This can be the suggested change.
```c
reg_read_m(RTL837x_TBL_DATA_0);
sfr_data[1] &= 0xfc;
sfr_data[2] |= (TBL_LUTREAD_NEXT_L2UC << 6);
reg_write_m(RTL837x_TBL_DATA_0);
```

PR #342, opened 2026-08-15
sfp: bound the wait for an I2C transfer

DrDoof:
You closed #327 because the bounds in it were illusory: every one sat on an outer loop calling reg_read_m(), which ends in its own unlimited wait on the register bus, so the hang just moved down a level. That was right. I'm opening this one because I think it's the case that argument doesn't cover, and that's what I'd like you to judge it on.

sfp_read_reg() polls the I2C completion bit with no limit, and it's reached from handle_sfp(), which the main loop calls right before handle_rx(). A module holding the clock line therefore stops packet handling, STP and the web interface along with the read that provoked it, and plugging the module in is enough to start it. The bus runs to a part the switch doesn't control and that a user can plug in, which is what makes this different from a table operation nobody has ever seen get stuck.

What it doesn't fix, said plainly because a bound that only looks like one is worse than none: reg_read_m() keeps its unlimited wait, so an ASIC with a dead register interface still hangs below this. They are separate failures though. SFR_EXEC_STATUS settles inside the chip and doesn't depend on the state of an I2C transfer, so the register reads keep completing while the transfer doesn't. That's exactly the case a module can cause, and the one this bounds.

The wait gives up after 65536 polls and returns 0xff, which the callers already treat as a failed transfer: sfp_apply_quirks() reads that value as either a failed I2C read or a voltage the spec doesn't allow, and concludes the module has no diagnostics. The generated code loops on a 16 bit counter in r6 and r7 and leaves through `mov dpl,#0xff`, which I checked in the assembly.

The count isn't a tuned figure and I haven't measured what it costs in time, which matters: handle_sfp() does six reads for one insertion and I have no number for how long a read takes on this part. So a module that wedges the bus buys a bounded but possibly long pause rather than an instant answer. That beats never returning, but it isn't the same as being quick about it, and if you'd rather see a smaller count or a different failure value, say so and I'll change it.

Costs 31 bytes of the common segment and nothing in either bank, xdata or internal RAM. Built for SWTGW218AS and KP_9000_6XHML_X2 on sdcc 4.5.0. Not tested on hardware, since provoking it needs a module that misbehaves.

### DIFF
--- a/rtlplayground.c
+++ b/rtlplayground.c
@@ -1056,9 +1056,16 @@ uint8_t sfp_read_reg(uint8_t slot, uint8_t reg)
 	// Execute I2C Read
 	reg_bit_set(RTL837X_REG_I2C_CTRL, 0);
 
-	// Wait for execution to finish
+	/* Bound the wait. The bus runs to a module the switch does not control,
+	 * and this is reached from the main loop, so a module holding the clock
+	 * line stops packet handling, STP and the web interface with it. 0xff is
+	 * what the callers already treat as a failed transfer.
+	 */
+	uint16_t tries = 0;
 	do {
 		reg_read_m(RTL837X_REG_I2C_CTRL);
+		if (!++tries)
+			return 0xff;
 	} while (sfr_data[3] & 0x1);
 
 	reg_read_m(RTL837X_REG_I2C_OUT);


vDorst:
> A module holding the clock line therefore stops packet handling, STP and the web interface along with the read that provoked it, and plugging the module in is enough to start it. The bus runs to a part the switch doesn't control and that a user can plug in, which is what makes this different from a table operation nobody has ever seen get stuck.

I2C clock is always generated by the master, in this case the SOC. AFAIK SOC nor most I2C devices don't support i2c-feature `clock stretching`. Many I2C-master don't have the hardware to detect this kind of bus-faults. So even is clock is to low (shorted to gnd), the master keeps on going and should always finish because of the number of bytes the master needs to read. Only point that the master can see that there is a issue with the slave, is when the address is send, because the slave need to send a `ACK`. Otherwise the master can't tell the difference.

So I think this should not happen and also the SDK just does this, waiting for that trigger-bit 0. What the SDK does better is checking the fail-bit 1.

I was planning to rewrite this function to fix
- Added error reporting
- Master can read/write up to 16 bytes per transaction.

Feel free to pick this up for me.

vDorst:
A quick test with this code
```c
uint8_t sfp_read_reg(uint8_t slot, uint8_t reg)
{
	if (reg & 0x80) {	// Configure SFP readings address (0x51) as I2C device address
		reg &= 0x7f;
		REG_WRITE(RTL837X_REG_I2C_CTRL, 0x00, 0x1 << (I2C_MEM_ADDR_WIDTH-16) | 0,  0x51 >> 5, (0x51 << 3) & 0xff);
	} else {
		REG_WRITE(RTL837X_REG_I2C_CTRL, 0x00, 0x1 << (I2C_MEM_ADDR_WIDTH-16) | 0,  0x50 >> 5, (0x50 << 3) & 0xff);
	}

	sfr_mask_data(1, 0xfc, i2c_bus_from_scl_pin(machine.sfp_port[slot].i2c.scl) << 5 | i2c_bus_from_sda_pin(machine.sfp_port[slot].i2c.sda) << 2);
	reg_write_m(RTL837X_REG_I2C_CTRL);

	REG_WRITE(RTL837X_REG_I2C_IN, 0, 0, 0, reg);

	// Execute I2C Read
	reg_bit_set(RTL837X_REG_I2C_CTRL, 0);

	uint16_t cnt = 0;

	// Wait for execution to finish
	do {
		reg_read(RTL837X_REG_I2C_CTRL);
		if (++cnt == 0) {
			print_string("I2C-timeout 0x");
			print_byte(SFR_DATA_0);
			write_char('\n');
			break;
		}
	} while (SFR_DATA_0 & 0x1);

	if (SFR_DATA_0 & 0x02) {
		print_string("I2C failure: cnt=");
		print_short(cnt);
		write_char('\n');
		return '?';
	}

	reg_read(RTL837X_REG_I2C_OUT);
	return SFR_DATA_0;
}
```

Shorted the clock-pin to ground
```
> sfp

Slot 1 - Rate: I2C failure: cnt=0x006b
3f  Encoding: I2C failure: cnt=0x006d
3f
I2C failure: cnt=0x0066
?I2C failure: cnt=0x0067
?I2C failure: cnt=0x0068
?I2C failure: cnt=0x006a
?I2C failure: cnt=0x0066
?I2C failure: cnt=0x0066
?I2C failure: cnt=0x0069
?I2C failure: cnt=0x006a
?I2C failure: cnt=0x0067
?I2C failure: cnt=0x0068
?I2C failure: cnt=0x006a
?I2C failure: cnt=0x0067
?I2C failure: cnt=0x0068
?I2C failure: cnt=0x0069
?I2C failure: cnt=0x006a
?I2C failure: cnt=0x0067
?I2C failure: cnt=0x0067
?I2C failure: cnt=0x006a
?I2C failure: cnt=0x0066
?I2C failure: cnt=0x0068
?I2C failure: cnt=0x0069
?I2C failure: cnt=0x006a
?I2C failure: cnt=0x0067
?I2C failure: cnt=0x0068
?I2C failure: cnt=0x006a
?I2C failure: cnt=0x0066
?I2C failure: cnt=0x0068
?I2C failure: cnt=0x0069
?I2C failure: cnt=0x006a
?I2C failure: cnt=0x0067
?I2C failure: cnt=0x0068
?I2C failure: cnt=0x006a
?I2C failure: cnt=0x0064
?I2C failure: cnt=0x0068
?I2C failure: cnt=0x0069
?I2C failure: cnt=0x006a
?
Options: I2C failure: cnt=0x0069
3f

Slot 2 - Rate: I2C failure: cnt=0x0067
3f  Encoding: I2C failure: cnt=0x006d
3f
I2C failure: cnt=0x0066
?I2C failure: cnt=0x0068
?I2C failure: cnt=0x0069
?I2C failure: cnt=0x0066
?I2C failure: cnt=0x0067
?I2C failure: cnt=0x0068
?I2C failure: cnt=0x006a
?I2C failure: cnt=0x0066
?I2C failure: cnt=0x0066
?I2C failure: cnt=0x0069
?I2C failure: cnt=0x0064
?I2C failure: cnt=0x0067
?I2C failure: cnt=0x0068
?I2C failure: cnt=0x006a
?I2C failure: cnt=0x0066
?I2C failure: cnt=0x0068
?I2C failure: cnt=0x0067
?I2C failure: cnt=0x006a
?I2C failure: cnt=0x0067
?I2C failure: cnt=0x0068
?I2C failure: cnt=0x006a
?I2C failure: cnt=0x0066
?I2C failure: cnt=0x0068
?I2C failure: cnt=0x0069
?I2C failure: cnt=0x0066
?I2C failure: cnt=0x0067
?I2C failure: cnt=0x0068
?I2C failure: cnt=0x006a
?I2C failure: cnt=0x0066
?I2C failure: cnt=0x0068
?I2C failure: cnt=0x0069
?I2C failure: cnt=0x0066
?I2C failure: cnt=0x0067
?I2C failure: cnt=0x0068
?I2C failure: cnt=0x006a
?I2C failure: cnt=0x0064
?
Options: I2C failure: cnt=0x0069
3f
Temp: I2C failure: cnt=0x0065
3fI2C failure: cnt=0x0063
3f
Vcc: I2C failure: cnt=0x006c
3fI2C failure: cnt=0x0065
3f
TX Bias: I2C failure: cnt=0x0064
3fI2C failure: cnt=0x006e
3f
TX Power: I2C failure: cnt=0x006a
3fI2C failure: cnt=0x0064
3f
RX Power: I2C failure: cnt=0x006b
3fI2C failure: cnt=0x0065
3f
Laser: I2C failure: cnt=0x006b
3fI2C failure: cnt=0x0065
3f
State: I2C failure: cnt=0x006b
3f
```

Clock-pin back to normal.
```
> sfp

Slot 1 - Rate: 67  Encoding: 00
OEM             <snip>  R   
Options: 00

Slot 2 - Rate: 67  Encoding: 06
DELL            <snip>        A0  
Options: 68
Temp: 204b
Vcc: 819e
TX Bias: 0e7c
TX Power: 16ea
RX Power: 0ff0
Laser: 0000
State: 30
```

Results when only i2c-data of slot 2 is shorted to ground.
Shows no failure, only zeros.
```
> sfp

Slot 1 - Rate: 67  Encoding: 00
OEM             <snip>  R   
Options: 00

Slot 2 - Rate: 00  Encoding: 00

Options: 00
Temp: 0000
Vcc: 0000
TX Bias: 0000
TX Power: 0000
RX Power: 0000
Laser: 0000
State: 00
```

So the master keeps going and reports a i2c-failure.

*Updated* the code and results, now show the `cnt` value to see how fast it fails.
Added the results of a shorted DATA line.

DrDoof:
You are right, and the shorted clock settles it. I had no way to produce that measurement myself and I should have asked for it before opening this, rather than arguing from the call path.

The premise is simply wrong. There is no clock stretching for the master to wait on, it finishes the transaction either way, and your counter shows it giving up at around 0x66 polls. A limit of 65536 guarding a loop that provably ends after about a hundred is 31 bytes spent on nothing, and it is the second time I have proposed a bound that does not correspond to a failure anyone can produce, after #327. Closing it.

What your run does show is the part worth having, and it is not a bound. The fail bit is never looked at, so a failed read comes back as a plausible value. The shorted data line is worse, since it raises no failure at all and returns zeros that read exactly like a module answering with zeros.

I will pick up the rewrite you sketched, error reporting plus up to sixteen bytes per transaction, and send it as its own PR rather than reopening this one. Thanks for taking the time to short the pins.

vDorst:
> The premise is simply wrong. There is no clock stretching for the master to wait on, it finishes the transaction either way, and your counter shows it giving up at around 0x66 polls. 

Hardware bugs do exists with strange edge cages with only pop-up when multiple things happens at once. For example [STM32l443 errata](file:///home/r/Downloads/es0318-stm32l433xx443xx-device-errata-stmicroelectronics.pdf).

> The shorted data line is worse, since it raises no failure at all and returns zeros that read exactly like a module answering with zeros.

This can be checked. Before every transmission, check if both GPIO-levels are high to see that the bus is idle. But for now, it is a lot of extra checks with no real gain. If the bus is stuck the use see that here is no SFP update.

PR #343, opened 2026-08-15
port: pin the management MAC to the CPU port with a static L2 entry

bloqaudio:
An idle switch eventually stops answering on its management IP. The CPU's own MAC exists in the L2 table only as a learned entry, created whenever the CPU transmits, and it ages out like any other dynamic entry. Unknown unicast is not flooded to the CPU port, so once the entry expires, frames for the management IP no longer reach the CPU until the next CPU transmission (usually an ARP reply) re-teaches it, and then it ages out again. Active web sessions keep the entry alive continuously, which is why this mostly goes unnoticed: it bites exactly when you have not touched the switch for a while and then cannot reach it.

The fix is a static entry for the management MAC on the CPU port, installed after the startup config has been replayed so it lands in the final management VLAN, and moved when the management VLAN is changed at runtime.

Getting the entry into the table took some reverse engineering, and the commit message records what was established on hardware: a table write with read-method 0 hashes the key in TBL_DATA_IN and places the entry itself, since a search returns index 0 on a miss and there is no software-visible insertion slot. The age field in the third data word must be non-zero or the entry is invisible to lookup entirely, and the static flag is bit 0 of byte 1 of that word: in a paired experiment an entry with the flag held full age indefinitely while an identical entry without it aged to invisible within the aging period. Note this is a different bit than the L2 status page currently decodes as static, so the pinned entry displays as learned; correcting that display would be a separate change.

Tested on a SWTGW218AS against current main: after eight minutes of enforced silence (well past the observed aging time, with the ARP entry pinned on the client so no broadcast could wake the CPU) the switch answered 20 of 20 unicast pings, where unpatched firmware loses nearly all of them. The same test passes after moving the management VLAN away and back at runtime, and the move leaves exactly one CPU entry in the table with no stale leftover.

### DIFF
--- a/rtl837x_port.c
+++ b/rtl837x_port.c
@@ -911,3 +911,57 @@ bool port_ingress_vlan_filter_get(uint8_t port) __banked
 
 	return reg_bit_test(RTL837X_VLAN_PORT_IGR_FLTR, port);
 }
+
+// C1 bit 0: static (no aging); age at C3[4:2] must be non-zero or the
+// entry is invisible to lookup
+static void l2_mgmt_entry_fill(__xdata uint8_t *mac, __xdata uint16_t vlan)
+{
+	REG_WRITE(RTL837x_TBL_DATA_IN_A, mac[2], mac[3], mac[4], mac[5]);
+	REG_WRITE(RTL837x_TBL_DATA_IN_B, 0x20 | ((CPU_PORT & 0x3) << 6) | (vlan >> 8),
+		  vlan & 0xff, mac[0], mac[1]);
+	REG_WRITE(RTL837x_TBL_DATA_IN_C, 0x00, 0x01, 0x00, (7 << 2) | (CPU_PORT >> 2));
+}
+
+static void l2_mgmt_tbl_prepare(void)
+{
+	reg_read_m(RTL837x_TBL_DATA_0);
+	sfr_data[2] &= 0x3f;
+	sfr_data[1] &= 0xf8;
+	reg_write_m(RTL837x_TBL_DATA_0);
+}
+
+static void wait_table_ready(void)
+{
+	do {
+		reg_read(RTL837X_TBL_CTRL);
+	} while (SFR_DATA_0 & TBL_EXECUTE);
+}
+
+/** Pin the management MAC to the CPU port; remove_entry deletes it */
+void port_l2_static_mgmt(__xdata uint8_t *mac, __xdata uint16_t vlan, __xdata bool remove_entry) __banked
+{
+	wait_table_ready();
+
+	// vlan 0 = management VLAN disabled: untagged management resolves in
+	// VLAN 1; clients in other VLANs keep their learned entry and age as before
+	if (!vlan)
+		vlan = 1;
+	l2_mgmt_entry_fill(mac, vlan);
+	l2_mgmt_tbl_prepare();
+	if (remove_entry) {
+		// a search miss returns index 0, not a slot
+		REG_WRITE(RTL837X_TBL_CTRL, 0x00, 0x00, TBL_L2_UNICAST, TBL_EXECUTE);
+		wait_table_ready();
+		reg_read_m(RTL837x_TBL_DATA_0);
+		if (!(sfr_data[2] & 0x10))
+			return;
+		sfr_data[1] |= 0x04; // CLEAR-entry
+		reg_write_m(RTL837x_TBL_DATA_0);
+		__xdata uint16_t idx = (((uint16_t)sfr_data[2] & 0x0f) << 8) | sfr_data[3];
+		REG_WRITE(RTL837X_TBL_CTRL, idx >> 8, idx & 0xff, TBL_L2_UNICAST, TBL_WRITE | TBL_EXECUTE);
+	} else {
+		// method-0 write: the hardware hashes the key and places the entry
+		REG_WRITE(RTL837X_TBL_CTRL, 0x00, 0x00, TBL_L2_UNICAST, TBL_WRITE | TBL_EXECUTE);
+	}
+	wait_table_ready();
+}
--- a/rtlplayground.c
+++ b/rtlplayground.c
@@ -2306,6 +2306,8 @@ void main(void)
 	early_boot_handle_button();
 
 	execute_config();
+	// After the config so the entry lands in the final management VLAN
+	port_l2_static_mgmt(uip_ethaddr.addr, management_vlan, false);
 	/* After the config: a name from it wins, otherwise derive one. */
 	set_hostname_default();
 	print_cmd_prompt();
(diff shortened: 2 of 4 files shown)


rtl837x_port.c
```diff
@@ -839,3 +839,54 @@ bool port_ingress_vlan_filter_get(__xdata uint8_t port) __banked
...
+
+/** Pin the management MAC to the CPU port; clear == 1 removes the entry */
+void port_l2_static_mgmt(__xdata uint8_t *mac, __xdata uint16_t vlan, __xdata uint8_t clear) __banked
+{
+	do {
+		reg_read_m(RTL837X_TBL_CTRL);
+	} while (sfr_data[3] & TBL_EXECUTE);
+
+	if (!vlan)
+		vlan = 1;
```
vDorst, line 948: What happens when `management_vlan` is `0`? Which is a valid value.

rtl837x_port.c
```diff
@@ -839,3 +839,54 @@ bool port_ingress_vlan_filter_get(__xdata uint8_t port) __banked
...
+static void l2_mgmt_tbl_prepare(void)
+{
+	reg_read_m(RTL837x_TBL_DATA_0);
+	sfr_data[2] &= 0x3f;
+	sfr_data[1] &= 0xf8;
+	reg_write_m(RTL837x_TBL_DATA_0);
+}
+
+/** Pin the management MAC to the CPU port; clear == 1 removes the entry */
+void port_l2_static_mgmt(__xdata uint8_t *mac, __xdata uint16_t vlan, __xdata uint8_t clear) __banked
```
vDorst: change `uint8t clear` to `bool remove_entry` so it more clear what it does.

I also wonder maybe just put `__xdata uint8_t *mac, __xdata uint16_t vlan, __xdata uint8_t clear` in a static struct.
This struct can also be used in `l2_mgmt_entry_fill()`.

rtl837x_port.c
```diff
@@ -839,3 +839,54 @@ bool port_ingress_vlan_filter_get(__xdata uint8_t port) __banked
...
+{
+	do {
+		reg_read_m(RTL837X_TBL_CTRL);
+	} while (sfr_data[3] & TBL_EXECUTE);
```
vDorst: Maybe put this in a seperate function, because it is called at least 3 times.

```c
void wait_table_ready(void) {
	do {
		reg_read_m(RTL837X_TBL_CTRL);
}
```

rtlplayground.c
```diff
@@ -2206,6 +2206,8 @@ void main(void)
 	early_boot_handle_button();
 
 	execute_config();
+	// After the config so the entry lands in the final management VLAN
+	port_l2_static_mgmt(uip_ethaddr.addr, management_vlan, 0);
```
vDorst: Is this needed? Because `execute_config()` already calls `vlan mgmt x`.

vDorst (changes requested):

rtl837x_port.c
bloqaudio: With 0 the entry is installed in VLAN 1, since lookup is per VLAN and untagged management resolves there. A client managing from another VLAN in that mode still gets a learned entry from its ARP exchange and ages out as before, so the no-aging guarantee covers the VLAN 1 path and behavior elsewhere is unchanged from before this patch. I can note that in the code if you want it recorded.

rtl837x_port.c
bloqaudio: Taken, will push with the next update. I would keep the two parameters as they are, two values across two call sites did not seem worth a struct, but say so if you feel strongly.

rtl837x_port.c
bloqaudio: Done, pushed together with the rename.

rtlplayground.c
bloqaudio: It is needed for configurations that contain no vlan mgmt line at all, which never run that path during replay. When the line is present the second install rewrites the same entry, which the hash write makes harmless.

vDorst:
Please rebase, and also check your other PRs to see if there are any conflicts.

bloqaudio:
Rebased onto main (4c8e0ce) and re-tested on hardware. #324 was the only other open one and is rebased as well; its firmware image is byte-identical to main.

rtl837x_port.c
vDorst: Yes, Please add a note.

rtl837x_port.c
```diff
@@ -911,3 +911,55 @@ bool port_ingress_vlan_filter_get(uint8_t port) __banked
...
+	sfr_data[2] &= 0x3f;
+	sfr_data[1] &= 0xf8;
+	reg_write_m(RTL837x_TBL_DATA_0);
+}
+
+static void wait_table_ready(void)
+{
+	do {
+		reg_read_m(RTL837X_TBL_CTRL);
+	} while (sfr_data[3] & TBL_EXECUTE);
```
vDorst: To reduce the `sfr-data`-copy overhead, this can be:
```c
void wait_table_ready(void) {
	do {
		reg_read(RTL837X_TBL_CTRL);
	} while (SFR_DATA_0 & TBL_EXECUTE);
```

vDorst:
I saw there was a question I didn't answer.
Yes please add the note and update `wait_table_ready()`.

rtl837x_port.c
bloqaudio: Added.

rtl837x_port.c
bloqaudio: Done, it is 2 bytes smaller as well.

logicog:
The code looks good to me. I will leave it to @vDorst  to merge, though.
I have one comment, which is just informative: using functions once only like l2_mgmt_tbl_prepare(void) may still make for nicer code, but always comes with the cost of using up space on the tiny stack. Here it is not problematic, as it is only used in the context of the management vlan command, which is very shallow in the work loop of the device. But the situation may be different when serving web-pages with all the uip layer and page-handling functions already on the stack. That's why functions just for nicer code are usually avoided, sometimes even when this means repeating code: flash is relatively cheap but the stack is tiny. Although someone figured out that the 8051 in these devices has the possibility to use a stack via SFR registers in xmem, which may come in handy for special things like a regexp parser implemented on the stack.

vDorst:
Currently you can still remove the CPU entry on the website.
Do we want to block that?
So the only way to change that entry is maybe by a new function `change_cpu_port_mac()`.

This can also be in a other PR.

bloqaudio:
@vDorst Agreed, it should be refused. Today a delete just drops the switch back to a learned entry until the next `vlan mgmt` or reboot, so nothing breaks, but the guarantee is gone. I would do it as a follow-up PR: `l2_delete()` already reads the entry's MAC before rewriting it, so it is a 6-byte compare against `uip_ethaddr` there, inline, no new function. `l2 forget` is fine, it flushes dynamic entries only.

bloqaudio:
@logicog Thanks, noted for the httpd path. The helpers here only run from the console and boot paths.

bloqaudio:
Follow-up for the web delete is #393.

vDorst:
Thanks

PR #344, opened 2026-08-15
html: move the L2 table walk into a shared walkL2() helper

DrDoof:
This is the main.js and l2.js half of what I sketched on #335, put up on its own as I offered there. @bloqaudio is holding that PR as a draft waiting for it, so that the port decoding of /l2.json lives in one place before trunk entries start turning up in that field.

It stacks on #340. The first two commits here are that PR and only the last one is new, so review just `html: move the L2 table walk into a shared walkL2() helper`. Once #340 lands this rebases down to that single commit. Happy to sit on it until then if reviewing both at once is awkward.

`walkL2(onDone)` pages through the table once, parses idx and vlan out of hex, maps the port to a physical number or to `'CPU'`, and calls `onDone(entries, ok)`. It stops on a wrapped index, an empty page or 4096 entries, all of which set `ok`. l2.js keeps the s and l to label mapping, since that one needs the page's translations, redraws only when `ok` is set, and restarts the walk from its own callback either way. ports.js in #335 would use the same helper with its 15 second gap, keeping the per port grouping there.

@bloqaudio, one difference from the sketch: the callback takes a second argument. `entries => devRender(entries)` still works unchanged, but you probably want to skip the redraw when `ok` is clear, for the reason in the third point below.

Three things change while moving, and they are the parts worth arguing about:

The next request goes out from the previous reply rather than from a `setInterval` that fires whether or not the last one came back. With one connection served at a time, a timer that outruns the responses only queues work it cannot use.

A walk that reaches 4096 entries now hands over what it collected. Before it dropped the entries and cleared its own interval, so the page could not refresh again until it was reloaded.

A page that comes back as anything other than 200, or with a body `JSON.parse` rejects, is asked for again at the same index, up to three times, and only then does the walk end with `ok` clear. That is not a new idea, it is the old behaviour written down: the previous code ignored a non 200 and let the interval ask for the same index again, so a blip never disturbed the table on screen. I had this ending the walk instead, until I read my own test output properly and saw that it meant every timeout redraws the page with a truncated table. At one connection at a time and a 1500 ms timeout that is not a rare event.

I drove the helper itself in node against a scripted server rather than testing a copy of it:

| case | entries | requests | asked for | ok |
|---|---|---|---|---|
| empty table | 0 | 1 | 0 | yes |
| three pages, last repeats index 0 | 61 | 3 | 0, 30, 60 | yes |
| empty page after one full page | 30 | 2 | 0, 30 | yes |
| 500, then a good page | 31 | 4 | 0, 30, 30, 31 | yes |
| malformed body, then a good page | 31 | 4 | 0, 30, 30, 31 | yes |
| three failures in a row | 30 | 4 | 0, 30, 30, 30 | no |
| 4096 entries in one page | 4096 | 1 | 0 | yes |
| CPU port | decodes to `'CPU'` | | | |
| vlan and idx | come back as numbers | | | |

main.js grows by 1331 bytes and l2.js loses 930, so 401 bytes of flash against main. Worth saying plainly where they land, because it is the part I would understand pushback on: nine pages pull in main.js, and exactly one of them walks the L2 table today, two once #335 lands. So seven or eight page loads now carry 1331 bytes they will not use, served by an httpd that does one connection at a time.

If that trade looks wrong, the alternative is a third file that only l2.html and ports.html include. I did not do that because main.js is where the sketch on #335 put it and where logToPhysPort and sendXHTTP already live, but it is a small change if you would rather.

Not tested on hardware yet.

### DIFF
--- a/html/l2.js
+++ b/html/l2.js
@@ -1,7 +1,3 @@
-var l2GetInterval;
-var l2Entries = [];
-var l2CurrentEntry = 0;
-
 function fillStats() {
   var tbl = document.getElementById('statstable');
   if (!numPorts)
@@ -118,7 +114,6 @@ function fillL2(s)
   s = uniq(s);
   l2All = s;
   renderL2();
-  l2Entries = [];
 }
 
 function paintL2(tbl, s)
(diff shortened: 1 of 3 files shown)


vDorst:
Thanks

PR #345, opened 2026-08-15
sfp: notice when an I2C read fails

DrDoof:
@vDorst this is the first half of what you offered to hand over in #342, and I would rather send it in two pieces than guess at the second.

Your shorted clock run settled that the wait does not need bounding, and it also showed the thing that does need fixing: `sfp_read_reg()` reads the output register whatever the outcome, so an address nothing acknowledged comes back as an ordinary byte. Bit 1 is now checked, a failure sets `sfp_i2c_fail` and the read returns `0xff`. That value is already what `sfp_apply_quirks()` treats as either a failed transfer or a voltage the spec forbids, so that existing test starts being true when it should be. The insertion path and the `sfp` command clear the flag first and report afterwards instead of printing the bytes as if the module had sent them.

Two things I left out on purpose.

It detects but does not act. Skipping `sds_config()` when the rate read failed is the obvious next move, and I did not make it: a module that raises the bit spuriously would then never be configured at all, which is worse than what happens now, and I cannot tell from here how trustworthy the bit is. You can, so I would rather you decide that one.

It does not do the sixteen bytes per transaction. doc/sfp.md documents only the single byte path and does not name a length field, and the register write in the current code that looks like it might be one, `0x1 << (I2C_MEM_ADDR_WIDTH - 16)` into bits 23:16 of `RTL837X_REG_I2C_CTRL`, I can only guess at. If you tell me which field carries the transfer length and how the extra bytes come out of `RTL837X_REG_I2C_OUT`, I will write that part too. Guessing at a register I cannot test is how #342 went wrong and I would rather not repeat it.

Your shorted data line result is worth recording somewhere, by the way, because it is the case neither the flag nor a bound catches: no failure is raised at all and the read returns zeros that look exactly like a module answering with zeros.

40 bytes of the common segment, 51 of BANK2 and 1 of xdata, nothing in BANK1 or internal RAM. Built for SWTGW218AS and KP_9000_6XHML_X2 on sdcc 4.5.0.

Not tested on hardware, and it is easy for you to check: shorting the clock line as you did in #342 should now print the failure line rather than a plausible looking byte.


vDorst:
I would like to see that the `sfp_read_reg()` returns a `sfp_i2c_fail` instead of the I2c-value.

For the length question

Are you familiar with the linux SDK for this chip? See https://github.com/JiaY-shi/rtl83xx/blob/main/src/rtl8372/dal/rtl8373/dal_rtl8373_i2c.c#L243
This code controls the chip via the SlaveInterface. This is were we get some more inside how the chip works. This only shows a part how the switch chip works not the 8051 core.
`dal_rtl8373_i2c_data_read()` is the function that readsout the i2c device.

Datalength is set here: https://github.com/JiaY-shi/rtl83xx/blob/main/src/rtl8372/dal/rtl8373/dal_rtl8373_i2c.c#L263

> It detects but does not act. Skipping sds_config() when the rate read failed is the obvious next move, and I did not make it: a module that raises the bit spuriously would then never be configured at all, which is worse than what happens now, and I cannot tell from here how trustworthy the bit is. You can, so I would rather you decide that one.

I am not sure that this answers your question, but we can check the CRC of the SFP-data.
Ofcourse there are exception and quircks see linux kernel: [SFP check](https://github.com/torvalds/linux/blob/master/drivers/net/phy/sfp.c#L2533) [Quirk Example 1](https://github.com/torvalds/linux/blob/master/drivers/net/phy/sfp.c#L2419)

I think we want a statemachine per SFP-slot.
When we detect a SFP module
1. Wait x mS before accessing.
2. Retry it after a delay
3. Successful, validate the checksum and then, configure the rest of the system
   Failure: Report an error and keep the sds in low-power/idle mode.

But this is way to much for now.
I would like to see that the function just can detect errors and read more than one byte to speedup the transaction.

> Your shorted data line result is worth recording somewhere, by the way, because it is the case neither the flag nor a bound catches: no failure is raised at all and the read returns zeros that look exactly like a module answering with zeros.

You can look GPIO-input before every transmission, to see if the bus is idle.

DrDoof:
Thanks, that gives me the shape: `(memAddrWidth << 4) | (dataWidth - 1)`, a fail bit in the same register, data in consecutive registers.

One thing I cannot settle from that file, since the numeric masks are in the header: which of ours is `I2C_MST1_CTRL1`? We have `RTL837X_REG_I2C_MST_IF_CTRL` 0x0414, `RTL837X_REG_I2C_CTRL` 0x0418 and `RTL837X_REG_I2C_CTRL2` 0x041c, and only 0x0418 is used today. Our code puts the memory address width at bit 20 of 0x0418 where the SDK has it at bit 4, and `sfr_mask_data(1, 0xfc, scl << 5 | sda << 2)` already takes bits 23:21 and 19:18 of that register for the bus, which leaves no room for a four bit data width. Which register and which bits carry DATA_WIDTH and I2C_FAIL?

Meanwhile I am converting the callers to `sfp_read(slot, reg, buf, len)` returning an error instead of a value, as you asked.

vDorst:
List from [here](https://github.com/JiaY-shi/rtl83xx/blob/main/src/rtl8372/dal/rtl8373/rtl8373_reg_definition.h#L3760)
```c
#define RTL8373_I2C_MST1_CTRL1_ADDR                     (0x418)
#define RTL8373_I2C_MST1_CTRL1_READ_MODE_OFFSET       (22)
#define RTL8373_I2C_MST1_CTRL1_READ_MODE_MASK         (0x1 << RTL8373_I2C_MST1_CTRL1_READ_MODE_OFFSET)
#define RTL8373_I2C_MST1_CTRL1_MEM_ADDR_WIDTH_OFFSET  (20)
#define RTL8373_I2C_MST1_CTRL1_MEM_ADDR_WIDTH_MASK    (0x3 << RTL8373_I2C_MST1_CTRL1_MEM_ADDR_WIDTH_OFFSET)
#define RTL8373_I2C_MST1_CTRL1_DATA_WIDTH_OFFSET      (16)
#define RTL8373_I2C_MST1_CTRL1_DATA_WIDTH_MASK        (0xF << RTL8373_I2C_MST1_CTRL1_DATA_WIDTH_OFFSET)
#define RTL8373_I2C_MST1_CTRL1_SCL_OUT_SEL_OFFSET     (13)
#define RTL8373_I2C_MST1_CTRL1_SCL_OUT_SEL_MASK       (0x7 << RTL8373_I2C_MST1_CTRL1_SCL_OUT_SEL_OFFSET)
#define RTL8373_I2C_MST1_CTRL1_SDA_OUT_SEL_OFFSET     (10)
#define RTL8373_I2C_MST1_CTRL1_SDA_OUT_SEL_MASK       (0x7 << RTL8373_I2C_MST1_CTRL1_SDA_OUT_SEL_OFFSET)
#define RTL8373_I2C_MST1_CTRL1_DEV_ADDR_OFFSET        (3)
#define RTL8373_I2C_MST1_CTRL1_DEV_ADDR_MASK          (0x7F << RTL8373_I2C_MST1_CTRL1_DEV_ADDR_OFFSET)
#define RTL8373_I2C_MST1_CTRL1_RWOP_OFFSET            (2)
#define RTL8373_I2C_MST1_CTRL1_RWOP_MASK              (0x1 << RTL8373_I2C_MST1_CTRL1_RWOP_OFFSET)
#define RTL8373_I2C_MST1_CTRL1_I2C_FAIL_OFFSET        (1)
#define RTL8373_I2C_MST1_CTRL1_I2C_FAIL_MASK          (0x1 << RTL8373_I2C_MST1_CTRL1_I2C_FAIL_OFFSET)
#define RTL8373_I2C_MST1_CTRL1_I2C_TRIG_OFFSET        (0)
#define RTL8373_I2C_MST1_CTRL1_I2C_TRIG_MASK          (0x1 << RTL8373_I2C_MST1_CTRL1_I2C_TRIG_OFFSET)
```

>  Thanks, that gives me the shape: (memAddrWidth << 4) | (dataWidth - 1), a fail bit in the same register, data in consecutive registers.

> One thing I cannot settle from that file, since the numeric masks are in the header: which of ours is I2C_MST1_CTRL1? We have RTL837X_REG_I2C_MST_IF_CTRL 0x0414, RTL837X_REG_I2C_CTRL 0x0418 and RTL837X_REG_I2C_CTRL2 0x041c, 

`I2C_MST1_CTRL1` = `RTL8373_I2C_MST1_CTRL1_ADDR` = 0x418.

> and only 0x0418 is used today. Our code puts the memory address width at bit 20

That seems correct see above

> of 0x0418 where the SDK has it at bit 4,

I am thinking that you assume that `rtl8373_setAsicRegBits()` writes the full registers or the bits at that location.
But is only sets the bits based on the MASK and `rtl8373_setAsicRegBits()` shifts the bits also based on the mask as well.

> and sfr_mask_data(1, 0xfc, scl << 5 | sda << 2) already takes bits 23:21 and 19:18 of that register for the bus, which leaves no room for a four bit data width. Which register and which bits carry DATA_WIDTH and I2C_FAIL?

See definition above.


RTL837X_REG_I2C_MST_IF_CTRL = 0x414 is also used in `setup_i2c()`.

I hope this helps you a bit more.

DrDoof:
That settles both. `DATA_WIDTH` at 19:16 is the low nibble of the 23:16 byte and nothing else touches it, so it has been asking for one byte all along. And `I2C_FAIL` at bit 1 confirms what this PR already checks, which I had only from your shorted clock run.

I misread `sfr_mask_data()` earlier when I said the bus selection left no room: it indexes `sfr_data[3 - n]` and its mask says which bits to clear, so it lands on 15:10, exactly your SCL and SDA fields.

Multi byte read and the `sfp_read(slot, reg, buf, len)` signature coming together, since there is nothing left to guess at.

vDorst:
> Multi byte read and the sfp_read(slot, reg, buf, len) signature coming together, since there is nothing left to guess at.

I think that there are too many arguments for our target?
Possible solution could be:
- Fix buffer location. Like we do with `sfr_data`.
- Using a fix settings-struct to provide the arguments.

DrDoof:
@vDorst Your fixed buffer is the one I want, and the register list you posted settles why. The data sits in four consecutive registers from 0x424, with `DATA_CTRL_ADDR(index)` capped at index 15 and the width field only four bits wide, so a transfer is at most sixteen bytes and always lands in the same place. A fixed `sfp_data[16]` filled by `reg_read_m()` mirrors that one to one, the way `sfr_data` already does, and the pointer argument goes away with it. That leaves `sfp_read(slot, reg, len)` returning the error rather than a value, as you asked earlier.

I would rather not quote a byte figure yet, because the cost turns out to follow from how a parameter is declared and not just from how many there are. `sfp_read_reg` keeps `PARM_2` in data and setting it is three bytes, while `memset` declares its second parameter `__xdata` and the same store costs five there. So I will compile both shapes and send real numbers with the code.

One correction to what I wrote earlier in this thread. I said our `sfr_mask_data(1, ...)` takes bits 23:21 and 19:18 and leaves no room for a four bit data width. That was wrong. The helper indexes from the other end, so it writes bits 15:13 and 12:10, which is exactly your SCL and SDA offsets, and it leaves the two low bits of that byte for the device address. Bits 19:16 were free all along.

DrDoof:
I have pushed two more commits onto this branch. They touch the same function, so a separate PR would have shown this one's change again in its own diff. The title still says only what the first commit does, so it now covers less than the branch.

The second commit moves the transfer into the banked pins module, which frees 200 bytes of the common segment on every board. The third reads up to sixteen bytes per transaction instead of one, so inserting a module costs 8 I2C transactions instead of 87.

Tested on a switch with a Finisar 10G module: same vendor, model, serial, rate and encoding as before, and no I2C failures. The diagnostics block is untested, because this module reports no DDM.

Happy to rename this or split the extra commits out again, whichever suits you. Thanks!

vDorst:
This is already looking great.

I am debating whether we should split `sfp_read_block()` up, see details in collapsed part.

Review on the current PR.

- I think we should pre-computing the I2C-settings `i2c_bus_from_scl_pin(machine.sfp_port[slot].i2c.scl) << 5 | i2c_bus_from_sda_pin(machine.sfp_port[slot].i2c.sda) << 2)`.
  So with a extra helper that created this value at compile-time so we have to carry only 1 byte around.
  This helper/value can also be used in the machine-struct.

- I wonder if we can just combine all the write `RTL837X_REG_I2C_CTRL` just one transaction.
  
  So first write the `address`-data
  
  ```c
  REG_WRITE(RTL837X_REG_I2C_IN, 0, 0, 0, reg);
  ```
  
  Then all the settings including the start trigger in `RTL837X_REG_I2C_CTRL`.

- I see this also in other places in the code and I am in favor to change it. I am planning to make a PR for the entire codebase.
  When we only need to read or write from/to SOC, use SFR-registers `SFR_DATA_x` directly to save extra work and reducing the use of CPU-registers.

  So:
  ```c
    // Wait for execution to finish
	do {
	} while (sfr_data[3] & 0x1);
	```
  Can be:
	```c
    // Wait for execution to finish
	do {
		reg_read(RTL837X_REG_I2C_CTRL);
	} while (SFR_DATA_0 & 0x1);
	```

- 	Error value
	```c
    /* Bit 1 is the controller's own failure indication, which the vendor SDK
	 * looks at and this did not. Without it an unacknowledged address comes
	 * back as an ordinary byte and the caller cannot tell it from data. */
	if (sfr_data[3] & 0x2) {
		sfp_i2c_fail = 1;
		memset(dst, 0xff, len);
		return;
	}
	```
	Return error via the return value seems sufficient for me, so `sfp_i2c_fail` can be dropped.
	Changing the memory seems unnecessary.
	If a caller depends on that, than the caller should handle it by itself.

- I think `sfp_read_block()` can use `sfp_buf` directly.
  So we don't need to pass the buffer pointer and also coping the data direct from SFR to `sfp_buf`.
  All to reduce overhead.
  
 ```c
	for (uint8_t i = 0; i < len; i++) {
		if (!(i & 0x3))
			reg_read_m(RTL837X_REG_I2C_OUT + i);
		dst[i] = sfr_data[3 - (i & 0x3)];
	}
 ```

 ```c
	__xdata uint8_t *dst = &sfp_buf;
	uint8_t val;
	for (uint8_t i = 0; i < len; i++) {
		switch(i & 0x3) {
			case 0:
				reg_read(RTL837X_REG_I2C_OUT + i);
				val = SFR_DATA_0;
				break;
			case 1:
				val = SFR_DATA_8;
				break;
			case 2:
				val = SFR_DATA_16;
				break;
			default:
				val = SFR_DATA_24;
				break;
		}
		
		*dst++ = val;
	}
 ```



--- 
<details>

<summary>Technically it is a i2c write read transaction.</summary>

I think we should split `sfp_read_block()` up, so the I2C part is separated.

Like
```c
bool i2c_write_read(uint8_t bus, uint8_t i2c_addr, uint8_t __xdata uint8_t *dst, uint8_t len) {};
```
But I think is gets complicated with function arguments only.
Because we can send up to 3-bytes in the write phase and we may want to use its full potential.

When using a struct, that makes it possible.

```c
Struct {
	uint8_t bus;
	uint8_t i2c_addr;
	uint8_t write_len; // Or a enum
	uint8_t write_data[3];
	uint8_t read_len; // This is not limited to 16.
	__xdata uint8_t *read_buffer;
}
```

- Were `bus` is already precomputed value `i2c_bus_from_scl_pin(machine.sfp_port[slot].i2c.scl) << 5 | i2c_bus_from_sda_pin(machine.sfp_port[slot].i2c.sda) << 2)`.
  So with a extra helper that created this value at compile-time so we have to carry only 1 byte around.
  This helper/value can also be used in the machine-struct.

- `i2c_addr` I2C device address in 7-bit notation. So without the R/W-bit.

- `write_len` number of bytes to write 0-3. `I2C_MEM_ADDR_WIDTH`

- `write_data` bytes to write 0-3.

- `read_len` number of bytes to read.

- `dst` pointer to the buffer. But I think this can be discarded, just make direct use to the `sfp_buf`.
  Then rename the buffer `sfp_buf` to `i2c_buf`.

- `len` number of bytes to read. This is not limited to 16, because we can simply write_data with 16 and start a new transaction.

- Return value, a bool that indicated an error, like invalid length value or NACK.
  We can return a enum to better report an error.

Let me know if we want this and when? This can be a separate PR.

</details>

DrDoof:
Thank you, that is a helpful review. All of it is in, force pushed.

The memory address goes out first, so the device address, the bus selection and the start bit fit in one write to the control register. The reads take their result from the SFRs directly, `sfp_read_block()` fills `sfp_buf` and returns a bool, and `sfp_i2c_fail` is gone.

One measurement you may want: carrying the result back through the callers costs 92 bytes of the common segment and 3 of internal RAM, against the single xdata byte the flag used. Still ahead of where the branch started, so I left it as you asked.

Retested on a switch with a Finisar module and found no issues.

Yes please to the generic `i2c_write_read()`, and I am happy to write it as a separate PR once this one lands. Thanks!

vDorst:
> One measurement you may want: carrying the result back through the callers costs 92 bytes of the common segment and 3 of internal RAM, against the single xdata byte the flag used. Still ahead of where the branch started, so I left it as you asked.

I see that `_sfp_get_info_ok_10000_367`, `_handle_sfp_ok_50001_389` `_sfp_print_measurements_ok_10000_348` are eating SRAM.

- `sfp_print_measurements()`/`parse_sfp` can both return early and print a `i2c-error message`, because they are CLI-only. So `ok`-variable is not needed anymore.

- In `handle_sfp()` it should stop try to go further, because it is also making decision based on the data. So I think it should stop on a error, print a message and mark the SFP as unplugged. So that in the next `handle_sfp()`-cycle it will retry it.
Technically we could buffer the SFP-data 64+32-bytes per device, both blocks, check the CHECKSUM of the data to ensure they are valid, do some touch-ups for the stings were needed, and then continue to setup the sds.
In that case all the basic data is cached and validated and only the DOM/DDM data is readout in real-time.
More info about the data inside SFP module: https://github.com/torvalds/linux/blob/master/include/linux/sfp.h#L6, and how linux kernel check it https://github.com/torvalds/linux/blob/master/drivers/net/phy/sfp.c#L2574

- `sfp_get_info()`, I think in every step we should abort/return early.
  With the caching, in the previous item, this is not a issue anymore.

- `sfp_read_block()`:
  - Simplify the length check by decrease at the beginning.
  - It seems that `sfp_buf[i] = val;` generated smaller code than `*dst++`.
    So the function can looks like this. Saves around 60-bytes.

```c
   bool sfp_read_block(uint8_t slot, uint8_t reg, uint8_t len) __banked __reentrant
   {
	len--;
	if (len > 15)
		return false;

	uint8_t dev = (reg & 0x80) ? 0x51 : 0x50;	// 0x51 holds the diagnostics, 0x50 the module data
	reg &= 0x7f;

	REG_WRITE(RTL837X_REG_I2C_IN, 0, 0, 0, reg);

	REG_WRITE(RTL837X_REG_I2C_CTRL, 0x00,
		  0x1 << (I2C_MEM_ADDR_WIDTH - 16) | len,
		  (dev >> 5) | i2c_bus_from_scl_pin(machine.sfp_port[slot].i2c.scl) << 5
		  | i2c_bus_from_sda_pin(machine.sfp_port[slot].i2c.sda) << 2,
		  ((dev << 3) & 0xff) | 0x1);

	do {
		reg_read(RTL837X_REG_I2C_CTRL);
	} while (SFR_DATA_0 & 0x1);

	if (SFR_DATA_0 & 0x2)
		return false;

    uint8_t val;
    for (uint8_t i = 0; i < len; i++) {
		switch(i & 0x3) {
			case 0:
				reg_read(RTL837X_REG_I2C_OUT + i);
				val = SFR_DATA_0;
				break;
			case 1:
				val = SFR_DATA_8;
				break;
			case 2:
				val = SFR_DATA_16;
				break;
			default:
				val = SFR_DATA_24;
				break;
		}
		sfp_buf[i] = val;
    }

	return true;
}
```

DrDoof:
All in, thank you. The three `ok` variables are gone: every caller gives up on the first failed read, and the module read moved out of `handle_sfp()` into its own function so it can. A failed module is left marked absent, so the next poll retries it instead of configuring the SerDes from bytes that never arrived.

`sfp_read_block()` is 74 bytes smaller with your version, and internal RAM now has one byte more free than before the series. One thing to watch in the snippet: after `len--` the read loop needs `i <= len`, otherwise it reads one byte short.

Retested on a switch with a Finisar module and found no issues.

The caching with checksum validation sounds right to me. Shall I fold it into the `i2c_write_read()` PR, or keep it separate? Thanks!

vDorst:
> One thing to watch in the snippet: after len-- the read loop needs i <= len, otherwise it reads one byte short.

Good catch. I was wondering why it was half broken while testing my code while reviewing.

Both DAC and 10G-Fiber module with DOM reports fine.

Forcing i2c to fail while inserting the module.
This recovers fine.
```
<MODULE INSERTED>  Slot: 1SFP: an I2C read failed, retrying on the next poll

<MODULE INSERTED>  Slot: 1SFP: an I2C read failed, retrying on the next poll

<MODULE INSERTED>  Slot: 1  Rate: 67  Encoding: 00  Module: OEM             SFP-H10GB-CU3M  R   

sds_config sds: 01, mode: 1a

RTL837X_REG_SDS_MODES: 0x0000035a
```

> The caching with checksum validation sounds right to me.

You can add it to this PR if you want.

Otherwise I merge this PR.

DrDoof:
Please merge it as it is.

The caching deserves its own discussion: how much to keep per module, when to re-read it, and what a bad checksum should do to a module that is otherwise working. It also leans on the same transaction shape as `i2c_write_read()`, so I would rather bring the two together in the next PR than grow this one.

Thank you for testing the DAC and the fibre module with DOM, and for forcing the I2C failure. Those are the two paths I had no way to reach here: my only module reports no DDM, and I was not going to break the bus on a live uplink.

logicog:
Thanks for working so hard on this improvement!

vDorst:
> The caching deserves its own discussion: how much to keep per module, when to re-read it, and what a bad checksum should do to a module that is otherwise working. It also leans on the same transaction shape as i2c_write_read(), so I would rather bring the two together in the next PR than grow this one.

Yes I agree.

> Thank you for testing the DAC and the fibre module with DOM, and for forcing the I2C failure. Those are the two paths I had no way to reach here: my only module reports no DDM, and I was not going to break the bus on a live uplink.

No problem.

Thanks for the work.

PR #346, opened 2026-08-15
crtstart: put the bank switching helpers in HOME

DrDoof:
One line, and a latent one: nothing is wrong today, which is the reason it is easy to keep.

`__sdcc_banked_call` and `__sdcc_banked_ret` are assembled into GSFINAL. SDCC's own crtbank.asm declares the area order and then puts both in HOME, and GSFINAL is a poor place to keep them because it sits in the startup path: GSINIT ends exactly where GSFINAL begins, in both builds, so the processor falls into that area rather than being sent there.

It works at the moment only because crtstart.rel comes after every C object on the link line, so the GSFINAL fragment holding `LJMP __sdcc_program_startup` is laid down first and the helpers land behind it:

| | GSFINAL | `__sdcc_banked_call` |
|---|---|---|
| before | 0x00F8, 21 bytes | 0x00FB, in GSFINAL |
| after | 0x010A, 3 bytes | 0x0094, in HOME |

Reordering that line, or moving `main()` into another module, is enough to put the helper body at the fallthrough address instead. The processor would run `push _PSBANK`, `xch a,r0` and the rest, then return to whatever the stack held. That is a board which does not come up, out of a build that reports nothing wrong.

The image changes in more places than eighteen bytes, and it is worth saying why rather than leaving it to be found. Relocating the two symbols re-targets every reference to them: `lcall 0x00FB` occurs 130 times before and not once after, `lcall 0x0094` 130 times after. `__sdcc_call_dptr` and the gsinit code sit behind the helpers in HOME and shift by the same eighteen bytes, so the reset vector moves from `02 00 96` to `02 00 a8`. Comparing symbol tables, 17 of 401 change address and all 17 are in that startup region. Nothing else lands anywhere new, and between the reset vector and 0x0094 not a single byte differs, so no interrupt vector is disturbed. Segment sizes: HOME 150 to 168, GSFINAL 21 to 3, and CSEG, BANK1, BANK2, xdata and internal RAM identical on both machines.

Run on a SWTGW218AS. One caveat on how, since it matters: what I flashed was this line applied to my Spanning Tree branch rather than to main alone. The switch's stored configuration turns STP on, and the implementation currently in main sets every port to blocking in `stp_setup()` and then never moves one back, `port_timers[]` being written once and never read, so a main based image would have taken the board off the network for reasons unrelated to this patch. The board came back after about 42 seconds reporting the new build, with its stored configuration byte identical, every link at the speed it had before and no address learned on two ports at once. The reset vector points eighteen bytes further along now and it still starts, which is the thing worth knowing.

I left crtstart.asm without the area order preamble that crtbank.asm carries, since the C objects already establish the order and the map confirms it. Happy to add it if you would rather the file stand on its own.


vDorst:
> I left crtstart.asm without the area order preamble that crtbank.asm carries, since the C objects already establish the order and the map confirms it. Happy to add it if you would rather the file stand on its own.

So you mean this? From [crtbank.asm](https://sourceforge.net/p/sdcc/code/HEAD/tree/trunk/sdcc/device/lib/mcs51/crtbank.asm)
```asm
	.area HOME    (CODE)
	.area GSINIT0 (CODE)
	.area GSINIT1 (CODE)
	.area GSINIT2 (CODE)
	.area GSINIT3 (CODE)
	.area GSINIT4 (CODE)
	.area GSINIT5 (CODE)
	.area GSINIT  (CODE)
	.area GSFINAL (CODE)
	.area CSEG    (CODE)
```

So it is better to rename the file `crtstart.asm` to `crtbank.asm` so it is the same as SDCC intended and add the part above to it?

It that is the case I am fine with that.

DrDoof:
Yes, exactly that. Renamed to `crtbank.asm` with the area order block, and the helpers still land in HOME at 0x0094 with GSFINAL down to the three byte jump, same as before the rename.

Worth a note for anyone reading it later: the order block leaves CSEG open, so the file has to re-open HOME before the symbols or they end up in a banked area.

vDorst:
Thanks

PR #348, opened 2026-08-15
httpd: mask the low byte of the table write, not byte two

DrDoof:
@vDorst spotted this in #340. `REG_WRITE` fills `SFR_DATA_24` down to `SFR_DATA_0`, so `sfr_data[1] & 0xfc` clears bits 17:16 of `TBL_DATA_0` where 1:0 was meant. `send_l2()` and `l2_delete()` both have it.

Not tested on hardware, and I do not know what those bits do. If clearing 17:16 turns out to be deliberate, say so and I will close this.

### DIFF
--- a/httpd/page_impl.c
+++ b/httpd/page_impl.c
@@ -361,7 +361,7 @@ void send_l2(uint16_t idx)
 		entries_left--;
 		uint8_t port = 0;
 		reg_read_m(RTL837x_TBL_DATA_0);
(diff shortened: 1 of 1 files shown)


DrDoof:
You are right, closing. I read "3 instead of 1" as the code being the typo rather than the comment, which is on me.

The rewrite you showed is worth having on its own and I will fold it into #340, where `send_l2()` already changes; `l2_delete()` has the same line.

PR #349, opened 2026-08-15
lag: number the groups from one and bound what the command is given

DrDoof:
`lag show` has always printed the groups as 1 to 4 while `lag <n>` took the number literally, so typing what you saw configured the group beside it. Both `lag` and `lag hash` count from one now, matching how ports are numbered everywhere else, and reject anything outside 1 to 4.

Two bounds go in with it. The port argument indexed `machine.phys_to_log_port`, which holds nine entries, before it was checked, and a two digit argument reaches 109. And `port_lag_members_set()` and `port_lag_hash_set()` complained about a group out of range and then wrote the registers anyway, past the four the groups occupy; they return now. `lag hash` also read `cmd_words_b[1]` without checking a word was there.

Worth knowing before merging: a saved `lag 1 ...` line moves from hardware group 1 to group 0. Nothing behaves differently, `show` reports it as 1 either way now, but a register dump taken across the change will look like the group shifted.

Not tested on hardware.

### DIFF
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -273,7 +273,9 @@ void parse_lag(void)
 
 	if (cmd_words_len < 2 || !isnumber(cmd_buffer[cmd_words_b[1]]))
 		goto err;
-	group = cmd_buffer[cmd_words_b[1]] - '0';
+	group = cmd_buffer[cmd_words_b[1]] - '1';
+	if (group > 3)		/* '0' wraps well past three, so one test does both ends */
+		goto err;
 
 	uint8_t w = 2;
 	while (w < cmd_words_len) {
@@ -283,7 +285,9 @@ void parse_lag(void)
 			port = cmd_buffer[cmd_words_b[w]] - '1';
 			if (isnumber(cmd_buffer[cmd_words_b[w] + 1]))
 				port = (port + 1) * 10 + cmd_buffer[cmd_words_b[w] + 1] - '1';
-				port = machine.phys_to_log_port[port];
+			if (port > 8)	/* phys_to_log_port holds nine entries */
+				goto err;
+			port = machine.phys_to_log_port[port];
 		} else {
 			goto err;
 		}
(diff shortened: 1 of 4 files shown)


vDorst:
This don't build anymore because SRAM shortage. #344 seems to consumes a bit extra SRAM.

This patch reclaim 2 extra SRAM-bytes, 1 SRAM is used to store more bits into.
Also tried other combination like store it in xdata. But the compiler puts `valid` in a bit anyway.
```patch
index 7b12927..bf3e268 100644
--- a/httpd/page_impl.c
+++ b/httpd/page_impl.c
@@ -356,7 +356,7 @@ void send_l2(uint16_t idx)
         */
        __xdata uint16_t entry = idx & 0xfff;
        __xdata uint16_t first_entry = 0xffff; // An illegal entry index
-       bool first = true;
+       __bit first = true;
        char_to_html('[');
        while (1) {
                entries_left--;
@@ -370,7 +370,7 @@ void send_l2(uint16_t idx)
 
-               bool valid = (sfr_data[0] & 0x20) != 0;
+               __bit valid = (sfr_data[0] & 0x20) != 0;
                if (valid) {
                        /* separator + 74-byte worst-case entry + closing "]" */
                        if (slen + 76 > TCP_OUTBUF_SIZE)
```

With this patch, this PR can build again.
I think #330 needs more focus.

DrDoof:
@vDorst Applied, thanks. Measured here it gives three bytes back rather than two, with SSEG going from 131 to 134 on SWTGW218AS. I rebased the branch on current main first, so the patch sits on top of that.

I cannot reproduce the failure itself though. Both machines build for me on sdcc 4.5.0, before and after your patch. So your toolchain is finding the wall and mine is not, and it would help to know which sdcc you are on, because that decides how much internal RAM the rest of us think is left.

vDorst:
Thanks

My machine is `SWGT024_V2_0_MANAGED`. 
sdcc 4.5.0 from debian 13.

DrDoof:
@vDorst Same 4.5.0 here, build #15242. With the branch merged I get byte-identical internal RAM on both machines now, 134 bytes of stack either way, so whatever separated our builds went away with the bit-memory move. Thanks for closing the loop.

PR #351, opened 2026-08-15
port: let a pvid name an aggregation group

DrDoof:
@logicog In #347 I set out three steps towards treating an aggregate as one interface, and said the second of them, a port to group lookup, would come when something in the tree needed it. Looking back at that, I had it the wrong way round. Both callers we have ask which ports are in a group, and that is the part which landed. Nothing asks which group a port is in, so a standalone patch for the lookup would have added a function with no caller. It only earns its place next to something that uses it, so here it is with one.

The problem it fixes is real today. VLAN membership, PVID, egress tagging, ingress filtering, isolation, the learning constraint and MTU are all per physical port in this ASIC, and nothing stops a member of a working group being given a different PVID from its peers. The group then forwards asymmetrically depending on which member the hash picks, with no part of the firmware saying a word about it.

PVID is the smallest of those settings, so it is the one I brought. A port outside a group keeps the path it had. A port inside one now applies to the whole group, which also means the saved configuration replays the same way whichever member it names.

The open question is what the rest should do. My inclination is the same shape for the others, and for a member configured individually to be accepted and applied to the group rather than refused, since refusing turns an ordinary command into an error for reasons the user cannot see. Say if you would rather it refused, or would rather the whole direction stayed out of the tree.

Built for SWTGW218AS and KP_9000_6XHML_X2 on sdcc 4.5.0. Not run on hardware.

### DIFF
--- a/rtl837x_port.c
+++ b/rtl837x_port.c
@@ -87,10 +87,9 @@ vlan_ingress_mode_t port_ingress_filter_get(__xdata uint8_t port) __banked
 /*
  * Define a Primary VLAN ID for a port 
 */
-void port_pvid_set(uint8_t port, __xdata uint16_t pvid) __banked
+static void port_pvid_write(uint8_t port, __xdata uint16_t pvid)
 {
 	// r4e1c:00001001 R4e1c-000017d0 r6738:00000000 R6738-00000000 (no filtering)
-	print_string("\nport_pvid_set called \n");
 	uint16_t reg = RTL837x_PVID_BASE_REG + ((port >> 1) << 2);
 
 	reg_read_m(reg);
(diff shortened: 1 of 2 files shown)


vDorst:
Can we assume in a LAG setup, that all the ports have the same settings, like speed etc. So it is just an extra but the same type of link?
Maybe better question, what that be different between the ports and what must be the same.

I am thinking for the settings that must be the same, error out and report that it is a lag-setup.
Modify the commands to accept `lx` `x = lag 1..4` as valid `port` which applies the settings to all the lag members.
I am not sure if that is easy to do, because we want it also on the WWW-side.

DrDoof:
Gentle nudge on this one, it has been open a couple of weeks without a look.

I rechecked it against current main today: it still merges cleanly, and it costs nothing in the common segment, 1111 bytes free both before and after, same on every board I built.

It is 27 lines across 2 files, so it should not take long to read. If the approach looks wrong to you I am happy to rework it, or to close it if you would rather not carry the feature.

Thanks for your time.

vDorst:
@DrDoof I see that you missed my question on reaction of your commit question below.

> The open question is what the rest should do. My inclination is the same shape for the others, and for a member configured individually to be accepted and applied to the group rather than refused, since refusing turns an ordinary command into an error for reasons the user cannot see. Say if you would rather it refused, or would rather the whole direction stayed out of the tree.

I reviewed the code and looks good.
We can merge this and move the discussion to a new issue.

DrDoof:
Sorry, our comments crossed. Yours landed while I was writing the nudge, so it read as if I had walked straight past it.

On what has to match: the hash picks a member per frame, so anything that changes how a frame is treated has to be identical across members or the behaviour becomes a coin toss. That is VLAN membership and pvid, ingress filtering and isolation, and speed and duplex. What never touches a frame can differ safely, EEE and the LEDs among them. None of it is enforced today, the group is only a member mask and a hash setting.

Your `lx` idea reads better to me than refusing. It is the same shape this PR gives pvid, so the commands would end up consistent rather than one command knowing about groups.

Both halves are written up in #377.

vDorst:
> Sorry, our comments crossed.

No problem!

Thanks for this, and for making a new issue.

PR #352, opened 2026-08-16
health: a CLI dump of where the main loop's time goes

DrDoof:
This adds a `health` command to the serial console. It prints a one page snapshot of the firmware's vital signs: the main loop pass counter next to the tick counter, per phase timing of the loop (link poll, SFP, RX, TX, STP, command dispatch), a frame counter from handle_rx, a painted stack high water mark, the whole TCP connection table and the httpd transfer state. Everything is raw hex, and two dumps taken a known time apart give you rates.

I wrote it in the middle of a real incident and it changed the debugging completely. The web interface of my SWTGW218AS had died and stayed dead, and everything I could measure from the outside was ambiguous: ping was slow, HTTP timed out, and every probe disturbed the thing it was measuring. One dump over the serial console settled it in minutes. The loop was running at full speed, all phases were clean, and the single TCP slot sat in ESTABLISHED with nothing in flight, held by a client that had left without closing. You cannot get that diagnosis from the outside, because from out there a starved socket and a stalled CPU look identical.

Since then I reach for it constantly. It caught retransmission tails eating the slot after every close, its RX rate ruled out a broadcast storm hypothesis in one reading, and after every change it confirms that the loop still makes one pass per tick. The stack watermark quietly answers a question I used to have no way of asking.

The implementation sits in cmd_parser next to the other command handlers, no new files, about 1.1 kB of code in bank 2 plus a few bytes of xdata for the counters. There is a doc page with a sample dump and notes on reading it, including the honest detail that printing the dump inflates its own cmd phase counters.

Proven in daily use on a SWTGW218AS as part of a larger tree. This standalone form builds clean for SWTGW218AS and SWGT024_V2_0_MANAGED.


feelfree69:
Just gave it a try, works as expected and is considered to be useful for debugging-purposes.
Regarding all the counter values: Woudn't it be more understandable to print decimals intead of hex-values?

One question remains for me:
```
> health
up 000001ef ticks 0x00018a69 loops 0x00017f6c rx 0x038a
link max 01 slow 0x0000
sfp  max 6c slow 0x0001
rx   max 06 slow 0x0077
tx   max 01 slow 0x0000
stp  max 04 slow 0x0001
cmd  max 0e slow 0x000c
sp 71 untouched 6f
tcp st 00 tmr 04 rtx 00 rto 04 len 0x0000 mss 0x05b4
  lport 0x0050 rport 0xe37a rip 0xc0a8 0x0af0
httpd left 0x087b entry 12 stp 01 mvlan 0x0001
```
How do I have to interpret the httpd line?

feelfree69:
There are new warnings at this branch:
```
sdcc -MMD -mmcs51 -I. -Ihttpd -Iuip -DMACHINE_SWTGW218AS -o output/SWTGW218AS/cmd_parser.rel -c cmd_parser.c
cmd_parser.c:2034: warning 126: unreachable code
cmd_parser.c:2035: warning 126: unreachable code
cmd_parser.c:2031: warning 110: conditional flow changed by optimizer: so said EVELYN the modified DOG
```

DrDoof:
Thanks for trying it out, and for catching the warnings.

Fixed in 3ebe050. The tcp dump printed the slot index behind `if (UIP_CONNS > 1)`, which with a single connection is a constant-false condition wrapped around live code, so sdcc flagged the body twice as unreachable and the once-only loop as flow it had rewritten. It prints the one slot directly now, with the assumption written as a build-time check, so a build that raises UIP_CONNS stops right here instead of quietly calling every line slot zero. Warning-free again, machine_check included.

On the httpd line, it is the state of the file transfer in progress:

- `left` is how many bytes of the current file still have to go out. httpd sends a file across several segments, so this is what is left of it; 0x0000 means nothing in flight.
- `entry` is the index into the embedded file table, so it says which file that is. 0xff means nothing matched, which is what a dynamic endpoint or a 404 looks like.
- `stp` and `mvlan` are just `stp_enabled` and the management VLAN, there so a pasted report tells you whether STP was on and which VLAN management was living in.

So your `left 0x087b entry 12` was mid-transfer of file 0x12 with 2171 bytes still to go, while an idle switch shows `left 0x0000 entry ff`.

On decimals I would rather stay with hex. Half of these fields are not quantities but bit patterns and identifiers (`st` is the TCP state flags, plus `mss`, the ports, `rip`, `mvlan`), and those read better in hex, so going decimal would mean two bases in one dump and having to remember which field is which. `print_byte` and `print_short` are also the cheap primitives here.

Two things I learned pointing this at hardware, both worth knowing before trusting the counters:

- A command that arrives over HTTP runs inside the RX handler, so its cost shows up under `rx`, not `cmd`. Only what you type on the console lands in `cmd`. Running `l2` over HTTP moved `rx` from 8 to 23 ticks while `cmd` stayed at 1.
- `health_phase` clamps a phase at 255 ticks, so 1.275 s is the ceiling. Anything slower just reads 0xff, and you cannot tell 1.3 s from 9 s.

logicog:
I am wondering, whether  this feature should not be compile-time dependent. Similar frameworks are typically only optionally enable on an OS such as Linux (lock-debugging, performance measurement and the likes). The reason is that it adds complexity and resource usage while not being useful to most of the users, as users need to have a very deep understanding of the underlying software. For developer use or when a user needs to report readings for debugging, then this could always be enabled and compiled in.

DrDoof:
Thanks for the nudge, and sorry it took me a week to come back to it. I had been running `health` every day for a month, and somewhere along the way it stopped feeling like a probe and started feeling like a feature. Your comment made me look at it the way a user would, and from there the answer was obvious.

So it is opt-in now: `make MACHINE=... HEALTH=1`. Without the flag the hooks in the main loop compile to nothing and the command does not exist; I checked that a default image is byte for byte the one main produces. With the flag, SWTGW218AS spends about 1.1 kB of bank 2, 100 bytes of the common segment and 28 bytes of xdata on it. I also let CI build once with the flag on, because code behind a switch nobody sets is the first thing to quietly stop compiling. Rebased on current main while I was at it.

@feelfree69 thanks again for trying it and for the warnings; those went away with the rebase.

feelfree69:
I don't know if the fault is on my side. Quickly compiled with HEALTH=1.

I don't see any measurements anymore:

```
> health
up 000001fc ticks 0x000193eb loops 0x00000000 rx 0x0000
link max 00 slow 0x0000
sfp  max 00 slow 0x0000
rx   max 00 slow 0x0000
tx   max 00 slow 0x0000
stp  max 00 slow 0x0000
cmd  max 00 slow 0x0000
sp 72 untouched 00
tcp st 00 tmr 78 rtx 00 rto 03 len 0x0000 mss 0x05b4
  lport 0x0050 rport 0xd13f rip 0xc0a8 0x0af0
httpd left 0x085c entry ff stp 01 mvlan 0x0001
```

feelfree69:
> I don't see any measurements anymore:

That is true for my RTL8372N-based 4+2 device.

My RTL8373-based 8+1 device still shows plausible numbers.

DrDoof:
Not your side, and the way I wrote the hooks is what hid it. Look at your own dump: every number fed by a hook is zero, while everything read live is fine. sp 72 and untouched 00 sit in the same line, one an SFR read, the other painted by a hook that never ran.

The hooks come from cmd_parser.h and expand to nothing without -DHEALTH. Dependency files track headers, not flags, so adding HEALTH=1 to a tree that was already built rebuilds cmd_parser.c and leaves rtlplayground.rel alone. The command exists, nothing increments the counters, and no symbol goes unresolved because the calls vanish rather than dangle.

Each MACHINE has its own output directory, which is why the other board looked fine.

Could you try rm -rf output/<your machine> and build again?

feelfree69:
> Could you try rm -rf output/ and build again?

Done, and everything's fine again. Thanks for the hint.

feelfree69 (approved):

Issue #334, opened 2026-08-14
`isolate` accepts the CPU port, the driver behind it does not


DrDoof:
`parse_isolate()` in cmd_parser.c takes a port number, allows 0 to 9, and treats 9 as the CPU port on purpose: there is a branch for the two digit form with the comment "CPU-port, logical port 9", and when it lists the current members it prints "CPU" for that bit rather than a front panel number.

The two functions it calls disagree with that. `port_isolation_get()` returns 0 for anything above `machine.max_port`, which on an 8+1 device is 8, so reading the isolation of the CPU port answers "no members" whatever the hardware holds. `port_isolate()` guards on the same bound, so setting it is declined.

The hardware does have the slot. `RTL837X_PORT_ISOLATION_BASE + (9 << 2)` is 0x50e4, and on an SWTGW218AS it reads 0x000003ff, the same all-members default as the eight front panel ports below it.

So a command that the parser accepts, documents in its own listing, and validates for, reaches a driver that quietly refuses it. Nothing reports an error, because both guards fail silently.

Worth saying how this surfaced, since it explains why nobody noticed. `REG_SET` is a run of statements without a do/while around it, so `if (port <= machine.max_port) REG_SET(...);` only guarded the first assignment and the write landed anyway, though with the top byte of the mask left over from an earlier register operation. Writing the CPU port therefore appeared to work while reading it returned zero. Once the macro is wrapped the two agree, and they agree on refusing. #331 wraps it, so this becomes visible the moment that lands, which is why I am raising it separately rather than deciding it there.

Either bound would be defensible, but they should be the same one, and the parser should match whichever is chosen. If the CPU port is meant to be isolatable, both guards want `CPU_PORT` rather than `machine.max_port`. If it is not, `parse_isolate()` should reject it up front instead of accepting it and printing "CPU" in a listing that can never contain anything.

vDorst:
I think in many cases we never want to include the CPU-port manual.
If the CPU-port needs to be involved because it provides a service, it should be added automatically based of the settings of the service.
Features like `mirroring` we don't want the CPU-port to be used, we don't want the CPU to be sniffed or CPU can't handle of the packet towards it. Also can't do much with these packets, we are not a linux-device. If a development needs this for easy debugging, lets make a `flag` and/or documentation so it can be enabled.
For `isolation` this is not needed, when does it should also be automatically.

DrDoof:
[...]

PR #354, opened 2026-08-17
port: reject the CPU port in isolate instead of refusing it silently

DrDoof:
This is the parser side of #334.

`parse_isolate()` took a two digit port and mapped it to logical port 9, the CPU port, while `port_isolate()` and `port_isolation_get()` both refuse anything above `machine.max_port`. Setting the isolation of the CPU port was declined without a word, and reading it always answered no members whatever the hardware held. @vDorst settled the direction in the issue: the CPU port should not be named by hand, so the parser now bounds it to the front panel.

The digit is checked before it indexes `phys_to_log_port[]` as well. `isolate x show` used to read well past that table.

I left the member list alone. It still takes `10` for the CPU port and `isolate <port> off` sets that bit on purpose, so dropping the manual form there means every isolate has to add the bit by itself. #326 rewrites that loop anyway.

### DIFF
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -428,11 +428,11 @@ void parse_isolate(void)
 
 	print_string("\nISOLATE ");
 
-	__xdata int8_t port_configured = cmd_buffer[cmd_words_b[1]] - '1';
-	port_configured = machine.phys_to_log_port[port_configured];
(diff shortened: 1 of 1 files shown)


vDorst:
Thanks

PR #356, opened 2026-08-17
httpd: buffer a configuration upload before touching the flash

bloqaudio:
Fixes #353.

Saving the configuration works in Chrome and fails in Firefox, and the difference is only how each browser splits the request. Two faults meet:

1. **The sector was erased as soon as the request arrived**, before the body was complete, so a body arriving in more than one piece corrupts what is stored.
2. **The parser worked one TCP segment at a time**, so it required every boundary and every part header to arrive whole. Firefox splits inside a part header, exactly at `filename="`, and from there the parser had lost its place: it never recognised the end of the upload, wrote nothing, and answered nothing, which is the timeout users see.

The configuration is limited to two kilobytes, so the whole request body now accumulates in xdata and is parsed once the closing boundary has arrived. The parts are walked in one pass, the part carrying a `filename` is written to a freshly erased sector, and the client gets a `200` instead of the previous silent close. No client segmentation can confuse this, since the parser only ever sees a complete body.

The firmware upload path still streams, since a megabyte cannot be buffered, and is untouched.

Tested on a SWTGW218AS with a configuration that crosses a flash page:

| request shape | before | after |
|---|---|---|
| split the way Firefox splits it | sector blank, no response | stored byte-identical, `200` |
| sent as one piece | stored, silent close | stored byte-identical, `200` |
| split three ways, one cut inside a boundary | sector blank, no response | stored byte-identical, `200` |

Issue #353, opened 2026-08-16
Saving configuration fails in Firefox


janh:
After saving the configuration in Firefox, it only contains garbage. Unlike Chrome, where the POST request returns quickly with empty response, it times out in Firefox.

**Packet captures:**

<details>
<summary>Firefox</summary>

```
00000000  50 4f 53 54 20 2f 63 6f  6e 66 69 67 20 48 54 54   POST /co nfig HTT
00000010  50 2f 31 2e 31 0d 0a 48  6f 73 74 3a 20 31 39 32   P/1.1..H ost: 192
00000020  2e 31 36 38 2e 31 30 2e  32 34 37 0d 0a 55 73 65   .168.10. 247..Use
00000030  72 2d 41 67 65 6e 74 3a  20 4d 6f 7a 69 6c 6c 61   r-Agent:  Mozilla
00000040  2f 35 2e 30 20 28 58 31  31 3b 20 4c 69 6e 75 78   /5.0 (X1 1; Linux
00000050  20 78 38 36 5f 36 34 3b  20 72 76 3a 31 35 33 2e    x86_64;  rv:153.
00000060  30 29 20 47 65 63 6b 6f  2f 32 30 31 30 30 31 30   0) Gecko /2010010
00000070  31 20 46 69 72 65 66 6f  78 2f 31 35 33 2e 30 0d   1 Firefo x/153.0.
00000080  0a 41 63 63 65 70 74 3a  20 2a 2f 2a 0d 0a 41 63   .Accept:  */*..Ac
00000090  63 65 70 74 2d 4c 61 6e  67 75 61 67 65 3a 20 64   cept-Lan guage: d
000000A0  65 2c 65 6e 2d 55 53 3b  71 3d 30 2e 39 2c 65 6e   e,en-US; q=0.9,en
000000B0  3b 71 3d 30 2e 38 0d 0a  41 63 63 65 70 74 2d 45   ;q=0.8.. Accept-E
000000C0  6e 63 6f 64 69 6e 67 3a  20 67 7a 69 70 2c 20 64   ncoding:  gzip, d
000000D0  65 66 6c 61 74 65 0d 0a  52 65 66 65 72 65 72 3a   eflate.. Referer:
000000E0  20 68 74 74 70 3a 2f 2f  31 39 32 2e 31 36 38 2e    http:// 192.168.
000000F0  31 30 2e 32 34 37 2f 73  79 73 74 65 6d 2e 68 74   10.247/s ystem.ht
[...]

PR #359, opened 2026-08-21
httpd: buffer a firmware upload's part header before streaming

bloqaudio:
Follow-up to #356 (and depends on it: this branch is stacked on `fix/config-post-race`). #353 turned out to have a second half, reported by janh: the firmware update over `POST /upload` also hangs in Firefox.

Same root cause as the configuration save, a client that splits the request inside a multipart part header. The upload body is already resumable, `stream_upload()` keeps `bindex` across TCP segments, but the preamble is not: `handle_post()` walks the part headers from the start of whichever segment it holds, so a cut inside the octet-stream part header (Firefox splits right after `filename="`) loses the parser, streaming never starts and the request times out.

A firmware image cannot be buffered whole, so this buffers the multipart body only until the octet-stream part header is complete, locates the payload, then streams from that point exactly as before. The header is small and reuses the configuration buffer, which is idle during a firmware upload, so no extra memory is needed.

Tested on a SWTGW218AS with a valid-CRC image (a reboot proves a complete parse):

| request shape | before | after |
|---|---|---|
| whole body | reboots | reboots |
| split inside the payload | reboots | reboots |
| split inside the part header (Firefox) | times out, no reset | reboots |
| 8 KB body across several segments | not applicable | reboots |

The 8 KB case is past the buffer, confirming only the header is buffered and the body still streams. Configuration save (#356) is unaffected: whole and Firefox-split uploads both return `200` with the stored config byte-identical.

### DIFF
--- a/httpd/httpd.c
+++ b/httpd/httpd.c
@@ -401,20 +401,27 @@ static uint16_t preamble_payload_start(uint16_t n)
 }
 
 
+// Source window for stream_upload(); filled by the caller before the call
+__xdata struct {
+	__xdata uint8_t *p;
+	uint16_t bptr;
+	uint16_t plen;
+} upload_settings;
+
 /*
  * Reads post data from the http stream and writes it into flash memory
- * Input: the current position in the TCP buffer (uip_appdata)
+ * Input: upload_settings, set by the caller
  * Returns 1: More data to read, 0: Upload complete, all parts reads
  */
-uint8_t stream_upload(__xdata uint8_t *p, uint16_t bptr, uint16_t plen)
+uint8_t stream_upload(void)
 {
 	__xdata struct httpd_state * __xdata s = &(uip_conn->appstate);
 
 	dbg_string("Stream_upload called: ");
-	dbg_short(bptr); dbg_char('\n');
+	dbg_short(upload_settings.bptr); dbg_char('\n');
 
 	do {
-		if (bptr >= plen) {
+		if (upload_settings.bptr >= upload_settings.plen) {
 			s->tstate = TSTATE_POST;
 			return 1;
 		}
@@ -451,15 +458,15 @@ uint8_t stream_upload(__xdata uint8_t *p, uint16_t bptr, uint16_t plen)
 			flash_region.addr = uptr;
 			flash_region.len = 1;
 			flash_write_bytes(flash_buf);
-			if (bptr >= plen)
+			if (upload_settings.bptr >= upload_settings.plen)
 				return 0;
 			return 1;
 		}
-		if (p[bptr] == boundary[bindex]) {
+		if (upload_settings.p[upload_settings.bptr] == boundary[bindex]) {
 			if (!bindex)
 				crc_final = crc_value;
-			crc16(p + bptr);
-			bptr++;
+			crc16(upload_settings.p + upload_settings.bptr);
+			upload_settings.bptr++;
 			bindex++;
 		} else {
 			if (bindex) {
@@ -467,8 +474,8 @@ uint8_t stream_upload(__xdata uint8_t *p, uint16_t bptr, uint16_t plen)
 				write_len += bindex;
 				bindex = 0;
 			}
-			crc16(p + bptr);
-			flash_buf[write_len++] = p[bptr++];
+			crc16(upload_settings.p + upload_settings.bptr);
+			flash_buf[write_len++] = upload_settings.p[upload_settings.bptr++];
 			if (write_len >= FLASH_PAGE_SIZE) {
 				dbg_string("len: "); dbg_short(write_len); dbg_char(' ');
 				dbg_string("CRC16: "); dbg_short(crc_value); dbg_char('\n');
@@ -644,7 +651,10 @@ void handle_post(void)
 		// clear any stale response so the completion check in the
 		// appcall POST branch cannot send leftovers
 		slen = 0;
-		stream_upload(config_buf, cfg_end, pre_acc);
+		upload_settings.p = config_buf;
+		upload_settings.bptr = cfg_end;
+		upload_settings.plen = pre_acc;
+		stream_upload();
 
 		dbg_string("Done reading first fragment\n");
 		return;
@@ -727,7 +737,10 @@ void httpd_appcall(void)
 	} else if (uip_newdata() && s->tstate == TSTATE_POST) {
 		// Check here maxupload by subtracting uip_len and close socekt if fails!
 		if (max_upload - uip_len > 0) {
-			stream_upload(uip_appdata, 0, uip_len);
+			upload_settings.p = uip_appdata;
+			upload_settings.bptr = 0;
+			upload_settings.plen = uip_len;
+			stream_upload();
 			// A completed part with a built verdict must go out
 			// through the normal TX path
 			if (s->tstate == TSTATE_NONE && slen)
--- a/httpd/httpd.c
+++ b/httpd/httpd.c
@@ -49,6 +49,7 @@ __xdata uint8_t boundary[72];
 __xdata uint8_t config_upload;
 __xdata uint8_t config_buf[CONFIG_UPLOAD_BUF];
 __xdata uint16_t cfg_pos, cfg_hdr, cfg_body, cfg_end, cfg_last;
+__xdata uint16_t pre_acc;
 __xdata uint8_t cfg_bl;
 __xdata uint8_t * __xdata content_type = 0;
 __xdata uint8_t * __xdata session = 0;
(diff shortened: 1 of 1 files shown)


janh:
This seems to work for me with Firefox!

bloqaudio:
#356 is merged, so the dependency note above is resolved: this branch is rebased onto current main and now carries only its own commit (60d779a).

Revalidated on a SWTGW218AS, each firmware case a real flash cycle with a distinct build stamp to prove the delivered image booted:

| case | result |
|---|---|
| upload cut inside `filename="` (the Firefox shape from #353) | flashes, device resets, new image boots |
| upload cut between disposition and content type | flashes, device resets, new image boots |
| unsplit upload | flashes, device resets |
| configuration matrix from #356, all shapes and the 4095/4096/4300 sizes | unchanged, all pass |

For reference, current main without this fix accepts a split upload byte by byte but never flashes and never answers; the client waits until it gives up.

bloqaudio:
Pushed a second commit carrying over the firmware upload verdict from #322, reworked for the buffered upload path: an explicit 200/400 with the CRC result and Content-Length, and the chip reset deferred until the response is fully ACKed. The config side of #322 is not needed anymore since the config rework already sends a real response. Re-tested on hardware: a corrupted image returns the 400 body with no reset, a good image returns the full 200 body and then reboots.

httpd/httpd.c
```diff
@@ -384,21 +378,38 @@ static uint8_t config_take(void)
...
+		return 0;
+	cfg_pos += 24;
+	while (cfg_pos + 3 < n && !strstart(&config_buf[cfg_pos], "\r\n\r\n"))
+		cfg_pos++;
+	if (cfg_pos + 3 >= n)
+		return 0;
+	return cfg_pos + 4;
+}
+
+
```
vDorst: My opinion is that `cfg_pos` and some other variable (`__xdata uint16_t cfg_pos, cfg_hdr, cfg_body, cfg_end, cfg_last;`) that are declared globally and don't hold a state, should be removed and locally declared within the function itself.
Let the compiler do the allocation, and only help the compiler when needed.

```c
static uint16_t preamble_payload_start(uint16_t n)
{
	uint16_t cfg_pos = 0;

	while(1) {
		if (cfg_pos + 24 > n)
			return 0;
		if (strstart(&config_buf[cfg_pos], "application/octet-stream"))
			break;
		cfg_pos++;
	}
	cfg_pos += 24;

	while (1) {
		if (cfg_pos + 4 > n)
			return 0;
		if (strstart(&config_buf[cfg_pos], "\r\n\r\n"))
			break;
		cfg_pos++;
	}
	return cfg_pos + 4;
}
```
Now the function is 60 bytes smaller.

httpd/httpd.c
```diff
@@ -49,6 +49,7 @@ __xdata uint8_t boundary[72];
 __xdata uint8_t config_upload;
 __xdata uint8_t config_buf[CONFIG_UPLOAD_BUF];
 __xdata uint16_t cfg_pos, cfg_hdr, cfg_body, cfg_end, cfg_last;
+__xdata uint16_t pre_acc;
```
vDorst, line 53: Maybe document what this is doing.

vDorst (commented):
Small comments.

httpd/httpd.c
bloqaudio: I tried this in three shapes on a SWTGW218AS build with SDCC 4.5.0. Plain locals for the whole cfg_* group do not link on this machine: direct data space is fully allocated, so the linker stops with "Could not get 6 consecutive bytes in internal RAM for area OSEG". Declaring them __xdata locals instead makes SDCC spill three config_take() temporaries into direct space (same link failure) and grows the code by about 120 bytes.

What does work is the leaf function: the scan cursor of preamble_payload_start() and its length parameter hold no state between calls, and as plain locals both end up in registers. That saves 64 bytes of BANK1 code and 2 bytes of xdata, with the internal RAM layout unchanged. Pushed that; the config_take() offsets stay in static xdata.

httpd/httpd.c
bloqaudio: Added a comment: it is the one variable here with cross-segment state, which is also why it stays global.

httpd/httpd.c
vDorst: I was playing a bit to move the arguments to a struct so the stats of the function is keep there. That seems to help.

Even without this patch it feels the code is getting messier of all the small patch on patch on patch. I hope you or maybe other people can clean it up a bit.

```patch
diff --git a/httpd/httpd.c b/httpd/httpd.c
index 9bd8292..692c038 100644
--- a/httpd/httpd.c
+++ b/httpd/httpd.c
@@ -407,20 +407,25 @@ static uint16_t preamble_payload_start(uint16_t n)
        return cfg_pos + 4;
 }
 
+__xdata struct {
+       __xdata uint8_t * p;
+       uint16_t bptr;
+       uint16_t plen;
+} upload_settings;
 
 /*
  * Reads post data from the http stream and writes it into flash memory
  * Input: the current position in the TCP buffer (uip_appdata)
  * Returns 1: More data to read, 0: Upload complete, all parts reads
  */
-uint8_t stream_upload(__xdata uint8_t *p, uint16_t bptr, uint16_t plen)
+uint8_t stream_upload(void)
 {
        dbg_string("Stream_upload called: ");
        dbg_short(bptr); dbg_char('\n');
 
-       static __xdata struct httpd_state * __xdata s = &(uip_conn->appstate);
+       __xdata struct httpd_state * __xdata s = &(uip_conn->appstate);
        do {
-               if (bptr >= plen) {
+               if (upload_settings.bptr >= upload_settings.plen) {
                        s->tstate = TSTATE_POST;
                        return 1;
                }
@@ -457,15 +462,15 @@ uint8_t stream_upload(__xdata uint8_t *p, uint16_t bptr, uint16_t plen)
                        flash_region.addr = uptr;
                        flash_region.len = 1;
                        flash_write_bytes(flash_buf);
-                       if (bptr >= plen)
+                       if (upload_settings.bptr >= upload_settings.plen)
                                return 0;
                        return 1;
                }
-               if (p[bptr] == boundary[bindex]) {
+               if (upload_settings.p[upload_settings.bptr] == boundary[bindex]) {
                        if (!bindex)
                                crc_final = crc_value;
-                       crc16(p + bptr);
-                       bptr++;
+                       crc16(upload_settings.p + upload_settings.bptr);
+                       upload_settings.bptr++;
                        bindex++;
                } else {
                        if (bindex) {
@@ -473,8 +478,8 @@ uint8_t stream_upload(__xdata uint8_t *p, uint16_t bptr, uint16_t plen)
                                write_len += bindex;
                                bindex = 0;
                        }
-                       crc16(p + bptr);
-                       flash_buf[write_len++] = p[bptr++];
+                       crc16(upload_settings.p + upload_settings.bptr);
+                       flash_buf[write_len++] = upload_settings.p[upload_settings.bptr++];
                        if (write_len >= FLASH_PAGE_SIZE) {
                                dbg_string("len: "); dbg_short(write_len); dbg_char(' ');
                                dbg_string("CRC16: "); dbg_short(crc_value); dbg_char('\n');
@@ -650,7 +655,12 @@ void handle_post(void)
                // clear any stale response so the completion check in the
                // appcall POST branch cannot send leftovers
                slen = 0;
-               stream_upload(config_buf, cfg_end, pre_acc);
+
+               upload_settings.p = config_buf;
+               upload_settings.bptr = cfg_end;
+               upload_settings.plen = pre_acc;
+
+               stream_upload();
 
                dbg_string("Done reading first fragment\n");
                return;
@@ -733,7 +743,12 @@ void httpd_appcall(void)
        } else if (uip_newdata() && s->tstate == TSTATE_POST) {
                // Check here maxupload by subtracting uip_len and close socekt if fails!
                if (max_upload - uip_len > 0) {
-                       stream_upload(uip_appdata, 0, uip_len);
+
+                       upload_settings.p = uip_appdata;
+                       upload_settings.bptr = 0;
+                       upload_settings.plen = uip_len;
+
+                       stream_upload();
                        // A completed part with a built verdict must go out
                        // through the normal TX path
                        if (s->tstate == TSTATE_NONE && slen)
```

bloqaudio: I rebased your struct onto the current branch head (your diff was against an older tree, and the dbg_short(bptr) at the top needed to follow the parameter into the struct) and measured three shapes with SDCC 4.5.0 on a SWTGW218AS build: the current three-parameter signature, your struct, and a variant restoring the original stream_upload(bptr) signature with the source window in two xdata globals.

Your struct is the only one that helps internal RAM: it frees 4 bytes of DSEG (stack grows from 128 to 132 bytes) for +94 bytes of BANK1 code and +6 bytes of xdata. The one-parameter variant surprisingly frees nothing in internal RAM and still costs +30 bytes of code, so I dropped it. Given DSEG is the segment with zero headroom, that trade looks right, so I pushed your version. Revalidated on hardware merged to current main: a good image flashes and reboots, a corrupted image returns the 400 body without resetting, and a Firefox upload (the natural split this PR exists for) flashes correctly.

On the cleanup point: agreed the upload path has accreted. I would rather keep this PR to the split-upload fix and take a dedicated cleanup pass as a follow-up, so the fix does not wait on a refactor discussion.

vDorst:
Thanks for fixing this!

> On the cleanup point: agreed the upload path has accreted. I would rather keep this PR to the split-upload fix and take a dedicated cleanup pass as a follow-up, so the fix does not wait on a refactor discussion.

I agree and I hope someone wants to pick this up.

PR #362, opened 2026-08-25
leds: move the SWTG018AS-V2.1.0 custom LED init to BANK2

### LINKED ISSUE #361: main does not link for MACHINE_PCB_SWTG018AS_V2_1_0 (bank 0 overflow)
Building current main for the SWTG018AS-V2.1.0 variant of the SWTGW218AS fails at the image step:

```
make MACHINE=PCB_SWTG018AS_V2_1_0
Error: Bank 0: code segment too large at 0x4000!
```

Bank 0 ends at 0x4066, 102 bytes over. The machine's custom LED register table and init loop from #317 sit in machine.c, which is common-bank code; main has grown past the point where they fit. machine_check compiles each machine but never links, so CI does not see it.

Fix in the linked PR: host the table and loop in rtl837x_leds.c, which is already a BANK2 file.

bloqaudio:
Fixes #361.

main currently fails to link for MACHINE_PCB_SWTG018AS_V2_1_0:

```
Error: Bank 0: code segment too large at 0x4000!
```

Bank 0 ends at 0x4066, 102 bytes past the limit. The machine's custom LED register table and its init loop live in machine.c, whose code and constants go to the common bank. The machine built and was validated when #317 was submitted; main has grown since, and the machine_check target compiles each machine but never links, so the overflow stayed invisible until an image build.

The first commit generalizes the fix as suggested in review: machine_custom_init() runs once at boot, so all the per-machine hooks move to a new machine_init.c compiled into BANK2, and the prototype becomes banked. No machine's init code or tables consume bank 0 any more, and machine_check compiles the new file per machine.

The second commit shrinks the table: measured against what the generic leds_setup() computes on this board, eight of the 21 stock values are identical to the generic result and eight more can revert to it with no change in LED behaviour at any tested link state. The five that remain each do something: 6528 selects blue over green at 10G, 6540 and 6548 carry the SFP and copper LED set behaviour, 65dc enables the LED outputs, and PIN_MUX_0 routes the blue-LED pin to the LED controller.

Bank 0 closes with 94 bytes free. Verified from a clean boot on a SWTG018AS-V2.1.0 board: copper green with activity blink at 2.5G, SFP green at 2.5G with a 2.5G module, SFP blue at 10G with a 10G module, LEDs off on link down.

A machine_check variant that also links each machine would catch this class of breakage in CI; happy to look at that as a follow-up if there is interest.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,6 +53,7 @@ create_build_dir:
 # Keep machine.c in first position to fail immediately on invalid $MACHINE value
 SRCS = \
 	machine.c \
+	machine_init.c \
 	cmd_editor.c \
 	cmd_parser.c \
 	dhcp.c \
@@ -142,6 +143,7 @@ machine_check:
 	do \
 	echo "Checking $${MACHINE}"; \
 	$(CC) $(CC_FLAGS) -DMACHINE_$${MACHINE} -MMD -o $(BUILDDIR)/tmp/machine_check -c machine.c; \
+	$(CC) $(CC_FLAGS) -DMACHINE_$${MACHINE} -MMD -o $(BUILDDIR)/tmp/machine_check -c machine_init.c; \
 	done
 	@rm -rf $(BUILDDIR)/tmp
 
--- a/machine.c
+++ b/machine.c
@@ -42,8 +42,6 @@ __code const struct machine machine = {
 		    },
 };
 
-void machine_custom_init(void) { }
-
 #elif defined MACHINE_KP_9000_6XH_X
 __code const struct machine machine = {
 	.machine_name = "keepLink KP-9000-6XH-X",
@@ -71,8 +69,6 @@ __code const struct machine machine = {
 			0 },
 		    },
 };
-
-void machine_custom_init(void) { }
 
 #elif defined MACHINE_KP_9000_6XH_X2
 __code const struct machine machine = {
@@ -122,10 +118,6 @@ __code const struct machine machine = {
 		},
 	};
 
-void machine_custom_init(void) {
-	reg_bit_set(RTL837X_REG_LED_GLB_IO_EN, 6);
-}
-
 #elif defined MACHINE_KP_9000_9XH_X_EU
 __code const struct machine machine = {
 	.machine_name = "keepLink KP-9000-9XH-X-EU",
@@ -150,8 +142,6 @@ __code const struct machine machine = {
 			LEDS_2G5 | LEDS_LINK | LEDS_ACT },
 		    },
 };
-
-void machine_custom_init(void) { }
 
 #elif defined MACHINE_KP_9000_9XHML_X_V2_2
 __code const struct machine machine = {
@@ -204,8 +194,6 @@ __code const struct machine machine = {
 	},
 };
 
-void machine_custom_init(void) { }
-
 #elif defined MACHINE_KP_9000_9XHML_X_V3_1
 __code const struct machine machine = {
 	.machine_name = "keepLink KP-9000-9XHML-X V3.1",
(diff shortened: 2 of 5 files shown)


logicog:
I would suspect that other devices may have similar issues with BANK 0 being full. Is there a way to generalize this fix for all devices? Probably device-init code that anyway runs only once on booting does not really need to be in BANK0, anyway.

bloqaudio:
Good suggestion, done that way now. The first commit moves all the per-machine hooks to a new machine_init.c compiled into BANK2, with the prototype banked, so no machine's one-shot init or its tables consume bank 0. machine_check compiles the new file per machine as well, so the hooks keep CI syntax coverage. A second commit trims the SWTG018AS-V2.1.0 table itself to the five registers with a measurable effect on the board.

vDorst:
> I would suspect that other devices may have similar issues with BANK 0 being full

I have seen it too while testing multipule PRs at once.

bloqaudio:
Heads up: main has grown since this PR was opened and the two commits alone no longer make the build fit. The #326 and #345 merges added roughly 113 bytes to bank 0, so a V2.1.0 build now ends 22 bytes past 0x4000 even with this PR applied. I have added a third commit moving the spanning tree module to BANK2: its entry points are already declared __banked and none of it runs in interrupt context, it just never had the placement pragma. That frees about 1.1 KB of bank 0 and restores headroom for all machines. A V2.1.0 build from current main including this PR is running on hardware here, and the stp off path was exercised on the running board to confirm the cross-bank call.

Makefile
```diff
@@ -142,6 +143,7 @@ machine_check:
 	do \
 	echo "Checking $${MACHINE}"; \
 	$(CC) $(CC_FLAGS) -DMACHINE_$${MACHINE} -MMD -o $(BUILDDIR)/tmp/machine_check -c machine.c; \
+	$(CC) $(CC_FLAGS) -DMACHINE_$${MACHINE} -MMD -o $(BUILDDIR)/tmp/machine_check -c machine_init.c; \
```
vDorst, line 146: Can this be just one line?
Seems to compile in my setup, but doesn't means that it works.

```patch
@@ -142,8 +142,7 @@ machine_check:
        for MACHINE in `grep -e ' MACHINE_' machine.c | sed -e 's%^.* MACHINE_%%' -e 's%[ ]*//.*$$%%' | sort -u`; \
        do \
        echo "Checking $${MACHINE}"; \
-       $(CC) $(CC_FLAGS) -DMACHINE_$${MACHINE} -MMD -o $(BUILDDIR)/tmp/machine_check -c machine.c; \
-       $(CC) $(CC_FLAGS) -DMACHINE_$${MACHINE} -MMD -o $(BUILDDIR)/tmp/machine_check -c machine_init.c; \
+       $(CC) $(CC_FLAGS) -DMACHINE_$${MACHINE} -MMD -o $(BUILDDIR)/tmp/machine_check -c machine.c -c machine_init.c; \
        done
        @rm -rf $(BUILDDIR)/tmp
```

bloqaudio: That form silently loses the second file: sdcc compiles only one source per invocation, so machine_init.c is dropped with "warning 120: cannot compile more than one source file" and exit code 0. Easy to check with a deliberately broken second file, the one-line form still passes while compiling that file alone fails. So the check needs both invocations.

elissa2333:
when I was run `make MACHINE=FG_4GT_2SX_V2_0` , the error output have like this report question.
```text
Input file size: 184934
Input file contains 3 banks
Error: Bank 0: code segment too large at 0x4000!
make: *** [Makefile:128: output/FG_4GT_2SX_V2_0/rtlplayground-v0.1.0-45d0aac-FG_4GT_2SX_V2_0.bin] Error 5
```

vDorst:
@bloqaudio Thanks for explanation and fixing this issue.

PR #365, opened 2026-08-27
Add PCB V2.1 to ZX-SWTGW215AS documentation and use MAC address on flash

janh:
ZX-SWTGW215AS also exists with PCB V2.1 instead of the already documented V2.0. It is slightly different, but doesn't require any changes to the machine definition.

This pull request mostly just extends the documentation to include the new PCB version.

In addition, it extends the machine definition to use the MAC address stored on flash at offset 0x1fc000.

@bennydiamond Since you originally added support for this device, maybe you can confirm if the MAC address also works for V2.0. Although it looks like the existing code for reading the MAC address should fall back gracefully anyway.


vDorst:
Thanks

bennydiamond:
> ZX-SWTGW215AS also exists with PCB V2.1 instead of the already documented V2.0. It is slightly different, but doesn't require any changes to the machine definition.
> 
> This pull request mostly just extends the documentation to include the new PCB version.
> 
> In addition, it extends the machine definition to use the MAC address stored on flash at offset 0x1fc000.
> 
> @bennydiamond Since you originally added support for this device, maybe you can confirm if the MAC address also works for V2.0. Although it looks like the existing code for reading the MAC address should fall back gracefully anyway.

I'm not quite sure what I sould be looking for but I updated with a build that now includes your change, my MAC stayed the same before and after the update.

janh:
@bennydiamond The expected result would be that the device now uses its factory-assigned MAC address instead of the auto-generated one.

On my device flash address 0x1fc000 contains a MAC address with OUI 1c:2a:a3 (Shenzhen HongRui Optical Technology Co., Ltd.), which was also used by the original firmware.

So, if nothing changed for you with the update, it means either the device never contained a MAC address at this offset, or it was overwritten at some point.

PR #366, opened 2026-08-27
Add factory MAC flash offset for keepLink KP-9000-9XHML-X V3.1 + mac command

gnacho:
## Problem

A keepLink KP-9000-9XHML-X V3.1 (RTL8373) comes up with a **locally administered MAC** (derived from the chip UUID) instead of the factory MAC printed on the label, because the `MACHINE_KP_9000_9XHML_X_V3_1` block in `machine.c` does not define `mac_flash_offset`. The boot code then skips the flash MAC read and takes the LAA fallback path.

There is also no way to override the MAC from the startup configuration, which matters for boards whose factory area is not at a known fixed location (or has been blanked).

## Fix

- `machine.c`: add `.mac_flash_offset = 0x1FC000` to the V3_1 block (same location other machines in the file use), so a valid factory MAC is picked up at boot.
- `cmd_parser.c`: add a `mac` console command:
  - `mac` - print the current management MAC.
  - `mac <aa:bb:cc:dd:ee:ff>` - validate (unicast, globally administered, non-zero OUI) and set the management MAC, re-registering the static L2 management entry so the change applies without a reboot. As a regular command it also works from the startup config (`execute_config` runs it before the final static L2 entry and the derived hostname are set up), so it persists across boots.
  - Parsing and printing live in `parse_mac()` / `print_mac()` helpers next to `parse_ip()` / `print_ip()` so future commands can reuse them.
- `rtlplayground.c` / `rtl837x_common.h`: add a small `memcmp()` helper (the project links no libc and already defines its own `memcpy()` / `memset()`).

The `html/config.js` change to keep the `mac` line when saving from the web UI is intentionally **not** in this PR: #315 rewrites that file and vDorst asked to rebase it on top once #315 is merged.

## Evidence

On the affected device the factory MAC was intact at `0x1FC000` (`78:d8:00:...`, verified with a flash dump), confirming only the offset was missing. With the offset defined the switch boots with the label MAC; the `mac` command gives the same result from the config without depending on the flash layout.

Built with sdcc for `KP_9000_9XHML_X_V3_1`. SRAM check per the project guide: the internal-RAM (DSEG) layout is **unchanged** against main; only XDATA grows by 6 bytes (the local `mac[6]` buffer) and ROM by ~715 bytes for the new code.


cmd_parser.c
```diff
@@ -1534,6 +1534,117 @@ void print_sw_version(void) __banked {
...
+static __xdata uint8_t fd_buf[16];
+static __xdata uint8_t fd_i;
+static __xdata uint8_t sm_mac[6];
+static __xdata uint8_t sm_i;
+static __xdata uint8_t sm_c;
+static __xdata uint8_t sm_nib;
+static __xdata uint8_t sm_pos;
+static __xdata uint8_t sm_blank;
+
+#define FACTORY_MAC_OFFSET 0x1FC000UL
```
feelfree69: Shouldn't this be set to the value of `.mac_flash_offset` in machine.c?
And probably refuse to use `setmac` if it is not defined for the specific machine.

cmd_parser.c
```diff
@@ -1534,6 +1534,117 @@ void print_sw_version(void) __banked {
...
+
+/* --- flashdump/setmac: factory MAC diagnostics and restore --- */
+static __xdata uint8_t fd_buf[16];
+static __xdata uint8_t fd_i;
+static __xdata uint8_t sm_mac[6];
+static __xdata uint8_t sm_i;
+static __xdata uint8_t sm_c;
+static __xdata uint8_t sm_nib;
+static __xdata uint8_t sm_pos;
+static __xdata uint8_t sm_blank;
```
vDorst: No need to all the variables on xdata and be static.
only use xdata and or static when needed. Here is a guide how I check that https://github.com/logicog/RTLPlayground/pull/368#discussion_r3886861122

So the general rule is, keep the variables local, put it in xdata when it consumes SRAM and add static when xdata is not enough.

Using xdata to pass larger data > `uint16_t` is fine.

cmd_parser.c
```diff
@@ -1534,6 +1534,117 @@ void print_sw_version(void) __banked {
...
+			sm_c -= 'a' - 10;
+		else
+			break;
+		if (sm_nib == 0xFF) {
+			sm_nib = sm_c;
+		} else {
+			sm_mac[sm_i++] = (sm_nib << 4) | sm_c;
+			sm_nib = 0xFF;
+		}
+	}
```
vDorst: Make this a seperate function like `parse_ip()`.
So we can reuse it in the future.

cmd_parser.c
```diff
@@ -1534,6 +1534,117 @@ void print_sw_version(void) __banked {
...
+	if ((sm_mac[0] & 0x03) || ((sm_mac[0] | sm_mac[1] | sm_mac[2]) == 0)) {
+		print_string("refusing: must be unicast, globally administered\n");
+		return;
+	}
+	print_string("setting MAC ");
+	for (fd_i = 0; fd_i < 6; fd_i++) {
+		print_byte(sm_mac[fd_i]);
+		if (fd_i != 5)
+			write_char(':');
+	}
```
vDorst: Make this a `print_mac()` like `print_ip()`.
Also replace some other places that prints a mac-address.

vDorst:
I think writing back the mac-adress to the flash is not the right way.
Because not all device has a fix location.
So I would like to see just a `mac <address>` command to write the value in config.
The issue is, when to apply the new value.
Via CLI, applying directly is fine, but via the web, it will loose connection because mac-address has changed.

gnacho:
@vDorst you are right, writing back to a fixed flash location is not portable. I reworked the PR to follow your suggestion:

- `setmac` and `flashdump` are gone.
- Added a `mac <aa:bb:cc:dd:ee:ff>` command that validates the address (unicast, globally administered, non-zero OUI), sets the management MAC and re-registers the static L2 entry so it applies live. Since it is a regular command, a line in the startup config persists it across boots (`execute_config` runs before the final static L2 entry and the derived hostname), and the web UI now keeps the `mac` line when saving the config.
- Kept only `.mac_flash_offset = 0x1FC000` on the V3_1 machine block, so boards that do have a valid factory MAC at the usual location pick it up at boot without any config line. Same pattern the SWTGW218AS / SWTG018AS-V2.1.0 entries already use.

The branch is rebased on current main and rebuilt (sdcc, KP_9000_9XHML_X_V3_1, only the pre-existing warnings). Thanks for the review.

cmd_parser.c
```diff
@@ -364,6 +369,93 @@ void print_ip(__xdata uint8_t * ptr)
...
+	uint8_t c;
+
+	if (cmd_words_len == 1) {
+		print_string("Current MAC: ");
+		print_byte(uip_ethaddr.addr[0]); write_char(':');
+		print_byte(uip_ethaddr.addr[1]); write_char(':');
+		print_byte(uip_ethaddr.addr[2]); write_char(':');
+		print_byte(uip_ethaddr.addr[3]); write_char(':');
+		print_byte(uip_ethaddr.addr[4]); write_char(':');
+		print_byte(uip_ethaddr.addr[5]); write_char('\n');
```
vDorst: You didn't read all my review points.
Add and use this `print_mac()` to print mac-address.
```c
// Prints an mac address.
void print_mac(__xdata uint8_t *ptr)
{
	uint8_t idx = 0;
	uint8_t num;
	while(1) {
		num = *ptr++;
		print_byte(num);
		if (++idx == 6)
			break;
		write_char(':');
	}
}
```

cmd_parser.c
```diff
@@ -364,6 +369,93 @@ void print_ip(__xdata uint8_t * ptr)
...
+			c -= 'a' - 10;
+		else
+			break;
+		if (hi == 0xff) {
+			hi = c;
+		} else {
+			mac_scratch[b++] = (hi << 4) | c;
+			hi = 0xff;
+		}
+	}
```
vDorst, line 426: make a parse_mac().

cmd_parser.c
```diff
@@ -364,6 +369,93 @@ void print_ip(__xdata uint8_t * ptr)
...
+	uip_ethaddr.addr[4] = mac_scratch[4]; uip_ethaddr.addr[5] = mac_scratch[5];
+	port_l2_static_mgmt(uip_ethaddr.addr, management_vlan, false);
+
+	print_string("MAC set to: ");
+	print_byte(uip_ethaddr.addr[0]); write_char(':');
+	print_byte(uip_ethaddr.addr[1]); write_char(':');
+	print_byte(uip_ethaddr.addr[2]); write_char(':');
+	print_byte(uip_ethaddr.addr[3]); write_char(':');
+	print_byte(uip_ethaddr.addr[4]); write_char(':');
+	print_byte(uip_ethaddr.addr[5]); write_char('\n');
```
vDorst: Also use the `print_mac()`

cmd_parser.c
```diff
@@ -364,6 +369,93 @@ void print_ip(__xdata uint8_t * ptr)
...
+	}
+	if ((mac_scratch[0] & 0x03) || ((mac_scratch[0] | mac_scratch[1] | mac_scratch[2]) == 0)) {
+		print_string("refusing: must be unicast, globally administered\n");
+		return;
+	}
+
+	// no-op if unchanged
+	if (mac_scratch[0] == uip_ethaddr.addr[0] && mac_scratch[1] == uip_ethaddr.addr[1] &&
+	    mac_scratch[2] == uip_ethaddr.addr[2] && mac_scratch[3] == uip_ethaddr.addr[3] &&
+	    mac_scratch[4] == uip_ethaddr.addr[4] && mac_scratch[5] == uip_ethaddr.addr[5]) {
```
vDorst: `memcmp()` seems useful here.

cmd_parser.c
```diff
@@ -364,6 +369,93 @@ void print_ip(__xdata uint8_t * ptr)
...
+		print_string("MAC unchanged\n");
+		return;
+	}
+
+	// swap the static management entry over to the new address
+	mac_old[0] = uip_ethaddr.addr[0]; mac_old[1] = uip_ethaddr.addr[1];
+	mac_old[2] = uip_ethaddr.addr[2]; mac_old[3] = uip_ethaddr.addr[3];
+	mac_old[4] = uip_ethaddr.addr[4]; mac_old[5] = uip_ethaddr.addr[5];
+	port_l2_static_mgmt(mac_old, management_vlan, true);
```
vDorst: `mac_old` can be remove, use `uip_ethaddr.addr` directly.
I see no benefits to copy it and then use it.

cmd_parser.c
```diff
@@ -364,6 +369,93 @@ void print_ip(__xdata uint8_t * ptr)
...
+
+	uip_ethaddr.addr[0] = mac_scratch[0]; uip_ethaddr.addr[1] = mac_scratch[1];
+	uip_ethaddr.addr[2] = mac_scratch[2]; uip_ethaddr.addr[3] = mac_scratch[3];
+	uip_ethaddr.addr[4] = mac_scratch[4]; uip_ethaddr.addr[5] = mac_scratch[5];
```
vDorst: memcpy() seems right here.

html/config.js
```diff
@@ -71,6 +72,7 @@ const conf_overwrite = [
   /^mtu\s+\d{1,2}\b/,
   /^bw\s+(in|out)\s+\d{1,2}\b/,
   /^hostname\b/,
+  /^mac\b/,
 ];
 
 function parseConf(s){
```
vDorst, line 76: This has to rebase ontop of #315 which is almost ready to merge.
Just wait until #315 is merge.

cmd_parser.c
```diff
@@ -67,6 +67,11 @@ __xdata	char save_cmd;
 
 __xdata uint8_t ip[4];
 
+/* Scratch for parse_mac(), in xdata: a local here would take internal RAM
+ * the linker has none of. */
```
vDorst: such comments are not needed. adding __xdata is fine to move the variable from SRAM to XDATA.
Only if the compiler need extra help, I would like to see a comment with a link to #386. Like done here https://github.com/logicog/RTLPlayground/pull/374/changes/daf9d8965a97b0acec330c6955db4edc42f2a628#diff-b6e192fdbc314f152dbba6d0f8526feb58d597abdf808cd5badd4956f82b61a5R326
So we can investigate it later.

cmd_parser.c
```diff
@@ -67,6 +67,11 @@ __xdata	char save_cmd;
...
+__xdata uint8_t mac_scratch[6];       /* parsed MAC while validating */
```
vDorst: This can just be a local variable.

vDorst (changes requested):
See the comments.

gnacho:
@vDorst thanks for the careful review. I reworked the PR addressing all the points:

- **Variables**: the global `__xdata` scratch is gone. `parse_mac_cmd()` now uses a local `__xdata uint8_t mac[6]` (only XDATA, +6 B), and `mac_old` is removed entirely: the old static L2 entry is removed while the address is still in `uip_ethaddr`, then the new one is written with the project's own `memcpy()` and re-added.
- **`parse_mac()`**: extracted as a reusable helper next to `parse_ip()` (`uint8_t parse_mac(uint8_t idx, __xdata uint8_t *mac)`, returns bytes consumed or 0). The `mac` command is now `parse_mac_cmd()` so the helper name stays free for future use.
- **`print_mac()`**: added next to `print_ip()` with your exact implementation, and used for both prints in the command.
- **`memcmp()`**: added a small `memcmp()` helper next to the project's own `memcpy()`/`memset()` in `rtlplayground.c` (declared in `rtl837x_common.h`). The project links no libc (it defines its own string helpers), so I could not pull the sdcc one in; a project-local helper matches the existing pattern. The "unchanged" check now uses it.
- **`config.js`**: removed from this PR. Per your note I will rebase the `mac` line on top of #315 once it is merged.

SRAM check following your guide (diff of `rtlplayground.mem` against main): internal RAM / DSEG layout is unchanged; only XDATA grows by 6 bytes (the local `mac[6]` buffer) and ROM by ~715 bytes. No spare-SRAM regression.

cmd_parser.c
```diff
@@ -364,6 +364,98 @@ void print_ip(__xdata uint8_t * ptr)
 	}
 }
 
+// Prints a MAC address.
+void print_mac(__xdata uint8_t *ptr)
```
vDorst: This needs the __banked keyword.
`void print_mac(__xdata uint8_t *ptr) __banked`
Also `void print_ip(__xdata uint8_t * ptr) __banked`

cmd_parser.c
```diff
@@ -364,6 +364,98 @@ void print_ip(__xdata uint8_t * ptr)
...
+// consumed, or 0 on error.
+uint8_t parse_mac(uint8_t idx, __xdata uint8_t *mac)
+{
+	uint8_t b = 0;
+	uint8_t hi = 0xff;   /* 0xff = waiting for the high nibble */
+	uint8_t c;
+
+	while (b < 6) {
+		c = cmd_buffer[idx];
+		if (c == NUL || c == ' ' || c == '\r' || c == '\n')
```
vDorst: `c == '\r' || c == '\n'` can be removed, these are already filtered out.

vDorst (changes requested):
The parser accepts also values like
- `mac 78:AA:00:::::::::::::::::::::::::::32:37::e3` 
  MAC set to: 78:aa:00:32:37:e3
  Save the startup config to keep it across reboots
- `mac 78:AA:00:32:37:e3:2aajfhalsdhfsadjfhlkfjhlkfjhkdshfkdsf`
  MAC unchanged

cmd_parser.c
gnacho: Done. print_mac() and print_ip() are now declared and defined with __banked, matching cmd_parser() which lives in codeseg BANK2 and is itself __banked. The build links cleanly for KP_9000_9XHML_X_V3_1.

cmd_parser.c
gnacho: Removed. The check for c == \r or c == \n was dropped from parse_mac; cmd_editor already strips carriage returns and newlines before the buffer is parsed, so only NUL and space are treated as end-of-token.

gnacho:
@vDorst following up on the review with the requested changes:

- **Strict parse_mac()**: the parser now only accepts a well-formed aa:bb:cc:dd:ee:ff address (or "-" separators). Each octet must be exactly two hex digits and the 5 separators must be the same single character, so the malformed inputs you flagged are rejected instead of normalized: `78:AA:00:::::::::::32:37::e3` errors, and `78:AA:00:32:37:e3:2aajf...` errors. Nothing is accepted after the 6th octet. Validation lives in a small reusable `hexval()` helper.
- **__banked**: `print_mac()` and `print_ip()` are now `__banked`, matching `cmd_parser()` which is in `codeseg BANK2`.
- **html/config.js**: no rebase needed. With #315 merged, `config.js` is gone (the assets are bundled into `html/main.js`), so there is no `mac` line to carry there. The command remains available as a startup-config line (`execute_config` runs before the final static L2 entry); the web UI path is a separate concern.
- **uip-neighbor.c**: left as-is; that MAC print is inside a commented-out block, so reusing print_mac there is not applicable.

Verified: rebuild clean with make MACHINE=KP_9000_9XHML_X_V3_1, internal-RAM (DSEG) layout unchanged against main, XSEG unchanged, ROM +184 bytes. A host-side test confirms the two review vectors are rejected. Thanks.

vDorst:
@gnacho First, can we stop with the AI-slop responses!

I don't like to codesize is +100 bytes more and end results.

This what I made.
Made `mac_parse_result` global, like the other parsers.
Made the code +20 smaller than the previous solution.

```patch
index 4a8cfa6..1cdd45a 100644
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -66,6 +66,7 @@ __xdata uint8_t cmd_available;
 __xdata	char save_cmd;
 
 __xdata uint8_t ip[4];
+__xdata uint8_t mac_parse_result[6];
 
 /* Scratch for parse_syslog(), in xdata: a local here would take internal RAM
  * the linker has none of. */
@@ -383,36 +384,47 @@ void print_mac(__xdata uint8_t *ptr) __banked
 // Parses a MAC address (12 hex digits, ':' or '-' separators optional) from
 // cmd_buffer starting at idx into mac[6]. Returns the number of bytes
 // consumed, or 0 on error.
-uint8_t parse_mac(uint8_t idx, __xdata uint8_t *mac)
+uint8_t parse_mac(uint8_t idx)
 {
 	uint8_t b = 0;
-	uint8_t hi = 0xff;   /* 0xff = waiting for the high nibble */
-	uint8_t c;
+	uint8_t c = ':';
 
-	while (b < 6) {
-		c = cmd_buffer[idx];
-		if (c == NUL || c == ' ' || c == '\r' || c == '\n')
-			break;
-		idx++;
-		if (c == ':' || c == '-')
-			continue;
-		c |= 0x20;
-		if (c >= '0' && c <= '9')
+	do {
+		if (c != ':' && c != '-')
+			goto err;
+
+		__bit again = true;
+		uint8_t v = 0;
+
+		while(1) {
+			c = cmd_buffer[idx++];
 			c -= '0';
-		else if (c >= 'a' && c <= 'f')
-			c -= 'a' - 10;
-		else
-			break;
-		if (hi == 0xff) {
-			hi = c;
-		} else {
-			mac[b++] = (hi << 4) | c;
-			hi = 0xff;
+			if (c > 9) {
+				c |= 0x20;
+				c += '0' - 'a';
+				if (c > 5)
+					goto err;
+				c += 10;
+			}
+			// force swap-instr.
+			v = (v << 4) | (v >> 4);
+			v |= c;
+
+			if (again)
+				again = false;
+			else
+				break;
 		}
-	}
-	if (b != 6 || hi != 0xff)
-		return 0;
-	return idx;
+		mac_parse_result[b++] = v;
+
+		c = cmd_buffer[idx++];
+	} while(b < 6);
+
+	if (c == ' ' || c == NUL)
+		return idx;
+
+err:
+	return 0;
 }
 
 // Sets the management MAC from a "mac <aa:bb:cc:dd:ee:ff>" command (or prints
@@ -423,23 +435,21 @@ uint8_t parse_mac(uint8_t idx, __xdata uint8_t *mac)
 // on every boot (execute_config runs before the final static L2 entry).
 void parse_mac_cmd(void)
 {
-	__xdata uint8_t mac[6];
-
 	if (cmd_words_len == 1) {
 		print_string("Current MAC: ");
 		print_mac(uip_ethaddr.addr);
 		write_char('\n');
 		return;
 	}
-	if (cmd_words_len != 2 || !parse_mac(cmd_words_b[1], mac)) {
+	if (cmd_words_len != 2 || !parse_mac(cmd_words_b[1])) {
 		print_string("Error: mac [<aa:bb:cc:dd:ee:ff>]\n");
 		return;
 	}
-	if ((mac[0] & 0x03) || ((mac[0] | mac[1] | mac[2]) == 0)) {
+	if ((mac_parse_result[0] & 0x03) || ((mac_parse_result[0] | mac_parse_result[1] | mac_parse_result[2]) == 0)) {
 		print_string("refusing: must be unicast, globally administered\n");
 		return;
 	}
-	if (memcmp(mac, uip_ethaddr.addr, 6) == 0) {
+	if (memcmp(mac_parse_result, uip_ethaddr.addr, 6) == 0) {
 		print_string("MAC unchanged\n");
 		return;
 	}
@@ -447,7 +457,7 @@ void parse_mac_cmd(void)
 	// Swap the static management entry over to the new address: remove the
 	// old one (still in uip_ethaddr) before overwriting it.
 	port_l2_static_mgmt(uip_ethaddr.addr, management_vlan, true);
-	memcpy(uip_ethaddr.addr, mac, 6);
+	memcpy(uip_ethaddr.addr, mac_parse_result, 6);
 	port_l2_static_mgmt(uip_ethaddr.addr, management_vlan, false);
 
 	print_string("MAC set to: ");
diff --git a/cmd_parser.h b/cmd_parser.h
index d8ad883..7564b45 100644
--- a/cmd_parser.h
+++ b/cmd_parser.h
@@ -15,7 +15,6 @@ void execute_config(void) __banked;
 void execute_commands(__xdata uint8_t *p) __banked;
 void print_ip(__xdata uint8_t *ptr) __banked;
 void print_mac(__xdata uint8_t *ptr) __banked;
-uint8_t parse_mac(uint8_t idx, __xdata uint8_t *mac);
 void print_sw_version(void) __banked;
 void clear_command_history(void) __banked;
```

gnacho:
Adopted your version: mac_parse_result is a global like the other parsers, parse_mac() takes only the index, and the local buffer and hexval() are gone.

Build for KP_9000_9XHML_X_V3_1: ROM 110302 -> 110193 bytes, XSEG and internal RAM unchanged. The two malformed inputs from the review are still rejected.

Pushed as 12fb6fd.

vDorst:
Thanks

PR #368, opened 2026-08-28
Disable all `__sfr32` SDCC bug 4070 and fix all compiler warnings.

vDorst:
- Disable all `__sfr32` variables and added workaround. Because of SDCC bug 4070.
- Many times we need help the compiler to reason about variable type or define.

This fixing all these warning
```
rtlplayground.c:1819: warning 110: conditional flow changed by optimizer: so said EVELYN the modified DOG
rtlplayground.c:1826: warning 110: conditional flow changed by optimizer: so said EVELYN the modified DOG
rtl837x_igmp.c:95: warning 110: conditional flow changed by optimizer: so said EVELYN the modified DOG
rtl837x_igmp.c:96: warning 110: conditional flow changed by optimizer: so said EVELYN the modified DOG
rtl837x_phy.c:473: warning 110: conditional flow changed by optimizer: so said EVELYN the modified DOG
rtl837x_phy.c:499: warning 126: unreachable code
rtl837x_phy.c:586: warning 238: inline function 'rtl8224_read_reg_u16' might lose function attributes
rtl837x_phy.c:593: warning 158: overflow in implicit constant conversion
rtl837x_phy.c:605: warning 238: inline function 'rtl8224_write_reg_u16' might lose function attributes
rtl837x_port.c:832: warning 85: in function port_ingress_vlan_filter_set unreferenced function argument : 'enabled'
httpd/httpd.c:611: warning 110: conditional flow changed by optimizer: so said EVELYN the modified DOG
httpd/httpd.c:732: warning 110: conditional flow changed by optimizer: so said EVELYN the modified DOG
uip/uip.c:377: warning 110: conditional flow changed by optimizer: so said EVELYN the modified DOG
uip/uip.c:380: warning 110: conditional flow changed by optimizer: so said EVELYN the modified DOG
uip/uip.c:1464: warning 110: conditional flow changed by optimizer: so said EVELYN the modified DOG
```

### DIFF
--- a/rtl837x_phy.h
+++ b/rtl837x_phy.h
@@ -28,7 +28,7 @@ void phy_show(uint8_t port) __banked;
 void phy_reset(uint8_t port) __banked;
 void rtl8224_read_reg_u16(uint16_t reg) __banked;
 void rtl8224_write_reg_u16(uint16_t reg, uint16_t val) __banked;
-void rtl8224_sds_write(uint16_t sds_cmd, uint16_t val) __banked;
+void rtl8224_sds_write(uint16_t sds_cmd, __xdata uint16_t val) __banked;
 void phy_config_8261(uint8_t phy, uint8_t sds) __banked;
 
 #define	RTL8224_SDS_WRITE(sds_id, page, reg, v) uint16_t _sdscmd = (uint16_t)(sds_id & 0x01) | (1 << 14) | (1 << 15); \
--- a/rtlplayground.c
+++ b/rtlplayground.c
@@ -1813,6 +1813,12 @@ void init_smi(void)
 	/* Set the SMI(i.e.I2C) type for PHY polling, 0b01 is 2.5/10G PHY. Disable (0b00) for the SFP-ports
 	 * which are at port 8 and additionally at port 3 for a dual SFP device
 	 */
+
+	// using 16bit value, because it can load cheap.
+	// Default: 0x00005555
+	// Workaround for SDCC BUG 4070: SFR_DATA_U32 = 0x00005555;
+	SFR_DATA_U16_UPPER = 0x0000;
+	SFR_DATA_U16= 05555;
 	if (machine.n_10g == 2) {
 		REG_SET(RTL837X_REG_SMI_MAC_TYPE, 0x00015555);
 	} else {
@@ -1820,11 +1826,20 @@ void init_smi(void)
 	}
 
 	// Configure polling of all PHYs by the MAC to detect link-state changes
-	if (machine_detected.isRTL8373) {
-		REG_SET(RTL837X_REG_SMI_PORT_POLLING, 0xff);
-	} else {
-		REG_SET(RTL837X_REG_SMI_PORT_POLLING, machine.n_sfp == 2 ? 0xf0 : 0x1f8);
+	// Default: 0x000000ff
+	// Workaround for SDCC BUG 4070: SFR_DATA_U32 = 0x000000ff;
+	SFR_DATA_U16_UPPER = 0x0000;
+	SFR_DATA_U16= 0x00ff;
+	if (!machine_detected.isRTL8373) {
+		if (machine.n_sfp == 2) {
+			SFR_DATA_0 = 0xf0;
+		} else {
+			SFR_DATA_0 = 0xf8;
+			SFR_DATA_16 = 0x1f;
+		}
 	}
+	reg_write(RTL837X_REG_SMI_PORT_POLLING);
+
 	// Enable MDC
 	reg_read_m(RTL837X_REG_SMI_CTRL);
 	sfr_mask_data(1, 0, 0x70); 	// Set bits 12-14 to enable MDC for SMI0-SMI2
--- a/rtlplayground.c
+++ b/rtlplayground.c
@@ -1814,16 +1814,17 @@ void init_smi(void)
 	 * which are at port 8 and additionally at port 3 for a dual SFP device
 	 */
 
-	// using 16bit value, because it can load cheap.
 	// Default: 0x00005555
 	// Workaround for SDCC BUG 4070: SFR_DATA_U32 = 0x00005555;
 	SFR_DATA_U16_UPPER = 0x0000;
-	SFR_DATA_U16= 05555;
+	SFR_DATA_U16 = 0x5555;
 	if (machine.n_10g == 2) {
-		REG_SET(RTL837X_REG_SMI_MAC_TYPE, 0x00015555);
-	} else {
-		REG_SET(RTL837X_REG_SMI_MAC_TYPE, machine.n_sfp == 2 ? 0x00005515 : 0x00005555);
-	}
+		// 0x00015555, only change the bytes that differs from the default.
+		SFR_DATA_16 = 0x01;
+	} else if (machine.n_sfp == 2)
+		// 0x00005515
+		SFR_DATA_0 = 0x15;
+	reg_write(RTL837X_REG_SMI_MAC_TYPE);
 
 	// Configure polling of all PHYs by the MAC to detect link-state changes
 	// Default: 0x000000ff
@@ -1839,7 +1840,6 @@ void init_smi(void)
 		}
 	}
 	reg_write(RTL837X_REG_SMI_PORT_POLLING);
-
 	// Enable MDC
 	reg_read_m(RTL837X_REG_SMI_CTRL);
 	sfr_mask_data(1, 0, 0x70); 	// Set bits 12-14 to enable MDC for SMI0-SMI2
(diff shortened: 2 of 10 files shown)


rtlplayground.c
```diff
@@ -1813,18 +1813,26 @@ void init_smi(void)
...
+	// using 16bit value, because it can load cheap.
+	// Default: 0x00005555
+	uint16_t mask = 0x0055;
 	if (machine.n_10g == 2) {
-		REG_SET(RTL837X_REG_SMI_MAC_TYPE, 0x00015555);
-	} else {
-		REG_SET(RTL837X_REG_SMI_MAC_TYPE, machine.n_sfp == 2 ? 0x00005515 : 0x00005555);
-	}
+		// 0x00015555
+		mask = 0155;
```
logicog: Is this correct? Should this no be 0x0155?

rtlplayground.c
```diff
@@ -1813,18 +1813,26 @@ void init_smi(void)
...
+	} else if (machine.n_sfp == 2)
+		// 0x00005515
+		mask = 0015;
```
logicog: Same here.

logicog (commented):
Amazing, that you found out how to get rid of the dog-warnings. I was wondering where this came from and tried getting rid of them for quite some time...
I compile-tested and there are no more warnings at all. I also went through the code and it looks great. I just have some tiny observations...

rtlplayground.c
vDorst: No, but I refactor this part.

rtlplayground.c
vDorst: Same here, No, but I refactor this part.

vDorst:
Refactor that part and also rebase.
I saw I messed up, because I locally merge other PR in this branch.
I think I restored that now.

Please review it again!

rtl837x_phy.h
```diff
@@ -28,7 +28,7 @@ void phy_show(uint8_t port) __banked;
 void phy_reset(uint8_t port) __banked;
 void rtl8224_read_reg_u16(uint16_t reg) __banked;
 void rtl8224_write_reg_u16(uint16_t reg, uint16_t val) __banked;
-void rtl8224_sds_write(uint16_t sds_cmd, uint16_t val) __banked;
+void rtl8224_sds_write(uint16_t sds_cmd, __xdata uint16_t val) __banked;
 void phy_config_8261(uint8_t phy, uint8_t sds) __banked;
```
feelfree69, line 32: I still have not understood when __xdata should be used and when not.
Is there a general advice, eg. one/first argument always without __xdata, 2nd and following arguments with __xdata?

rtlplayground.c
```diff
@@ -1813,18 +1813,30 @@ void init_smi(void)
...
 	 */
+
+	// Default: 0x00005555
+	SFR_DATA_U32 = 0x00005555;
 	if (machine.n_10g == 2) {
-		REG_SET(RTL837X_REG_SMI_MAC_TYPE, 0x00015555);
-	} else {
-		REG_SET(RTL837X_REG_SMI_MAC_TYPE, machine.n_sfp == 2 ? 0x00005515 : 0x00005555);
-	}
+		// 0x00015555, only change the bytes that differs from the default.
```
feelfree69, line 1822: According to the comment, in line 1821 byte 2 shall be changed. And in line 1823, byte 4 shall be changed. But both writes goes to the same SFR_DATA_16.
I suppose either the comment is wrong or the code.

vDorst: code is wrong, well spotted.

rtl837x_phy.h
logicog: I don't think there is a general rule. In the given case I assume that the 2 16 bit variables in the call-signature are not put into register variables when the call takes place. Instead, sdcc puts them into DSEG, which is vasteing RAM. The rule seems to be that you can place 3 bytes into registers when a call takes place.
Use xdata in calls when the parameters would otherwise end up in DSEG.

feelfree69 (approved):

rtl837x_phy.h
vDorst: You know it when you need it, when SRAM used goes up.

How to check:

What I do it, start with main-branch and compile it normally, so we have a baseline.
- Copy `rtlplayground.mem` to a other location so it can be use in a later step
  `copy output/<your_device>/rtlplayground.mem ..`.
- Checkout/merge or edit your code.
- Compile it and then diff both mem-files to see if the SRAM has changed.
  `diff ../rtlplayground.mem output/<your_device>/rtlplayground.mem`
- When nothing has changes, you are fine.

Otherwise:

If I change back this code above, I get this diff
```patch
4,5c4,5
< 0x10:|a|a|a|a|a|a|a|a|a|b|b|b|b|c|c|j|
< 0x20:|B|B|T|d|d|d|d|d|d|d|d|d|d|d|d|d|
---
> 0x10:|a|a|a|a|a|a|a|a|a|b|b|b|b|j| | |
> 0x20:|B|B|T|c|c|c|c|d|d|d|d|d|d|d|d|d|
7,10c7,10
< 0x40:|e|e|e|e|e|f|f|f|f|f|f|f|f|f|f|f|
< 0x50:|f|f|f|f|g|g|g|g|g|g|g|g|g|g|h|h|
< 0x60:|h|h|h|h|h|h|h|i|i|i|i|i|i|i|i|i|
< 0x70:|i|i|i|Q|Q|Q|Q|Q|Q|Q|Q|S|S|S|S|S|
---
> 0x40:|d|d|d|d|e|e|e|e|e|f|f|f|f|f|f|f|
> 0x50:|f|f|f|f|f|f|f|f|g|g|g|g|g|g|g|g|
> 0x60:|g|g|h|h|h|h|h|h|h|h|h|i|i|i|i|i|
> 0x70:|i|i|i|i|i|i|i|Q|Q|Q|Q|Q|Q|Q|Q|S|
21,22c21,22
< 16 bit mode initial stack starts at: 0x7b (sp set to 0x7a) with 133 bytes available.
< No spare internal RAM space left.
---
> 16 bit mode initial stack starts at: 0x7f (sp set to 0x7e) with 129 bytes available.
> The largest spare internal RAM space starts at 0x1e with 2 bytes available.
28,29c28,29
<    EXTERNAL RAM     0x0001   0x36f2   14066    16777216
<    ROM/EPROM/FLASH  0x0000   0x2d6cd  96086    16777216
---
>    EXTERNAL RAM     0x0001   0x36f0   14064    16777216
>    ROM/EPROM/FLASH  0x0000   0x2d6af  95889    16777216
```

Now I know something is wrong.
Because I know the source file I can look into generated `rtl837x_phy.asm` and for section `internal ram data`.

```
; internal ram data                                                                                                                    
        .area DSEG    (DATA)                                                                                                           
_phy_config_8261_PARM_2:                                                                                                               
        .ds 1                                                                                                                          
_phy_show_port_10000_116:                                                                                                              
        .ds 1                                                                                                                          
_rtl8224_sds_write_PARM_2:                                                                                                             
        .ds 2                                                                                                                          
;-------------------
```

You can see `_rtl8224_sds_write_PARM_2` is placed in SRAM. Now you know which parameter to place in __xdata or maybe rewrite the code to eliminate it.

I made a Rust program that makes a report of all the asm-files.
So I can diff before and after.

Look like this, notice the green text on the bottom.

Screenshot_20260829_164138

This is how I hunted down SRAM usage and the patch series for it.
Some are easy, but some need refactor large part of the code to make it better.

So I hope this helps.

vDorst:
I think this patch breaks the `l2` command both in CLI and WWW.

```
<new link: 2400004000, was 2450004000>
l2

>
tcp: found no unused connections.
```
It sees data but no value.

Update: It is patch #370.

vDorst:
Rebase because of conflict.

vDorst:
Making a fix to disable all `__sfr32` access, because all are corrupted.

vDorst:
I added `Disable all __sfr32 variables and added workaround. Because of SDCC bug 4070.`.
Maybe not ideal to add this workaround, but my fixes also made use of this variable so I added it here.

rtlplayground.c
```diff
@@ -1813,18 +1813,33 @@ void init_smi(void)
...
+	// Default: 0x000000ff
+	// Workaround for SDCC BUG 4070: SFR_DATA_U32 = 0x000000ff;
+	SFR_DATA_U16_UPPER = 0x0000;
+	SFR_DATA_U16= 0x00ff;
+	if (!machine_detected.isRTL8373) {
+		if (machine.n_sfp == 2) {
+			SFR_DATA_0 = 0xf0;
+		} else {
+			SFR_DATA_0 = 0xf8;
+			SFR_DATA_16 = 0x1f;
```
feelfree69: Original code sets 0x1f8; modified code probably sets 0x1f00f8 (?)

vDorst: Yes it was wrong.
Fixed.

PR #374, opened 2026-08-31
httpd: clean up the upload path in handle_post()

bloqaudio:
Follow-up to #359, as offered in the review discussion there: a dedicated cleanup pass over the upload path, with no behavior change intended.

Three commits:

1. **httpd: scope the multipart parse cursors to their functions.** `cfg_pos`, `cfg_hdr`, `cfg_body`, `cfg_end`, `cfg_last` and `cfg_bl` are only used by `config_take()`, so they move from file scope into the function. The five cursors stay `static` because SDCC register-caches automatic xdata locals and spills them through DSEG slots, which would cost 6 bytes of direct RAM; `cfg_bl` is a plain local, and the mechanism is written up in #386. `handle_post()` reused `cfg_pos` and `cfg_end` as scratch for unrelated values; those get named locals instead.
2. **httpd: track buffered upload bytes in one accumulator.** The config path counted buffered bytes in `write_len`, the variable the firmware path uses for flash staging. Both paths now accumulate in `pre_acc`.
3. **httpd: split the upload fragment handlers out of handle_post().** The tail of `handle_post()` interleaved the config and firmware paths behind one conditional. Each path gets its own function and `handle_post()` reduces to routing. Both handlers keep their own copy of the buffer-and-bounds-check prologue: a shared helper costs 2 bytes of direct RAM for the fragment pointer in every shape I tried. Amended after review: the handler parameters are no longer `__xdata`-homed, which the review measurement showed saves 4 bytes of xdata and 12 of code with an identical RAM map.

**Memory:** the internal RAM layout is byte for byte identical to `main` before and after the series, checked on `PCB_SWTG018AS_V2_1_0` (stack 156 bytes free) and `SWTGW218AS` (157 bytes free). xdata grows by 7 bytes and code by 71 bytes. `make machine_check` passes for all machines.

**Tested** on a SWTG018AS V2.1.0 board, the full matrix re-run on the final head (005209f): config upload round trip in six TCP segmentations with exact readback; 4095 bytes accepted, 4096 and 4300 rejected with `400` (sector bound and buffer bound), each in single-segment, split-header and split-body shapes; the original configuration restored and surviving a reboot; a corrupt firmware image rejected with "NO: checksum failed, not applied" and no reset (reachability-monitored); and a valid firmware upload through Firefox (which naturally splits the stream) answering "OK: checksum verified, rebooting" and coming back with the configuration intact.

### DIFF
--- a/httpd/httpd.c
+++ b/httpd/httpd.c
@@ -48,7 +48,8 @@ __xdata uint8_t boundary[72];
 #define CONFIG_UPLOAD_BUF (CONFIG_LEN + 384)
 __xdata uint8_t config_upload;
 __xdata uint8_t config_buf[CONFIG_UPLOAD_BUF];
-// part-header bytes buffered so far; accumulates across TCP segments
+// bytes buffered in config_buf so far (config body, or a firmware part
+// header); accumulates across TCP segments
 __xdata uint16_t pre_acc;
 __xdata uint8_t * __xdata content_type = 0;
 __xdata uint8_t * __xdata session = 0;
@@ -332,7 +333,7 @@ static uint8_t config_take(void)
 	// the body is complete once the closing boundary has arrived
 	cfg_last = 0;
 	while (1) {
-		if (cfg_last + cfg_bl + 1 >= write_len)
+		if (cfg_last + cfg_bl + 1 >= pre_acc)
 			return 0;
 		if (strstart_x(&config_buf[cfg_last], boundary)
 		    && strstart(&config_buf[cfg_last + cfg_bl], "--"))
@@ -548,7 +549,7 @@ void handle_post(void)
 			dbg_string("Configuration upload\n");
 			verify_crc = 0;
 			config_upload = 1;
-			write_len = 0;
+			pre_acc = 0;
 		}
 		// Check for other POST requests, which are not multipart, below
 	} else {
@@ -604,15 +605,15 @@ void handle_post(void)
 		}
 		if (config_upload) {
 			frag_len = uip_len - (p - uip_appdata);
-			if (write_len + frag_len >= CONFIG_UPLOAD_BUF) {
+			if (pre_acc + frag_len >= CONFIG_UPLOAD_BUF) {
 				print_string("Configuration too large, aborting.\n");
 				config_upload = 0;
 				s->tstate = TSTATE_NONE;
 				send_bad_request();
 				return;
 			}
-			memcpy(config_buf + write_len, p, frag_len);
-			write_len += frag_len;
+			memcpy(config_buf + pre_acc, p, frag_len);
+			pre_acc += frag_len;
 			uint8_t taken = config_take();
 
 			if (!taken) {
--- a/httpd/httpd.c
+++ b/httpd/httpd.c
@@ -503,12 +503,83 @@ uint8_t stream_upload(void)
 }
 
 
+static void handle_config_fragment(__xdata uint8_t *p)
+{
+	__xdata struct httpd_state * __xdata s = &(uip_conn->appstate);
+	__xdata uint16_t frag_len;
+	uint8_t taken;
+
+	frag_len = uip_len - (p - uip_appdata);
+	if (pre_acc + frag_len >= CONFIG_UPLOAD_BUF) {
+		print_string("Configuration too large, aborting.\n");
+		config_upload = 0;
+		s->tstate = TSTATE_NONE;
+		send_bad_request();
+		return;
+	}
+	memcpy(config_buf + pre_acc, p, frag_len);
+	pre_acc += frag_len;
+	taken = config_take();
+	if (!taken) {
+		s->tstate = TSTATE_MULTIPART;
+		return;
+	}
+	config_upload = 0;
+	s->tstate = TSTATE_NONE;
+	if (taken == 2) {
+		send_bad_request();
+		return;
+	}
+	slen = strtox(outbuf, "HTTP/1.1 200 OK\r\nConnection: close\r\n\r\n");
+}
+
+
+static void handle_firmware_fragment(__xdata uint8_t *p)
+{
+	__xdata struct httpd_state * __xdata s = &(uip_conn->appstate);
+	__xdata uint16_t frag_len, payload_start;
+
+	frag_len = uip_len - (p - uip_appdata);
+	if (pre_acc + frag_len >= CONFIG_UPLOAD_BUF) {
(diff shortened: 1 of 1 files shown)


httpd/httpd.c
```diff
@@ -500,6 +504,78 @@ uint8_t stream_upload(void)
 }
 
 
+static void handle_config_fragment(__xdata uint8_t * __xdata p)
```
vDorst: Can be just `static void handle_config_fragment(__xdata uint8_t *p)`.

httpd/httpd.c
```diff
@@ -500,6 +504,78 @@ uint8_t stream_upload(void)
...
+	s->tstate = TSTATE_NONE;
+	if (taken == 2) {
+		send_bad_request();
+		return;
+	}
+	slen = strtox(outbuf, "HTTP/1.1 200 OK\r\nConnection: close\r\n\r\n");
+}
+
+
+static void handle_firmware_fragment(__xdata uint8_t * __xdata p)
```
vDorst: Can be just `+static void handle_config_fragment(__xdata uint8_t *p)`

httpd/httpd.c
```diff
@@ -325,12 +324,17 @@ void gen_random_hex_chars(__xdata uint8_t * b, __xdata uint8_t bytes)
 /* 0: body incomplete, 1: configuration stored, 2: malformed */
 static uint8_t config_take(void)
 {
+	// static: automatic xdata locals get register-cached and spill
+	// through DSEG slots, which direct RAM has no room for
+	static __xdata uint16_t cfg_pos, cfg_hdr, cfg_body, cfg_end, cfg_last;
+	static __xdata uint8_t cfg_bl;
```
vDorst: A yes, in this case without `static` it want to consume 6 SRAM bytes.
I wonder why only 6, while this is around 10. Also why it still wants to put on SRAM any-ways because we marked as XDATA.
I have see this also happen in `early_boot_handle_button()`.
Good to know that `static` helps to mitigate it.

`static __xdata uint8_t cfg_bl;` can be `__xdata uint8_t cfg_bl;`.

vDorst (commented):
Some comments.

DrDoof:
Read all three. Measured against your base, not today's main, which is why the numbers agree: direct RAM identical byte for byte on a SWTGW218AS, xdata +12, BANK1 +61, bank 0 untouched. First time round I compared against main and got +10 and +55. My mistake, not yours.

The accumulator swap holds up, all four init sites still land where they should. The split reads line for line against what it replaced. And the cursors really are only in `config_take()` now, I went looking.

Small thing: the firmware error path picked up a `config_upload = 0` the original didn't have. Harmless, that path zeroes it on entry anyway, but you said no behaviour change so I thought I'd mention it.

@vDorst on the `static`: SDCC register-caches automatic xdata locals and spills them through DSEG, and DSEG is full at 0x80 here. I've walked into it three times this week.

httpd/httpd.c
bloqaudio: Measured it and you are right: with the handlers as their own functions, the `__xdata` homing of the parameter buys nothing. Dropping it on both handlers leaves the internal RAM map byte for byte identical and saves 4 bytes of xdata and 12 of code (SDCC 4.5, full builds). The 2 byte cost I had measured was for the shared-prologue helper shape and does not transfer here. Amended into the third commit; the branch is updated and the numbers in the description are refreshed.

httpd/httpd.c
bloqaudio: Same change applied here, covered by the measurement in the thread above.

httpd/httpd.c
bloqaudio: Numbers from full builds of each shape (SWTGW218AS, SDCC 4.5), relative to this branch:

- all six cursors as plain `__xdata` automatics: the stack start moves `0x63` to `0x69`, so exactly 6 bytes of direct RAM, plus 173 bytes of code
- only `cfg_bl` plain: direct RAM map identical, xdata 1 byte smaller, but 54 bytes more code

Why 6 and not ~11: `__xdata` places the variable's home in xdata, but SDCC still register-caches automatics, and what lands in direct RAM are the spill homes (`sloc`s) for values that must survive a register-clobbering point, such as the `strstart_x()` calls inside the loops. Those slots are shared and overlaid, so the count follows the register pressure in the function rather than the number of variables; `sloc` references in the generated asm for `httpd.c` go from 197 to 271 with all six plain. A `static` is never register-allocated, so it produces no spill homes at all, at the price of the extra loads and stores (that is the +54 and +173 of code). Same mechanism you saw in `early_boot_handle_button()`.

Given that `cfg_bl` alone is free only at today's register pressure, and one plain local in the block invites the next edit to de-static the rest, I kept all six `static` under the one comment. Happy to flip `cfg_bl` if you prefer it; the direct RAM map stays identical either way.

bloqaudio:
Branch updated: the two-line comment above the `config_take()` statics is removed, per the review guidance on #375 about xdata and DSEG comments. The rationale stays in the commit message. Comment-only change, codegen identical.

httpd/httpd.c
```diff
@@ -325,12 +324,15 @@ void gen_random_hex_chars(__xdata uint8_t * b, __xdata uint8_t bytes)
 /* 0: body incomplete, 1: configuration stored, 2: malformed */
 static uint8_t config_take(void)
 {
+	static __xdata uint16_t cfg_pos, cfg_hdr, cfg_body, cfg_end, cfg_last;
+	static __xdata uint8_t cfg_bl;
```
vDorst: In this case I could like to see a comment about that SRAM blowup.
So we can investigate it later.
Something like: `#<ISSUE_NUMBER>: Needs static otherwise it still lands in SRAM/DSEG`.
I would like a fix identifier/text/issuenumber that can be easily found back later.

`static __xdata uint8_t cfg_bl;` In my test this static was not needed.

bloqaudio: Both done. #386 records the spill mechanism with the measurements, and the line above the statics carries its number. `cfg_bl` is a plain `__xdata` local now: on the new head the internal RAM map is unchanged against `main` on both machines, xdata is one byte smaller and code 40 bytes larger. Re-tested on the board with the same matrix as the last head: config round trip in six segmentations, the 4095/4096/4300 bounds single and split, a corrupt image rejected without a reset, and a valid image through Firefox with the configuration intact. Branch and the numbers in the description are updated.

vDorst: Thanks for making an detailed issue.
So we can investigate it later.

vDorst:
Thanks!

PR #375, opened 2026-08-31
httpd: buffer a plain POST body that arrives after the headers

bloqaudio:
Found while reviewing #329: a `POST /cmd` whose body arrives in a separate TCP segment from the headers executes nothing, because `handle_post()` reads the body from the segment that carried the request line. The endpoint still answers as if it succeeded, so every scripted command becomes a silent no-op. Python's urllib and requests both write headers and body separately, which makes this the default experience for exactly the clients most likely to script `/cmd`. `/login` has the same hole. The multipart endpoints were fixed in #356 and #359; this covers the plain ones.

Five commits, the first three standalone fixes the review turned up:

1. **httpd: match header field names case-insensitively.** New `header_value()`: lower-case name anchored at the line start, request bytes folded as compared, value returned past any blanks. `content-type:` and `Content-Type:multipart/...` were both silently treated as absent before.
2. **httpd: saturate `parse_short()` instead of wrapping.** `?vid=65540` read as VLAN 4.
3. **httpd: keep `scan_header()` on the terminator of a truncated header.** The scan stepped over the NUL, so the end-of-header check read a stale byte.
4. **httpd: parse `Content-Length` in `scan_header()`**, through `parse_short()`.
5. **httpd: buffer a plain POST body that arrives after the headers.** When the first segment holds fewer body bytes than announced, the bytes accumulate in `config_buf` (idle outside multipart uploads) and the connection waits in `TSTATE_POSTBODY`; once complete, the endpoint runs against the buffer. The `/cmd` and `/login` executions move into `run_cmd_body()` and `run_login_body()`, which share `send_ok()` with the config path, and `run_login_body()` checks the `pwd=` prefix instead of skipping four bytes blindly. The announced length is the contract on both paths: a body that arrives with the headers is terminated at `Content-Length`, so pipelined bytes are no longer executed as commands; no usable `Content-Length` answers `411`; a length at or above the buffer answers `400`. The wait state aborts the connection after five seconds without progress, measured on the system tick, so headers followed by silence (including on the unauthenticated `/login`) cannot hold the single uIP slot.

**Memory:** internal RAM is byte for byte identical to `main` on `PCB_SWTG018AS_V2_1_0` and `SWTGW218AS`; xdata grows by 5 bytes and code by 783. `make machine_check` passes for all machines.

**Tested** on a SWTG018AS V2.1.0 board with a scripted client that controls the segmentation (25 cases plus the two hold cases, re-run in full on the current head):

- split login accepted and refused correctly, and a login body without the `pwd=` prefix refused
- `/cmd` split after the headers, and with the body itself split three ways
- `Content-Length` with no space, lower-case, and with a tab
- trailing bytes past the announced length ignored in both shapes
- no, zero and unparseable `Content-Length` answer `411`
- overlong commands split either way answer `400`
- an announcement beyond the buffer and an unauthenticated `/cmd` are refused at once without waiting for a body
- a truncated header answers `404`
- headers-then-silence on `/login` and `/cmd` are dropped at 5.0 s and the slot serves the next request
- urllib login and `/cmd` get correct verdicts
- single segment requests unchanged

The same script run against the firmware before this PR reproduces the original no-op and the reviewer findings (junk executed, overlong split answered `200`). A config upload round trip (single segment, headers then body, cuts inside the opening boundary, the config text and the closing boundary, and three cuts at once) stores an identical config every time, and the firmware upload that installed this build went through `/upload` as before.

Still independent of #329; whichever merges first, the other picks up a small textual conflict in the `/cmd` branch that resolves mechanically.

### DIFF
--- a/httpd/httpd.c
+++ b/httpd/httpd.c
@@ -62,6 +62,12 @@ __xdata uint8_t verify_crc;
 __xdata uint32_t max_upload;
 __xdata uint16_t short_parsed;
 
+#define POSTBODY_CMD	1
+#define POSTBODY_LOGIN	2
+#define POSTBODY_TIMEOUT (5 * SYS_TICK_HZ)
+__xdata uint8_t postbody_endpoint;
+__xdata uint16_t postbody_start;
+
 __xdata char passwd[21];
 // Set when a verified firmware upload awaits its response ACK, after
 // which the chip resets to apply the staged image
@@ -78,6 +84,7 @@ __xdata uint32_t last_session_use;
 #define TSTATE_CLOSED 		3
 #define TSTATE_POST 		4
 #define TSTATE_MULTIPART	5
+#define TSTATE_POSTBODY		6
 
 extern __xdata uint16_t crc_value;
 __xdata uint16_t crc_final;
@@ -265,6 +272,18 @@ void send_to_login(void)
 void send_unauthorized(void)
 {
 	slen = strtox(outbuf, "HTTP/1.1 401 Unauthorized\r\nConnection: close\r\n\r\n");
+}
+
+
+void send_length_required(void)
+{
+	slen = strtox(outbuf, "HTTP/1.1 411 Length Required\r\nConnection: close\r\n\r\n");
+}
+
+
+void send_ok(void)
+{
+	slen = strtox(outbuf, "HTTP/1.1 200 OK\r\nConnection: close\r\n\r\n");
 }
 
 
@@ -559,7 +578,7 @@ static void handle_config_fragment(__xdata uint8_t *p)
 		send_bad_request();
 		return;
 	}
-	slen = strtox(outbuf, "HTTP/1.1 200 OK\r\nConnection: close\r\n\r\n");
+	send_ok();
 }
 
 
@@ -604,11 +623,94 @@ static void handle_firmware_fragment(__xdata uint8_t *p)
 }
 
 
+static void run_cmd_body(__xdata uint8_t *body)
+{
+	execute_commands(body);
+	if (err_status != ERR_OK) {
+		send_bad_request();
+		return;
+	}
+	send_ok();
+}
+
+
+static void run_login_body(__xdata uint8_t *body)
+{
+	if (strstart(body, "pwd=") && is_url_word_x(body + 4, passwd)) {
+		dbg_string("Password accepted!\n");
+		read_reg_timer(&last_session_use);
+		gen_random_hex_chars(session_id, SESSION_ID_LENGTH);
+		session_id[SESSION_ID_LENGTH] = NUL;
+		slen = strtox(outbuf, "HTTP/1.1 302 Found\r\nConnection: close\r\nLocation: index.html\r\n" \
+				      "Set-Cookie: session=");
+		for (uint8_t i = 0; i < SESSION_ID_LENGTH; i++)
+			outbuf[slen++] = session_id[i];
+		slen += strtox(outbuf + slen, "; SameSite=Strict\r\n\r\n");
+	} else {
+		dbg_string("Password invalid!\n");
+		slen = strtox(outbuf, "HTTP/1.1 302 Found\r\nConnection: close\r\nLocation: login.html\r\n\r\n");
+	}
+}
+
+
+static uint8_t post_body_take(__xdata uint8_t *p)
+{
+	uint16_t have;
+
+	if (!content_length) {
+		send_length_required();
+		return 0;
+	}
+	if (content_length >= CONFIG_UPLOAD_BUF) {
+		send_bad_request();
+		return 0;
+	}
+	have = uip_len - (p - uip_appdata);
+	if (have >= content_length) {
+		p[content_length] = NUL;
+		return 1;
+	}
+	memcpy(config_buf, p, have);
+	pre_acc = have;
+	postbody_start = ticks;
+	uip_conn->appstate.tstate = TSTATE_POSTBODY;
+	return 0;
+}
+
+
+static void post_body_continue(void)
+{
+	uint16_t take;
+
+	// no header scan runs while the body is pending: content_length is this request's
+	take = content_length - pre_acc;
+	if (take > uip_len)
+		take = uip_len;
+	memcpy(config_buf + pre_acc, uip_appdata, take);
+	pre_acc += take;
+	if (pre_acc < content_length) {
+		postbody_start = ticks;
+		return;
+	}
+	config_buf[pre_acc] = NUL;
+	uip_conn->appstate.tstate = TSTATE_NONE;
+	if (postbody_endpoint == POSTBODY_CMD)
+		run_cmd_body(config_buf);
+	else
+		run_login_body(config_buf);
+}
+
+
 void handle_post(void)
 {
 	__xdata struct httpd_state * __xdata s = &(uip_conn->appstate);
 	__xdata uint8_t *p = uip_appdata;
 	__xdata uint8_t *request_path = p + 6;
+
+	if (s->tstate == TSTATE_POSTBODY) {
+		post_body_continue();
+		return;
+	}
 
 	// Was the multipart header sent in multiple packets?
 	if (s->tstate != TSTATE_MULTIPART) {
@@ -662,11 +764,11 @@ void handle_post(void)
 			send_unauthorized();
 			return;
 		}
-		execute_commands(p);
-		if (err_status != ERR_OK) {
-			send_bad_request();
+		postbody_endpoint = POSTBODY_CMD;
+		if (!post_body_take(p))
 			return;
-		}
+		run_cmd_body(p);
+		return;
 	} else if (is_word(request_path, "login")) {
 		dbg_string("POST login\n");
 
@@ -676,21 +778,11 @@ void handle_post(void)
 			return;
 		}
 
-		p += 8; // Read also over "pwd="
-		if (is_url_word_x(p, passwd)) {
-			dbg_string("Password accepted!\n");
-			read_reg_timer(&last_session_use);
-			gen_random_hex_chars(session_id, SESSION_ID_LENGTH);
-			session_id[SESSION_ID_LENGTH] = NUL;
-			slen = strtox(outbuf, "HTTP/1.1 302 Found\r\nConnection: close\r\nLocation: index.html\r\n" \
-					      "Set-Cookie: session=");
-			for (uint8_t i = 0; i < SESSION_ID_LENGTH; i++)
-				outbuf[slen++] = session_id[i];
-			slen += strtox(outbuf + slen, "; SameSite=Strict\r\n\r\n");
-		} else {
-			dbg_string("Password invalid!\n");
-			slen = strtox(outbuf, "HTTP/1.1 302 Found\r\nConnection: close\r\nLocation: login.html\r\n\r\n");
-		}
+		p += 4;
+		postbody_endpoint = POSTBODY_LOGIN;
+		if (!post_body_take(p))
+			return;
+		run_login_body(p);
 		return;
 	} else if (s->tstate == TSTATE_MULTIPART || is_word(request_path, "upload") || is_word(request_path, "config")) {
 		dbg_string("POST upload/config request\n");
@@ -712,8 +804,6 @@ void handle_post(void)
 		send_not_found();
 		return;
 	}
-	slen = strtox(outbuf, "HTTP/1.1 200 OK\r\nConnection: close\r\n\r\n");
-	return;
 }
 
 
@@ -743,6 +833,11 @@ void httpd_appcall(void)
 		if (s->tstate == TSTATE_ACKED) {
 			dbg_string("Closing because everything has been transmitted\n");
 			uip_close();
+			s->tstate = TSTATE_CLOSED;
+		} else if (s->tstate == TSTATE_POSTBODY
+			   && (uint16_t)ticks - postbody_start > POSTBODY_TIMEOUT) {
+			dbg_string("Body never arrived\n");
+			uip_abort();
 			s->tstate = TSTATE_CLOSED;
 		}
 	} else if (uip_acked() && s->tstate == TSTATE_TX) {
@@ -811,10 +906,12 @@ void httpd_appcall(void)
 		dbg_char('\n');
 #endif
 		p = uip_appdata;
-		if (is_word(p, "POST") || s->tstate == TSTATE_MULTIPART) {
+		if (is_word(p, "POST") || s->tstate == TSTATE_MULTIPART
+		    || s->tstate == TSTATE_POSTBODY) {
 			handle_post();
 			// If this is an ongoing post stream, then wait for the next packet
-			if (s->tstate == TSTATE_POST || s->tstate == TSTATE_MULTIPART) {
+			if (s->tstate == TSTATE_POST || s->tstate == TSTATE_MULTIPART
+			    || s->tstate == TSTATE_POSTBODY) {
 				uip_len = 0;
 				return;
 			}
--- a/httpd/httpd.c
+++ b/httpd/httpd.c
@@ -231,7 +231,10 @@ uint8_t parse_short(__xdata uint8_t *p)
 		c = *p++ - '0';
 		if (c > 9) { break; }
 		err = 0;
-		short_parsed = (short_parsed * 10) + c;
+		if (short_parsed > 6552)
+			short_parsed = 0xffff;
+		else
+			short_parsed = (short_parsed * 10) + c;
 	}
 	return err;
 }
--- a/httpd/httpd.c
+++ b/httpd/httpd.c
@@ -54,6 +54,7 @@ __xdata uint8_t config_buf[CONFIG_UPLOAD_BUF];
 __xdata uint16_t pre_acc;
 __xdata uint8_t * __xdata content_type = 0;
 __xdata uint8_t * __xdata session = 0;
+__xdata uint16_t content_length;
 
 // Global variables holding POST state
 __xdata uint16_t bindex; // Current index into the boundary
@@ -272,6 +273,7 @@ __xdata uint8_t *scan_header(__xdata uint8_t * __xdata p)
 	__xdata uint8_t *v;
 
 	content_type = 0;
+	content_length = 0;
 	session = 0;
 	authenticated = 0;
 
@@ -282,7 +284,10 @@ __xdata uint8_t *scan_header(__xdata uint8_t * __xdata p)
 		p++;
 		if ((v = header_value(p, "\ncontent-type:")))
 			content_type = v;
-		else if ((v = header_value(p, "\ncookie:"))) {
+		else if ((v = header_value(p, "\ncontent-length:"))) {
+			parse_short(v);
+			content_length = short_parsed;
+		} else if ((v = header_value(p, "\ncookie:"))) {
 			/* Scan for the "session" key: the header may hold several
 			 * cookies in any order. Match "session" not "session=" -
 			 * is_word() requires a separator after the match and '=' is
--- a/httpd/httpd.c
+++ b/httpd/httpd.c
@@ -138,6 +138,24 @@ bool is_word(__xdata uint8_t *xdata_str_p, __code uint8_t * __xdata code_str_p)
 }
 
 
+/* name must be lower-case, starting with the '\n' of the previous line's end */
+__xdata uint8_t *header_value(__xdata uint8_t *p, __code uint8_t *name)
+{
+	uint8_t u, c;
+
+	while ((c = *name++)) {
+		u = *p++;
+		if (u >= 'A' && u <= 'Z')
+			u += 'a' - 'A';
+		if (u != c)
+			return 0;
+	}
+	while (*p == ' ' || *p == '\t')
+		p++;
+	return p;
+}
+
+
 bool is_url_word_x(__xdata uint8_t *uri_str_p, __xdata uint8_t *src_str_p)
 {
 	uint8_t u, s;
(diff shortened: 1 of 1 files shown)


httpd/httpd.c
```diff
@@ -258,7 +269,19 @@ __xdata uint8_t *scan_header(__xdata uint8_t * __xdata p)
 			break;
 		if (is_word(p, "\nContent-Type:"))
 			content_type = p + 15;
-		else if (is_word(p, "\nCookie:")) {
+		else if (is_word(p, "\nContent-Length:")) {
+			__xdata uint8_t *c = p + 16;	/* past "\nContent-Length:" */
+			while (*c == ' ')
+				c++;
```
vDorst: Can we make a new function that parses the headers and replaces all the `is_word()` on other places that are interacting with the headerdata?
This new function does
- search for the wanted "headername" but case-insensitive.
  I think it is still fine that is matches `\n` on the front.
  Maybe add in the comment that `\n` is part of the search string and the search string must allready be lower-case to simplify the code. Can we make a macro for that, that adds '\n' and makes the string lowercase?
- Already strip the whitespaces.
- returns the length that was consumed. I think a `uint8_t` is sufficient.
  So we don't have to manual add the "headername" length.

httpd/httpd.c
```diff
@@ -258,7 +269,19 @@ __xdata uint8_t *scan_header(__xdata uint8_t * __xdata p)
...
+			while (*c >= '0' && *c <= '9') {
+				// saturate: a wrapped length would pass the size checks
+				if (content_length > 6552)
+					content_length = 0xffff;
+				else
+					content_length = content_length * 10 + (*c - '0');
+				c++;
+			}
```
vDorst: Maybe make it a seperate funcion. This we this with `cmd_parser()`.
So it can be reuse later.

vDorst (commented):
Some quick comments

httpd/httpd.c
```diff
@@ -500,12 +523,99 @@ uint8_t stream_upload(void)
...
+		return 1;
+	if (content_length >= CONFIG_UPLOAD_BUF) {
+		send_bad_request();
+		return 0;
+	}
+	memcpy(config_buf, p, have);
+	pre_acc = have;
+	postbody_len = content_length;
+	s = &(uip_conn->appstate);
+	s->tstate = TSTATE_POSTBODY;
```
DrDoof: This parks the connection with no response sent, and `uip-conf.h` has

```c
#define UIP_CONF_MAX_CONNECTIONS 1
```

so all of web management depends on this one slot coming back.

`POST /login` is the problem: it is the only endpoint that reaches here without `authenticated`. `/cmd` checks auth before calling `post_body_take()`, and `/upload` and `/config` only reach `TSTATE_MULTIPART` past their own auth check. So these headers alone, followed by silence, are enough:

```
POST /login HTTP/1.1
Content-Type: application/x-www-form-urlencoded
Content-Length: 100

```

`have` is 0, `0 >= 100` is false, `100 >= CONFIG_UPLOAD_BUF` is false, so we land on this line and `httpd_appcall()` does `uip_len = 0; return;`. Nothing goes out and the connection stays ESTABLISHED with no outstanding data.

That state never ages out. In `uip.c` the timer is only advanced for

```c
if(uip_connr->tcpstateflags == UIP_TIME_WAIT ||
   uip_connr->tcpstateflags == UIP_FIN_WAIT_2) {
```

and the only other route to `UIP_CLOSED` is behind `uip_outstanding(uip_connr)` reaching `UIP_MAXRTX`. ESTABLISHED with nothing in flight matches neither -- and `httpd_appcall()` handles `uip_aborted()` but not `uip_timedout()` in any case.

Before this patch the same request got its 302 and went TSTATE_TX -> ACK -> TSTATE_ACKED, where the `uip_poll()` branch closes the socket. Now the slot is held until reboot, and not only by someone doing it on purpose: an interrupted client -- Ctrl-C in a script, a laptop going to sleep, a link dropping mid-POST -- wedges it the same way, where it used to complete the request, wrongly but completely.

The cheap fix fits in the `uip_poll()` branch that is already there: `handle_tx()` calls `uip_periodic()` once per `idle()`, so a tick counter reset on entry to `TSTATE_POSTBODY` plus a `uip_close()` past a threshold costs two bytes of xdata.

httpd/httpd.c
```diff
@@ -258,7 +269,19 @@ __xdata uint8_t *scan_header(__xdata uint8_t * __xdata p)
 			break;
 		if (is_word(p, "\nContent-Type:"))
 			content_type = p + 15;
-		else if (is_word(p, "\nCookie:")) {
+		else if (is_word(p, "\nContent-Length:")) {
```
DrDoof: Adding to @vDorst's point: matching case-insensitively is not only tidier here, it changes behaviour.

`is_word()` requires a separator after the pattern:

```c
if (c == NUL) {
        if (u != NUL && u != ' ' && u != '\t' && u != ':' && u != '?' && u != '=' && u != '\n' && u != '\r')
                return false;
        return true;
}
```

The pattern ends in `:`, so `Content-Length:42` -- no space, which RFC 7230 permits since the OWS is optional -- fails on the `4`, and `content-length: 42` fails on the first byte. The `while (*c == ' ')` loop just below never gets the chance to do its job.

The failure is silent and lands right back on the bug this PR fixes: `content_length` stays 0, `have >= content_length` is then always true, the old path runs against an empty body and the client is told 200 OK for nothing.

Might be worth settling at the same time whether `POST /cmd` with no usable Content-Length and an empty body should answer 411 or 400 rather than 200. As it stands there is still no positive witness that a command actually arrived.

httpd/httpd.c
```diff
@@ -258,7 +269,19 @@ __xdata uint8_t *scan_header(__xdata uint8_t * __xdata p)
...
+			__xdata uint8_t *c = p + 16;	/* past "\nContent-Length:" */
+			while (*c == ' ')
+				c++;
+			while (*c >= '0' && *c <= '9') {
+				// saturate: a wrapped length would pass the size checks
+				if (content_length > 6552)
+					content_length = 0xffff;
+				else
+					content_length = content_length * 10 + (*c - '0');
```
DrDoof: A concrete target for this: `parse_short()` further up in this file already runs the same loop, and it serves `/vlan.json`, `/l2.json` and `/l2_del.json`:

```c
short_parsed = 0;
while(1) {
        c = *p++ - '0';
        if (c > 9) { break; }
        err = 0;
        short_parsed = (short_parsed * 10) + c;
}
```

It wraps, so `?vid=65540` reads as 4 today. Moving the saturation from here into `parse_short()` fixes both places at once and keeps a single copy of the loop, which seems worth it on this budget.

The threshold itself is right, for what it is worth: 6552 * 10 + 9 = 65529 fits, 6553 would not, and once set the 0xffff is sticky.

DrDoof (commented):
Read through the whole patch against `main`. The diagnosis is right and the core of the fix is sound. Things I checked that hold up:

- the offset arithmetic for both endpoints -- `p += 4` plus `body += 4` inside `run_login_body()` reproduces the old `p += 8`, and the buffered path keeps the same contract;
- the `content_length >= CONFIG_UPLOAD_BUF` bound is strict, so `config_buf[pre_acc] = NUL` at `pre_acc == postbody_len` still lands inside the buffer. `>` would have been an off-by-one;
- the `TSTATE_POSTBODY` check sits first in `handle_post()`, so a body segment that happens to start with "POST" is not re-parsed as a new request;
- the removed trailing `200 OK` was reachable only from the cmd branch; every other branch already returned;
- a body up to `CONFIG_UPLOAD_BUF` cannot overrun `execute_commands()`, since `cmd_idx` is clamped to `CMD_BUF_SIZE - 1` and returns `ERR_CMD_TOO_LONG`.

One finding looks blocking to me: the new wait state lets an unauthenticated client hold the single TCP connection slot indefinitely. Details inline, together with two gaps that let the fixed bug come back on a slightly differently shaped client, and a concrete target for @vDorst's refactor suggestion.

I have not rebuilt this, so the memory figures in the description are taken on trust.

httpd/httpd.c
```diff
@@ -558,11 +668,11 @@ void handle_post(void)
...
 			return;
 		}
-		execute_commands(p);
-		if (err_status != ERR_OK) {
-			send_bad_request();
+		postbody_endpoint = POSTBODY_CMD;
+		if (!post_body_take(p))
 			return;
-		}
+		run_cmd_body(p);
```
DrDoof, line 770: The buffered path bounds the body at the announced length:

```c
// never take more than announced: pipelined bytes are not body
take = postbody_len - pre_acc;
```

This path does not. `run_cmd_body(p)` runs to the NUL that `httpd_appcall()` wrote at `p[uip_len]`, so when `have > content_length` it also executes whatever follows the body. Same endpoint, two different contracts depending on how the client segmented the request.

`p[content_length] = NUL;` before the call closes it, and is safe by construction: `have >= content_length` implies `(p - uip_appdata) + content_length <= uip_len`, `uip_buf` is declared `[UIP_CONF_BUFFER_SIZE + 2]`, and `p[uip_len] = 0` is already written there.

httpd/httpd.c
```diff
@@ -500,12 +523,99 @@ uint8_t stream_upload(void)
...
+ * waits in TSTATE_POSTBODY for the rest: 0 means answer nothing yet.
+ */
+static uint8_t post_body_take(__xdata uint8_t * __xdata p)
+{
+	// static: automatic xdata locals get register-cached and spill
+	// through DSEG slots, which direct RAM has no room for
+	static __xdata struct httpd_state * __xdata s;
+	static __xdata uint16_t have;
+
+	have = uip_len - (p - uip_appdata);
```
DrDoof, line 668: This can underflow. On a truncated header `scan_header()` leaves `p` past the end of the data, because the loop steps over the NUL before breaking:

```c
if (!*p++)
        break;
```

The guard back in `handle_post()` then tests `!*p` at `uip_buf[uip_len + 1]`, which holds a stale byte from an earlier packet rather than a guaranteed 0. If that byte is non-zero and a Content-Type was seen, this subtraction yields 0xffff and "the whole body is already here" becomes true for a body that is not.

Pre-existing -- the old code called `execute_commands(p)` just as blindly -- but this line turns it into arithmetic that the rest of the new code depends on, so clamping it to 0 when `p` is past `uip_appdata + uip_len` seems worth the two instructions.

httpd/httpd.c
bloqaudio: Done as `header_value()` in the first commit of the series. It takes the lower-case name with the leading `\n` and the trailing `:` as one literal (`"\ncontent-length:"`), folds the request bytes to lower case as it compares, skips the blanks after the colon, and returns a pointer to the value instead of a length, so the call site does not add anything. All three header matches in `scan_header()` go through it now.

I did not add a macro: the preprocessor cannot lower-case a string, and prefixing `"\n"` alone did not seem worth hiding the anchor behind a name. If you would rather have the length returned in a `uint8_t` I can flip it; the pointer form saved a local at each call site.

httpd/httpd.c
bloqaudio: Folded into `parse_short()` as @DrDoof suggested below, so `Content-Length` shares the one loop that already serves the JSON endpoints.

httpd/httpd.c
bloqaudio: Agreed, and thanks for tracing it through `uip.c`. **The wait state now has a deadline**: the poll handler aborts the connection (RST, so the slot is free at once rather than after a FIN handshake with a client that has gone quiet) when the body has made no progress for five seconds.

I measured it on the 200 Hz system tick (`ticks`) rather than by counting polls: `idle()` runs `uip_periodic()` once per wake-up, and the CPU wakes on every interrupt, so a poll count runs fast under serial or NIC load. Two bytes of xdata for the start tick.

Verified on a board: headers followed by silence on both `/login` (unauthenticated) and `/cmd` get the RST at 5.0 s, and the next request on a fresh connection succeeds.

The multipart states (`TSTATE_MULTIPART`, `TSTATE_POST`) have the same shape of exposure and always did; they sit behind authentication, so I left them for a follow-up rather than widen this PR.

httpd/httpd.c
bloqaudio: Right, and that was the same silent failure for `Content-Type`: header names are compared case-insensitively by `header_value()` now and the whitespace after the colon is optional. Tested `Content-Length:12`, `content-length: 12` and a tab after the colon; all execute the split body.

On the 411 question: taken. **A plain POST with no usable `Content-Length` (absent, unparseable, or 0) answers `411 Length Required`.** Without one there is no way to know when the body has arrived, and the old `200 OK` for an empty body is the bug this PR fixes. Every real client sends it (browsers for fetch/XHR, curl, urllib, requests), so this only changes what a hand-written request gets told.

httpd/httpd.c
bloqaudio: Done, in its own commit: `parse_short()` saturates at `0xffff` and `Content-Length` goes through it. Checked the three existing consumers: `vlan_get()` refuses anything from 4095 up, `send_l2()` masks the index to the table and `l2_delete()` masks the high byte, so the sticky `0xffff` is safe for all of them. `/vlan.json?vid=65537` now answers like `vid=4095` instead of like `vid=1`.

httpd/httpd.c
bloqaudio: Fixed: `post_body_take()` writes the NUL at `p[content_length]` before returning 1, so both paths execute exactly the announced body. Verified with a trailing `\nhostname junk` after the announced length, in the same segment and in a later one; the junk no longer runs.

httpd/httpd.c
bloqaudio: Fixed at the source instead of clamping: `scan_header()` now breaks before stepping over the NUL, so on a truncated header it returns a pointer to the NUL and the existing `!*p` check in `handle_post()` sees it (own commit). With that, `p` can never lie past `uip_appdata + uip_len` when `post_body_take()` subtracts, so no clamp is needed there. Tested: a header cut before the blank line answers `404` immediately.

httpd/httpd.c
```diff
@@ -284,8 +284,10 @@ __xdata uint8_t *scan_header(__xdata uint8_t * __xdata p)
 
 	while (*p != '\r' || *(p + 1) != '\n' || *(p + 2) != '\r' || *(p + 3) != '\n') {
```
vDorst: maybe also useful to replace this with `!strstart(p, "\r\n\r\n")`.

httpd/httpd.c
```diff
@@ -537,12 +555,111 @@ uint8_t stream_upload(void)
...
+		send_bad_request();
+		return;
+	}
+	slen = strtox(outbuf, "HTTP/1.1 200 OK\r\nConnection: close\r\n\r\n");
+}
+
+
+static void run_login_body(__xdata uint8_t * __xdata body)
+{
+	body += 4; // Read also over "pwd="
```
vDorst: use `strstart()` to validate `pwd=` is there.

httpd/httpd.c
```diff
@@ -537,12 +555,111 @@ uint8_t stream_upload(void)
...
+	}
+	if (content_length >= CONFIG_UPLOAD_BUF) {
+		send_bad_request();
+		return 0;
+	}
+	memcpy(config_buf, p, have);
+	pre_acc = have;
+	postbody_start = ticks;
+	s = &(uip_conn->appstate);
+	s->tstate = TSTATE_POSTBODY;
```
vDorst: Why not use `((struct httpd_state *)(uip_conn->appstate))->tstate` directly.
It is just be used in only place.

httpd/httpd.c
```diff
@@ -537,12 +555,111 @@ uint8_t stream_upload(void)
...
+ * there at its announced length. Otherwise the bytes move to config_buf
+ * and the connection waits in TSTATE_POSTBODY for the rest: 0 means
+ * answer nothing yet, or that a refusal has been built.
+ */
+static uint8_t post_body_take(__xdata uint8_t * __xdata p)
+{
+	// static: automatic xdata locals get register-cached and spill
+	// through DSEG slots, which direct RAM has no room for
+	static __xdata struct httpd_state * __xdata s;
+	static __xdata uint16_t have;
```
vDorst: Is static needed? I see a reduction in xdata and code size when used non-static.
Same for `post_body_continue()`.

Please don't add comments about DSEG everywhere. I think it is not useful and only noisy.
I hope that you tested it first without xdata and when getting in trouble you put in on xdata.

httpd/httpd.c
```diff
@@ -537,12 +555,111 @@ uint8_t stream_upload(void)
...
+}
+
+
+/*
+ * Returns 1 when the whole body sits at p in the rx buffer, terminated
+ * there at its announced length. Otherwise the bytes move to config_buf
+ * and the connection waits in TSTATE_POSTBODY for the rest: 0 means
+ * answer nothing yet, or that a refusal has been built.
+ */
+static uint8_t post_body_take(__xdata uint8_t * __xdata p)
```
vDorst: On my setup p don't have to be passed via xdata.
`(__xdata uint8_t *p)` is just fine.

httpd/httpd.c
```diff
@@ -537,12 +555,111 @@ uint8_t stream_upload(void)
...
+	have = uip_len - (p - uip_appdata);
+	if (have >= content_length) {
+		// pipelined bytes past the announced length are not body
+		p[content_length] = NUL;
+		return 1;
+	}
+	if (content_length >= CONFIG_UPLOAD_BUF) {
+		send_bad_request();
+		return 0;
+	}
```
vDorst: I think `if (content_length >= CONFIG_UPLOAD_BUF) {` should be about line 607.

httpd/httpd.c
```diff
@@ -138,6 +138,29 @@ bool is_word(__xdata uint8_t *xdata_str_p, __code uint8_t * __xdata code_str_p)
 }
 
 
+/*
+ * Match a header field name case-insensitively. name is lower-case and
+ * spans the '\n' ending the previous line through the ':', so a match is
+ * anchored to a line start. Returns the start of the value, past any
+ * blanks, or 0 when the line is some other header.
+ */
+__xdata uint8_t *header_value(__xdata uint8_t *p, __code uint8_t * __xdata name)
```
vDorst: This can be just `__xdata uint8_t *header_value(__xdata uint8_t *p, __code uint8_t *name)`.

httpd/httpd.c
```diff
@@ -537,12 +555,111 @@ uint8_t stream_upload(void)
...
 
 
+static void run_cmd_body(__xdata uint8_t * __xdata body)
+{
+	execute_commands(body);
+	if (err_status != ERR_OK) {
+		send_bad_request();
+		return;
+	}
+	slen = strtox(outbuf, "HTTP/1.1 200 OK\r\nConnection: close\r\n\r\n");
```
vDorst: This is used in two places
```
httpd/httpd.c
565:    slen = strtox(outbuf, "HTTP/1.1 200 OK\r\nConnection: close\r\n\r\n");
765:                    slen = strtox(outbuf, "HTTP/1.1 200 OK\r\nConnection: close\r\n\r\n");
```

Maybe make it a function like the other responces like `send_bad_request()`.

httpd/httpd.c
```diff
@@ -537,12 +555,111 @@ uint8_t stream_upload(void)
 }
 
 
+static void run_cmd_body(__xdata uint8_t * __xdata body)
```
vDorst: Can be just `static void run_cmd_body(__xdata uint8_t * body)`

httpd/httpd.c
```diff
@@ -537,12 +555,111 @@ uint8_t stream_upload(void)
...
+	execute_commands(body);
+	if (err_status != ERR_OK) {
+		send_bad_request();
+		return;
+	}
+	slen = strtox(outbuf, "HTTP/1.1 200 OK\r\nConnection: close\r\n\r\n");
+}
+
+
+static void run_login_body(__xdata uint8_t * __xdata body)
```
vDorst: Can be just `static void run_login_body(__xdata uint8_t * body)`

vDorst (commented):
More general comments.
- Please try/compile a variable without `__xdata` first.
  When it consums SRAM change it so XDATA.
  See also my guide how to check if it is needed: https://github.com/logicog/RTLPlayground/pull/368#discussion_r3886861122
- Please remove comments about XDATA is needed because SRAM is used.
  For me that feels noisy and not useful in most cases.
  As mention before, I assume your tried it first without  explicitly putting it on XDATA.
- When `tstate` is accessed only once in the function.
  Just use `((struct httpd_state *)(uip_conn->appstate))->tstate = <value>`.
  Only if it has a downside doing it in this way.
- putting `static` on local variables seems unnecessary.

DrDoof:
Following up on the one thing I left open: I have rebuilt it now, against current main on a SWTGW218AS. Bank 0 is unchanged at 1111 bytes free and internal RAM is identical, DSEG 0x80 and OSEG 0x08 on both sides. I get +21 xdata and +710 in BANK1 rather than 23 and 914, which I put down to #351 and #368 having landed since you measured.

The `pwd=` skip @vDorst points at is the same spot I would have raised on length, and validating the prefix covers both. Nothing further from me.

httpd/httpd.c
bloqaudio: Done, folded into the `header_value()` commit.

httpd/httpd.c
bloqaudio: Done: `run_login_body()` now requires the `strstart(body, "pwd=")` prefix and takes the invalid branch (302 to `login.html`) otherwise; the test script got a case for it.

httpd/httpd.c
bloqaudio: Done. `appstate` is embedded in `uip_conn`, so it reads `uip_conn->appstate.tstate = TSTATE_POSTBODY;`, and with that the local vanished from both functions. No downside found: the RAM map is identical and the code smaller.

httpd/httpd.c
bloqaudio: You are right on both. Rebuilt with plain automatics throughout the new code and measured per your guide from #368: the internal RAM map is byte for byte identical to `main` on both machines, the DSEG section of `httpd.asm` carries slocs only for the pre-existing `stream_upload()` and `httpd_appcall()`, and both xdata and code shrank (xdata +5 vs `main` instead of +23, code +783 instead of +914). The `static` idiom came from an earlier series where automatics did cost direct RAM; here they do not, so it is gone. The comments about it are gone too, throughout.

httpd/httpd.c
bloqaudio: Done, same for `run_cmd_body()` and `run_login_body()`.

httpd/httpd.c
bloqaudio: Done, it now sits right after the 411 check, before the completeness test. Behavior-identical either way (`uip_len` is far below `CONFIG_UPLOAD_BUF`, so an oversized announcement could never take the complete-body path), but it reads in the right order now.

httpd/httpd.c
bloqaudio: Done.

httpd/httpd.c
bloqaudio: Done, and the config-upload path uses it too.

httpd/httpd.c
bloqaudio: Done, covered by the measurement in the statics thread.

httpd/httpd.c
bloqaudio: Done, same measurement.

bloqaudio:
All of the general points are adopted in the updated branch: plain automatics first, measured per the guide from #368 (internal RAM map byte for byte identical to `main` on both machines, the `httpd.asm` DSEG section clean), the xdata and DSEG comments removed throughout, `tstate` stored directly where it is touched once, and the statics gone. Net effect of this round: xdata +5 instead of +23 and code +783 instead of +914, both relative to `main`.

Re-tested on a SWTG018AS V2.1.0 board: the full segmentation matrix passes (27 scripted cases including a refused login body without the `pwd=` prefix, both headers-then-silence holds dropped at 5.0 s with the connection slot free afterwards, and the config round trip and size bounds in every split shape).

vDorst:
Can you rebase and resolve the conflic.

bloqaudio:
Rebased onto current `main` (74a1c68) and pushed; the branch is now the same five commits on top of the #374 merge. The conflict was the tail of `handle_post()`, which #374 split into the two fragment handlers: the resolution keeps that shape, and the one change this series made inside the old block, the inline 200 response becoming `send_ok()`, now lives in `handle_config_fragment()`. One duplicate `ticks` declaration dropped. Re-measured against the new base: internal RAM map, DSEG, OSEG and bank 0 byte for byte identical to `main` on both machines, xdata +5 and BANK1 +783, the same deltas as before the rebase; `make machine_check` passes. Re-tested on the board: the POST body harness including the held-connection cases, config round trip in six segmentations, and the 4095/4096/4300 bounds.

vDorst:
Thanks

PR #380, opened 2026-08-31
Fix VLAN command documenation. 'u'-suffix is not accepted anymore.

### LINKED ISSUE #378: setting untagged VLAN changed
I updated from a build from May to the recent main and it seems that setting untagged VLANs changed, but the error message is the old style.

Do I assume right that untagged is now just without additional "u" like (since I only saw handling for t)
`vlan 1 1 2 3t`

Then we need to update the error message in cmd_parser.c:
`print_string("Error: vlan (<vlan-id>|show) [port][t/u]...\n");`

vDorst:
Fix VLAN command documentation.
`u`-suffix is not accepted anymore after #326.
The new parser now checks if a SPACE or `t` is behind the port number.
Old parser did not check this and accepted any character.

Remove the optional `u` from the VLAN documentation and help message.

Fixes #378

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,11 @@
   - VLAN don't accept port `u`-suffix anymore.
     So `vlan 1 4u` is not valid anymore.
     Replace it with `vlan 1 4`.
+- Commands
+  - port zero/`0` is treated as the `CPU_PORT`. #326
+  - Many commands don't accept the CPU_PORT anymore. See #334.
+    When the CPU_PORT is needed, the command/service will add the CPU_PORT automaticly.
+    Only `isolate` accept the CPU_PORT as destination port.
 
 ## [v0.1_aplha] - 2025-11-03
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,20 @@
+# Change Log
+
+## [0.x] - 2026-xx-XX
+
+## Added
+
+## Changed
+
+## Fixed
+
+## Breaking changes
+
+- Config
+  - VLAN don't accept port `u`-suffix anymore.
+    So `vlan 1 4u` is not valid anymore.
+    Replace it with `vlan 1 4`.
+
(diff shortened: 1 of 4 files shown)


feelfree69 (approved):

DrDoof:
Docs change looks right to me, the `u` never actually did anything.

One thing though. The old wording called it optional, so someone may well have typed it and saved. That config won't come back after a reboot: the digit parses fine, the member bit is already set, then the `u` fails the space-or-nul check and the line bails out before `vlan_create()`. The VLAN just isn't there, and nothing obvious says why.

The web page only ever wrote a bare port or `t`, so it's the people who read the docs and used the console who get bitten. Bit unlucky for them.

Two lines in the port loop would let it swallow a stray `u`. I don't feel strongly either way, I'd just rather it was a decision than a surprise.

vDorst:
> The web page only ever wrote a bare port or t, so it's the people who read the docs and used the console who get bitten. Bit unlucky for them.
Two lines in the port loop would let it swallow a stray u. I don't feel strongly either way, I'd just rather it was a decision than a surprise.

I would like to remove it.
Lets add a Changelog to add this as a breaking change.

DrDoof:
Sounds right. Removing it is the cleaner end of the two, and a changelog entry is what the case was missing, since the only people it can bite are the ones who did what the docs said.

Say if you want me to write it, otherwise it's yours.

vDorst:
I already added a CHANGELOG.md to this PR.
So I think this is fixed and ready for review.

vDorst:
I have to check: the website also has to reject the `u`.

vDorst:
Update JS regex and Changelog about the this and CPU_PORT change.

vDorst:
Can somebody look at this PR?
So we can set #315 in motion.

feelfree69 (approved):
Looks good to me.

logicog:
Perfect! Thanks for taking on this housekeeping task!

PR #382, opened 2026-08-31
sfp: stop the web pages printing the previous slot's bytes

### LINKED ISSUE #381: sfp-name is corrupted on boot when no SFP is insert. This also effects information.json
```json
{"ip_address":"192.168.10.247","ip_gateway":"192.168.10.1","ip_netmask":"255.255.255.0","syslog_server_ip":"0.0.0.0","mac_address":"06:05:16:11:11:11","hostname":"RTLPlayground-11111","sw_ver":"v0.1.0-70b1abc-dirty","build_date":"2026-08-31 20:27:52","hw_ver":"SWGT024 V2.0 Managed","flash_size":"4 MB","sfp_slot_0":"    A0  ␃R®0␎  A0  ␃R®0␎      A0  ","sfp_slot_1":"DELL            WTRD1           A0  "}
```

Data of the first SFP slot is garbage.

On the webpage, info is missing because the json data is non-compliant.

Image

I didn't notice before because I always have two SFP modules insert.
I think #345 is the issue.

DrDoof:
Fixes #381.

An empty SFP slot used to read back `0xa0` per byte, and the filter in `send_sfp_info()` dropped it, so the field came out empty. Block reads report the failure instead and leave `sfp_buf` holding whatever the last successful read put there, so the two callers in `page_impl.c` that ignore the return print the other slot's bytes. Control characters among them are what makes the JSON non-compliant, which is why the page shows nothing.

Both callers now say nothing when there is nothing to read, which is what the other eight callers of `sfp_read_block()` already do. No unchecked call remains in the tree.

Bank 0 gains 55 bytes against main on SWTGW218AS and SWTG024AS_V2_0, and the build is warning free.

### DIFF
--- a/httpd/page_impl.c
+++ b/httpd/page_impl.c
@@ -178,8 +178,8 @@ void send_sfp_info(uint8_t sfp)
 {
 	// This loops over the Vendor-name, Vendor OUI, Vendor PN and Vendor rev ASCII fields
 	for (uint8_t i = 16; i < 64; i++) {
-		if (!(i & 0xf))
-			sfp_read_block(sfp, i, 16);
+		if (!(i & 0xf) && !sfp_read_block(sfp, i, 16))
+			return;
 		if (i < 20 || i >= 60 || (i >= 36 && i < 40)) // Skip Non-ASCII codes
 			continue;
 		uint8_t c = sfp_buf[i & 0xf];
@@ -195,7 +195,8 @@ void sfp_send_data(uint8_t slot, uint8_t reg, uint8_t len)
 	if (len > 16)
 		return;
 
-	sfp_read_block(slot, reg, len);
+	if (!sfp_read_block(slot, reg, len))
+		return;
 
 	for (uint8_t i = 0; i < len; i++)
 		byte_to_html(sfp_buf[i]);


feelfree69 (approved):

vDorst:
I also tested this it yesterday and it works.

It is good that json is not corrupted when the module doesn't response or there is a I2C issue.
I think we also need to use the knownlage that the module is pressent, so we just skip try to read all together. maybe we want a different responce. Like `"sfp_slot_0": null` or `<Not Pressent>`/`<I2C error>`.

DrDoof:
Agreed, and the console already does it. `sfp` on its own checks `pin_detect` first and says " - empty", then falls back to " - I2C read failed on this slot" if the read fails. So the CLI tells those two apart and the JSON is the one that doesn't.

`gpio_pin_test()` is in the common area so `send_sfp_info()` can reach it. Detect is active low, true means nothing in the slot.

I'd rather not grow this one though. It changes what the page gets, and main_info.js prints the value straight out, so `null` would literally show up as the word null. Empty string for absent and something for the error case is probably right, but that wants deciding with the page rather than bolted onto a bugfix. I'll put it up once this lands.

vDorst:
> I'd rather not grow this one though. It changes what the page gets, and main_info.js prints the value straight out, so null would literally show up as the word null. Empty string for absent and something for the error case is probably right, but that wants deciding with the page rather than bolted onto a bugfix. I'll put it up once this lands.

Thanks! Thanks for making another PR.

Issue #379, opened 2026-08-31
passwd hangs in flash


Kabbone:
When I execute passwd over serial everything works fine, but as soon as I put it in either config.txt or save in flash from the GUI it seems to hang indefinitely, even in serial and doesn't completely come up anymore (no WebGUI)

Device is a Horaco SWTGW218AS with a borked flash but the old build works fine so I assume it's not the flash

DrDoof:
Two questions while I dig into this.

Which build is the one that works? You said the old one is fine, and there have been a lot of merges lately, so knowing roughly where the good one sits would narrow the window a lot.

And without telling me the password: does it contain a space, or anything outside plain ASCII letters and digits? Asking because `parse_passwd()` explicitly handles a space in there, while the page that composes the saved config only accepts a password with no whitespace. Those two already disagree. It wouldn't hang anything by itself, but it is the one place I have found so far where the console and the stored config treat the same command differently.

DrDoof:
Think I have it, and `passwd` is not really the culprit.

`send_cmd_log()` walks the ring with `while (p != cmd_history_ptr)` and masks `p` every turn, but the writer leaves the pointer unmasked: `cmd_history[cmd_history_ptr++] = '\n'` right after the masked assignment, so it can end up at 0x400. `p` can never reach that, the loop never ends, and it keeps writing past the 2500 bytes of `outbuf`. Dead serial, no GUI, power cycle. The console `history` command has the same loop.

`flashSave()` in system.js fetches `/cmd_log` on every save, which is why saving trips it and typing the command doesn't. The pointer moves by command length, so a `passwd` line is just what happened to land yours on 0x3FF.

```diff
-cmd_history[cmd_history_ptr++] = '\n';
+cmd_history[cmd_history_ptr] = '\n';
+cmd_history_ptr = (cmd_history_ptr + 1) & CMD_HISTORY_MASK;
```

I also ran the replay of a stored `passwd` line through a simulator in fourteen shapes and it completed every time, so the line itself is clean.

Separately: `execute_config()` does `goto config_done` on any tokenizer error, so one bad line drops the whole rest of the config and you come up with no IP. That alone would look like "doesn't completely come up".

Kabbone:
Thx for looking into it, you were too fast for me to answer 😄 
The tested password was only alphanumeric, so I assume you are on the right track

DrDoof:
[...]

PR #384, opened 2026-09-01
cmd: keep the history pointer inside the ring

DrDoof:
Fixes the hang in #379.

`cmd_history_ptr` is masked when it advances over the command text, then incremented once more for the newline without masking:

```c
cmd_history_ptr = (cmd_history_ptr + i) & CMD_HISTORY_MASK;
cmd_history[cmd_history_ptr++] = '\n';
```

so it can come to rest at `CMD_HISTORY_SIZE`, one past the ring. Both readers start from it and mask their own index every step:

```c
uint16_t p = (cmd_history_ptr + 1) & CMD_HISTORY_MASK;
while (p != cmd_history_ptr) { ... p = (p + 1) & CMD_HISTORY_MASK; }
```

`p` never reaches an index outside the ring, so the walk does not end. `history` on the console spins there, and `send_cmd_log()` does the same while writing `outbuf[slen++]` well past its 2500 bytes.

`flashSave()` in `html/system.js` fetches `/cmd_log` on every save, which is why saving settings is what trips it. Whether the pointer lands one past the end depends on the accumulated length of the commands entered, so it is not deterministic between devices.

Build unchanged and warning free on SWTGW218AS and SWTG024AS_V2_0.

### DIFF
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -1771,7 +1771,8 @@ void cmd_parser(void) __banked
 			// Copy last cmd-buffer to history.
 			cmd_history_ptr = (cmd_history_ptr + i) & CMD_HISTORY_MASK;
 			__xdata uint16_t p = cmd_history_ptr;
-			cmd_history[cmd_history_ptr++] = '\n';
+			cmd_history[cmd_history_ptr] = '\n';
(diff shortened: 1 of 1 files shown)


vDorst:
Thanks! I have seen it too that console reacted weard sometimes.

PR #389, opened 2026-09-02
httpd: let the login page fetch its translations

DrDoof:
The login page never comes up translated and it took me longer than I'd like to admit to see why. login.html does ask for i18n.js, but that file isn't on the small list of things httpd hands out before you log in, so the request gets the redirect back to login instead. The page then falls back to the English sitting in the markup, whatever your browser asked for.

So it's one line, adding i18n.js to that list. The file only holds the interface strings the page is already showing you in English, so there's nothing in it that isn't on screen anyway.

If you want to see it, point a browser set to a language you have a translation for at the login page. It picks the strings up afterwards.

This fell out of a bigger branch I keep locally, hence the one liner. Happy to move it if you'd rather have it somewhere else.

### DIFF
--- a/httpd/httpd.c
+++ b/httpd/httpd.c
@@ -994,6 +994,7 @@ void httpd_appcall(void)
 			if (!authenticated && !(f_data[entry].start == FDATA_START_login_html 
 						|| f_data[entry].start == FDATA_START_port_svg 
 						|| f_data[entry].start == FDATA_START_sfp_svg
+						|| f_data[entry].start == FDATA_START_i18n_js
 						|| f_data[entry].start == FDATA_START_style_css)) {
 				send_to_login();
 				goto do_send;


vDorst:
Is this still a issue after #315 is merged?
Because all the JS files is just one file.

DrDoof:
It isn't merged yet, so on main i18n.js is still a file of its own and the login page still gets bounced back.

I went and looked at the branch anyway, and it doesn't fix it there either, it just renames it. login.html on #315 asks for /main.js, and the pre-auth list in httpd.c is untouched, still login.html, port.svg, sfp.svg and style.css. So the login page would get the redirect for main.js instead, same silence, different filename.

Which is a bit of an argument for having something that fails loudly when a file drops off that list, but that's a separate thought.

Either way it stays a one-liner. If #315 lands first I'll rebase this on top and swap the entry, whichever order suits you.

vDorst:
thanks

PR #390, opened 2026-09-02
httpd: reap idle TCP connections

DrDoof:
Reworked down at the TCP level, the way the thread landed on, so the next service that opens a socket is covered too.

uIP has no keepalive and no idle timeout on an ESTABLISHED connection. If the peer just goes away, no FIN, no RST, the connection sits there. A crashed client does it, and so does a SYN that lands while the switch is still coming up. A leaked connection is usually nothing to worry about, but with UIP_CONF_MAX_CONNECTIONS at 1 it's the only slot there is, so that's the server gone. Port 80 stops answering SYN at all.

The mechanism now lives in uip.c. The retransmission timer does nothing while a connection has nothing in flight, so the idle count borrows it: one aging strobe a second off the uip_periodic() rate, cleared by any arriving segment, and past the limit the connection is dropped through the same path UIP_MAXRTX already uses. Both knobs sit in uipopt.h and default to off, so plain uIP is unchanged unless a project asks for it; ours asks for 30 s. I said earlier this cost no RAM, the honest figure is two bytes for the strobe, shared by the whole stack.

One thing I tripped over wiring it up: httpd never handled uip_timedout() at all, so the UIP_MAXRTX abort that already exists was leaving the transfer state stuck on a dead connection too. One line in the cleanup covers both.

Tested on the switch (SWTGW218AS). I leave a connection in ESTABLISHED with nothing sent, it takes the one slot and port 80 goes silent, then the timeout fires, the server RSTs the dead one and answers again, no reboot. Two runs came back at 25 and 29 s, the early side of 30, which is expected since interrupt wake-ups tick the strobe a bit faster than once a second.

### DIFF
--- a/uip/uip-conf.h
+++ b/uip/uip-conf.h
@@ -97,6 +97,16 @@ typedef unsigned short uip_stats_t;
 #define UIP_CONF_MAX_CONNECTIONS 1
 
 /**
+ * uip_periodic() runs from idle() roughly once per system tick
+ * (SYS_TICK_HZ, 200 Hz; interrupt wake-ups add sweeps, so the timeout
+ * can only fire early, never late). A silent ESTABLISHED peer is
+ * dropped after about 30 s: with a single connection slot it would
+ * otherwise wedge the web UI until a power cycle.
+ */
+#define UIP_CONF_IDLE_PERIODS 200
+#define UIP_CONF_IDLE_TIMEOUT 30
+
+/**
  * Maximum number of listening TCP ports. TODO: increase this!
  *
  * \hideinitializer
--- a/uip/uip-conf.h
+++ b/uip/uip-conf.h
@@ -97,11 +97,16 @@ typedef unsigned short uip_stats_t;
 #define UIP_CONF_MAX_CONNECTIONS 1
 
 /**
- * uip_periodic() runs from idle() roughly once per system tick
- * (SYS_TICK_HZ, 200 Hz; interrupt wake-ups add sweeps, so the timeout
- * can only fire early, never late). A silent ESTABLISHED peer is
- * dropped after about 30 s: with a single connection slot it would
- * otherwise wedge the web UI until a power cycle.
+ * This httpd closes after every response, so an ESTABLISHED connection
+ * that stays idle is a peer that died or never sent its request. With a
+ * single connection slot it would hold the web UI until the next power
+ * cycle, so it is aged out instead.
+ *
+ * uip_periodic() runs from idle() once per system tick; interrupt
+ * wake-ups only add sweeps, so the timeout can fire early but never
+ * late.
+ *
+ * \hideinitializer
  */
 #define UIP_CONF_IDLE_PERIODS 200
 #define UIP_CONF_IDLE_TIMEOUT 30
--- a/httpd/httpd.c
+++ b/httpd/httpd.c
@@ -842,7 +842,7 @@ void httpd_appcall(void)
 	} else if (uip_closed()) {
 		dbg_string("Connection closed\n");
 		s->tstate = TSTATE_CLOSED;
-	} else if (uip_aborted()) {
+	} else if (uip_aborted() || uip_timedout()) {
 		dbg_string("Connection aborted\n");
 		uip_close();
 		s->tstate = TSTATE_CLOSED;
--- a/uip/uip.c
+++ b/uip/uip.c
@@ -171,6 +171,11 @@ __xdata struct uip_udp_conn * __xdata uip_udp_conn;
 __xdata struct uip_udp_conn uip_udp_conns[UIP_UDP_CONNS];
 #endif /* UIP_UDP */
 
+#if UIP_IDLE_TIMEOUT
+__xdata static u8_t uip_idle_prescale;
+__xdata static u8_t uip_idle_age;
+#endif /* UIP_IDLE_TIMEOUT */
+
 __xdata static u16_t ipid;           /* Ths ipid variable is an increasing
 				number that is used for the IP ID
 				field. */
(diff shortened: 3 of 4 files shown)


vDorst:
```
uip/uip.c
Function: uip_process()
729:      if(uip_connr->timer == UIP_TIME_WAIT_TIMEOUT) {
```
It seems it has all the bits init.
So I wonder is uip used properly so it walks often enough thru this function to generate a timeout.

logicog:
I tend to agree with @vDorst , this is a TCP issue. Should we add other TCP services, it will bite us again if we do not fix it at the TCP-level. My understanding is that idle() calls handle_tx() which calls uip_periodic() on all open connections. This does
```
#define uip_periodic(conn) do { uip_conn = &uip_conns[conn]; \
                                uip_process(UIP_TIMER); } while (0)
```
uip_process handles the UIP_TIMER by doing all the time-outs. Maybe the clock rate assumed by uip is not correct, because we changed it a couple of times in the past?

feelfree69:
I don't think so. An open TCP connection stays open forever after the client silently disappeared. This is by design.

>  if(uip_connr->timer == UIP_TIME_WAIT_TIMEOUT) {

Without digging into the code, the TIME_WAIT is a TIMEOUT while waiting for the final ACK during the regular connection closure.

DrDoof:
@feelfree69 is right. I went through our uip.c to be sure: the UIP_TIMER branch only ever ages TIME_WAIT and FIN_WAIT_2, which is the closure handshake, and connections with outstanding data on the retransmit path. Idle ESTABLISHED sits behind the uip_outstanding() check and is never touched, so no polling rate will make it time out. Stock uIP simply has no keepalive.

@logicog your point about the next TCP service is fair though, and the move down turns out cheaper than I expected. conn->timer is unused whenever nothing is outstanding, so the idle count can live right there, no new field, no extra RAM. Count it up in that same branch, abort an ESTABLISHED connection past the limit, reset wherever a real segment arrives. httpd would behave exactly as it does in this PR.

Say the word and I'll rework it that way.

logicog:
> Say the word and I'll rework it that way.

Can you rework the PR that way? You seem to have thought everything through already and this will be the superior solution.

DrDoof:
Done, the PR carries the uip.c version now. One correction to what I said above: it does cost two bytes after all, a shared 1 Hz divider, because uip_periodic() here runs per system tick and a u8 timer can't stretch to 30 s at that rate. Still nothing per connection.

Wiring it up turned up a small hole that was already there: httpd never handled uip_timedout(), so the existing UIP_MAXRTX abort had the same stuck-state problem this fixes. One line in the cleanup sorts both.

Defaults are off in uipopt.h, this project opts in at 30 s. Tested on the switch: a connection stuck in ESTABLISHED with nothing sent holds the one slot and port 80 goes quiet, then the server RSTs it and answers again after the timeout, no reboot. Two runs recovered at 25 and 29 s, early side of 30 as expected.

uip/uip-conf.h
```diff
@@ -96,6 +96,16 @@ typedef unsigned short uip_stats_t;
  */
 #define UIP_CONF_MAX_CONNECTIONS 1
 
+/**
+ * uip_periodic() runs from idle() roughly once per system tick
+ * (SYS_TICK_HZ, 200 Hz; interrupt wake-ups add sweeps, so the timeout
+ * can only fire early, never late). A silent ESTABLISHED peer is
+ * dropped after about 30 s: with a single connection slot it would
+ * otherwise wedge the web UI until a power cycle.
```
logicog: Are you sure 30seconds is not too short? I investigated and believe Linux times out a TCP connection if after 2 hours the keepalive does not work. Basically, Linux does not actually timeout a connection but tries keepalive and if that fails, then closes the connection.
These are the defaults on Debian Trixie:
```
$ cd /proc/sys/net/ipv4$
$ cat tcp_keepalive_time 
7200
$ cat tcp_keepalive_probes 
9
$ cat tcp_keepalive_intvl 
75

```

uip/uip-conf.h
```diff
@@ -96,6 +96,16 @@ typedef unsigned short uip_stats_t;
  */
 #define UIP_CONF_MAX_CONNECTIONS 1
 
+/**
```
logicog, line 99: Please don't put the values (30s/200Hz)  into the comment. The values may change.

logicog (commented):
Code looks great. Need to test, still. Just 2 small comments.

uip/uip-conf.h
DrDoof: Done, the numbers are out of it. The comment now says why an idle connection gets aged out and leaves the values to the defines underneath, where they can change without making the prose wrong.

uip/uip-conf.h
DrDoof: Fair question, and Linux is the right place to look. I think this server is a different animal though.

Linux keepalive is opt-in and it exists for protocols that deliberately hold a socket open with nothing on it, so it can afford to wait two hours before it even starts probing. This httpd closes after every response and advertises `Connection: close` on every one of them, so it has no long lived idle connection to protect. The only idle ESTABLISHED here is the gap between the handshake and the request line, which is milliseconds, or a peer that has gone.

There's a precedent a few lines up in httpd.c, too: `POSTBODY_TIMEOUT` already gives up on a connection whose POST body hasn't arrived within 5 s. 30 s is six times more patient than that, for a client that hasn't even sent a request line yet.

The other half of it is the asymmetry. With one slot the timeout is the length of the outage. Too short costs a handshake the server was going to force anyway, too long and the web UI is gone for the duration, which is where this started.

That said, it's your call and it's a one character change. Happy to go to 60 or 120 if you'd rather have the headroom. The ceiling is 255, the counter is a u8.

DrDoof:
Rebased onto current main, no conflicts. Builds for SWTGW218AS and `machine_check` is clean across every machine.

One housekeeping note: the thread where you asked about 30 s now shows as outdated, because the comment block it pointed at is the same one I rewrote to take the numbers out. My answer is still there if you unfold it, but the question is worth having in one piece where it can't get buried, so, short version:

This httpd closes after every response and says `Connection: close` on all of them, so it never holds an idle connection on purpose. The only idle ESTABLISHED here is the gap before the request line, or a peer that's gone. `POSTBODY_TIMEOUT` a few lines up already gives up on a stalled POST body after 5 s, so 30 s is six times more patient than something the tree already ships. And with one connection slot the timeout is the length of the outage, which is the asymmetry: too short costs a handshake, too long costs the web UI.

It's your call though, and it's one character. 60 or 120 if you'd rather have the headroom; 255 is the ceiling, the counter is a u8.

logicog:
I think it is good as it is and for the application we have. I fully agree with you on this. Since we can also fix it in the future easily by making the counters longer, it is also future-proof.

PR #392, opened 2026-09-03
Added Siscolink SL-G0402F support

eeeple:
Added Siscolink SL-G0402F support.
Tested 2026-09-03 with latest version.


feelfree69:
Thanks! Could you please rotate the picture to be naturally orientated?
Apart from that, everything's looking perfect!

eeeple:
Aww shucks! I didn't even check, sorry! Should be better now :ok_hand:

feelfree69:
Thanks!
@logicog / @vDorst I know you are always looking at the commit history. For such a small change, I would just use squash and merge. Is that ok?

logicog:
I am sure we can do that for such a small commit, which anyway all belongs together.

tofurky:
A bit late - but per the included image and a quick Google (https://www.amazon.com/stores/SICSOLINK/page/B09BD807-E0C9-4A6D-8136-E4ECFCA64DCB), it seems this should be S**icso**link, not S**isco**link. Repeated a few times.

PR #395, opened 2026-09-03
Fix l2 cli static entry and use wait_table_ready() in more places

vDorst:
Fix same issue as #355 but for CLI.
CPU was not shown as `static`.

Move `wait_table_ready()` up so it can replace all the parts in the code.
This saves around 150 bytes of code.
```c
do {
  reg_read_m(RTL837X_TBL_CTRL);
```

### DIFF
--- a/rtl837x_port.c
+++ b/rtl837x_port.c
@@ -30,6 +30,14 @@ __xdata	uint32_t l2_head;
 
 __xdata struct vlan_settings vlan_settings;
 
+// Wait for the TABLE ready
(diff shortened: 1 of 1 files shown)


logicog:
I went through the code and it looks all very good to me.

vDorst:
Rebased.

Removed the SFR_DATA optimization.
Going to put it in another PR which other changes.

@logicog can you review it again.

feelfree69 (approved):

PR #396, opened 2026-09-04
Fix printing of syslog portNumber in WebUI

feelfree69:
A trivial one-liner: syslog portNumber was decoded as uint8 instead of uint16.


vDorst:
Thanks

PR #397, opened 2026-09-04
stp: treat a link aggregation group as one entity

DrDoof:
STP on this tree has no idea link aggregation exists. Every member of a LAG is its own port to it, so it sees several parallel paths to the same neighbour, does exactly what it is meant to do, and blocks all but one of them. A four-link aggregate ends up carrying one link's worth. The LAG side is fine and the STP side is fine; it is the join between them that is missing.

This puts entities above the physical ports, 10 to 13 for the four groups, so an entity is either a port or a LAG and everything downstream stops having to care which. Membership comes from `port_lag_members_get()` from #351, re-read once a second, so a group that gains or loses ports is picked up without anything else noticing. `stp_ent_apply()` is the only place that knows the formula for an entity's state, and it writes that state to every member. Members drop out of the port list in `stp status` and on the STP page, because they are not entities any more, the group is.

With no LAG configured the tree reduces to what is there today on every path. That was the case I most wanted not to break.

Tested on a SWTGW218AS: a static LAG of two ports, then a cable from one member back to an ordinary port on the same switch, so there is a loop through the aggregate. The LAG goes to blocking and the ordinary port stays forwarding, which is the right way round, the aggregate being entity 12 and losing the tie against port 3. The part I wanted to see was the ASIC rather than the status page: `regget 5310` came back `0x000ff577`, and both members are blocked in silicon, not only the one with the cable in it. Traffic on the looped pair stayed at BPDU rate throughout.

Saved and rebooted with the loop still in place. It latches from the first second and held over 40 s, register `0x000ff5f7`.

This is meant to replace #325, which carries the same idea but as 61 commits from before the trunk/lag rename, 45 of which have since landed here by other routes. Happy to close that one in favour of this, or the other way round if you prefer.

### DIFF
--- a/doc/stp.md
+++ b/doc/stp.md
@@ -58,6 +58,28 @@ port set to admit tagged frames only (`ingress <port>t`) never delivers one
 to the CPU. `stp_setup()` prints a warning for every STP-enabled port in that
 state.
 
+## Link aggregation
+
+A LAG is a port to the protocol, the way the vendor firmware presents it as
+Trk1 to Trk4. The four groups are entities alongside the nine physical ports:
+they carry their own path cost, priority, edge and guard settings, they hold
+their own timers, and they appear as their own rows on the Spanning Tree page.
+
+```
+stp lag 1 cost 10000    # the group decides, not its members
+stp lag 1 edge off
+```
+
+Membership comes from the aggregation registers the `lag` command writes, re-read
+once a second, so a group changing under LACP is picked up without any
+coordination between the two. A member port is not an STP port of its own:
+`stp port <n>` on a member says which group to configure instead. State
+changes are written to every member in one register write, since the hardware
+has no per-group state, and BPDUs are transmitted through the lowest
+member while carrying the group's own port id. Losing one member of a live
+LAG is not a topology change; the logical port only goes down with its last
+link.
+
 Port states live in `RTL837X_MSTP_STATES (0x5310)`, two bits per port:
 `00` disabled, `01` blocking, `10` learning, `11` forwarding. A port in
 blocking forwards nothing between ports, but it still sends what the CPU
--- a/doc/stp.md
+++ b/doc/stp.md
@@ -60,23 +60,27 @@ state.
 
 ## Link aggregation
 
-A LAG is a port to the protocol, the way the vendor firmware presents it as
-Trk1 to Trk4. The four groups are entities alongside the nine physical ports:
-they carry their own path cost, priority, edge and guard settings, they hold
-their own timers, and they appear as their own rows on the Spanning Tree page.
+A LAG (Link Aggregation Group) is handled like an additional port by the STP
+protocol, with its own timers and states. The switch devices support up to 4
+LAGs, which are shown alongside the physical ports in the STP status. A group
+carries its own path cost, priority, edge and guard settings, and gets its own
+row on the Spanning Tree page.
 
 ```
 stp lag 1 cost 10000    # the group decides, not its members
 stp lag 1 edge off
 ```
 
-Membership comes from the aggregation registers the `lag` command writes, re-read
-once a second, so a group changing under LACP is picked up without any
-coordination between the two. A member port is not an STP port of its own:
-`stp port <n>` on a member says which group to configure instead. State
-changes are written to every member in one register write, since the hardware
-has no per-group state, and BPDUs are transmitted through the lowest
-member while carrying the group's own port id. Losing one member of a live
+LAGs are configured via the `lag` command. Once a port is a member of a LAG it
+can no longer be configured individually for STP, so `stp port <n>` on a member
+tells you which group to configure instead. Membership is re-read from the
+aggregation registers once a second, so a group changing under LACP is picked
+up without any coordination between the two.
+
+The switch hardware does not handle STP state for a LAG as a whole. State
+changes for the members have to be done by the firmware, updating every member
+port, which is possible using a single register write. BPDUs go out through the
+lowest member and carry the group's own port id. Losing one member of a live
 LAG is not a topology change; the logical port only goes down with its last
 link.
 
--- a/rtl837x_stp.c
+++ b/rtl837x_stp.c
@@ -47,31 +47,36 @@ __xdata uint8_t  stp_rstp;		/* 1 = RST BPDUs, 0 = legacy Config BPDUs */
 __xdata uint8_t  stp_txhold;		/* BPDUs per port per second */
 
 
-__xdata uint8_t  stp_pflags[10];
-__xdata uint32_t stp_pcost[10];		/* 0 = auto */
-__xdata uint8_t  stp_pprio[10];
-__xdata uint8_t  stp_pp2p[10];		/* admin point-to-point: 0 auto, 1 on, 2 off */
+__xdata uint8_t  stp_ent_of[10];
+__xdata uint16_t stp_lag_mask[STP_LAG_COUNT];
+__xdata uint8_t  stp_ss_i;
+__xdata uint8_t  stp_map_dirty;
+__xdata uint8_t  stp_pflags[STP_ENTITIES];
+__xdata uint32_t stp_pcost[STP_ENTITIES];	/* 0 = auto */
+__xdata uint8_t  stp_pprio[STP_ENTITIES];
+__xdata uint8_t  stp_pp2p[STP_ENTITIES];	/* admin point-to-point: 0 auto, 1 on, 2 off */
 
 /* Designated bridge, port and cost last heard on the port; stp_bpdu_age tells
  * whether they are still current.
  */
-__xdata struct bridge stp_dbridge[10];
-__xdata uint16_t stp_dpid[10];
-__xdata uint32_t stp_dcost[10];
+__xdata struct bridge stp_dbridge[STP_ENTITIES];
+__xdata uint16_t stp_dpid[STP_ENTITIES];
+__xdata uint32_t stp_dcost[STP_ENTITIES];
 
 /* ---- Status / runtime ---- */
 __xdata struct bridge root_bridge;
 __xdata uint32_t root_bridge_cost;	/* our cost to the root (rx cost + root port cost) */
 __xdata uint8_t  stp_root_port;		/* 0xff = we are the root */
 __xdata uint16_t stp_tc_count;
-__xdata uint16_t stp_scratch16;	/* scratch for status printing only */
+__xdata uint16_t stp_scratch16;	/* scratch for status printing and the lag map */
+__xdata uint8_t  stp_st_of;
 
-__xdata uint16_t port_timers[10];	/* listen-period countdown (0 = not listening) */
-__xdata uint16_t port_hello[10];	/* hello TX countdown */
-__xdata uint16_t stp_bpdu_age[10];	/* ticks since last BPDU seen on port (saturating) */
-__xdata uint8_t  stp_loop_held[10];	/* port is out of forwarding because a loop was seen on it */
-__xdata uint8_t  stp_tx_budget[10];	/* tx hold: BPDUs left in the current second */
-__xdata uint8_t  stp_tx_count[10];	/* BPDUs actually put on the wire, wraps at 256 */
+__xdata uint16_t port_timers[STP_ENTITIES];	/* listen-period countdown (0 = not listening) */
+__xdata uint16_t port_hello[STP_ENTITIES];	/* hello TX countdown */
+__xdata uint16_t stp_bpdu_age[STP_ENTITIES];	/* ticks since last BPDU seen on port (saturating) */
+__xdata uint8_t  stp_loop_held[STP_ENTITIES];	/* port is out of forwarding because a loop was seen on it */
+__xdata uint8_t  stp_tx_budget[STP_ENTITIES];	/* tx hold: BPDUs left in the current second */
+__xdata uint8_t  stp_tx_count[STP_ENTITIES];	/* BPDUs actually put on the wire, wraps at 256 */
 __xdata uint16_t stp_sec_tick;		/* 1 s window for the tx budget */
 __xdata uint16_t stp_link_prev;		/* carrier bitmap as of the last check */
 __xdata uint16_t stp_link_now;
@@ -164,7 +169,12 @@ struct stp_pkt_in {
 /* Console messages name the port on the front panel, not the internal index. */
 static void print_port_nl(uint8_t port) __reentrant
 {
-	print_byte(machine.log_to_phys_port[port]);
+	if (port >= STP_LAG_BASE) {
+		write_char('L');
+		write_char('1' + port - STP_LAG_BASE);
+	} else {
+		print_byte(machine.log_to_phys_port[port]);
+	}
 	write_char('\n');
 }
 
@@ -183,6 +193,8 @@ static void print_bridge_id(uint8_t prio, uint8_t ext, __xdata uint8_t *mac) __r
 static __code const char stp_state_txt[] = "off  blocklearnfwd  ";
 static __code const char stp_role_txt[]  = "desgroot";
 static __code const char stp_edge_txt[]  = "no  yes ";
+
+static uint8_t stp_ent_active(uint8_t e) __reentrant;
 
 static void print_field(__code const char *txt, uint8_t idx, uint8_t width) __reentrant
 {
@@ -207,7 +219,12 @@ static void stp_status(void)
 		print_string(" (this switch)\n");
 	} else {
 		print_string(" port ");
-		print_byte(machine.log_to_phys_port[stp_root_port]);
+		if (stp_root_port >= STP_LAG_BASE) {
+			write_char('L');
+			write_char('1' + stp_root_port - STP_LAG_BASE);
+		} else {
+			print_byte(machine.log_to_phys_port[stp_root_port]);
+		}
 		print_string(" cost ");
 		print_long(root_bridge_cost);
 		write_char('\n');
@@ -217,11 +234,26 @@ static void stp_status(void)
 	write_char('\n');
 	print_string("port state role edge tx bpdu\n");
 	reg_read_m(RTL837X_MSTP_STATES);
-	for (stp_i = machine.min_port; stp_i <= machine.max_port; stp_i++) {
+	for (stp_i = 0; stp_i < STP_ENTITIES; stp_i++) {
+		if (!stp_ent_active(stp_i))
+			continue;
 		write_char(' ');
-		print_byte(machine.log_to_phys_port[stp_i]);
+		if (stp_i >= STP_LAG_BASE) {
+			write_char('L');
+			write_char('1' + stp_i - STP_LAG_BASE);
+		} else {
+			print_byte(machine.log_to_phys_port[stp_i]);
+		}
 		print_string("  ");
-		print_field(stp_state_txt, (sfr_data[3 - (stp_i >> 2)] >> ((stp_i << 1) & 0x7)) & 0x3, 5);
+		stp_st_of = stp_i;
+		if (stp_i >= STP_LAG_BASE) {
+			stp_st_of = 0;
+			while (stp_st_of < 10 && !((stp_lag_mask[stp_i - STP_LAG_BASE] >> stp_st_of) & 1))
+				stp_st_of++;
+			if (stp_st_of >= 10)
+				stp_st_of = 0;
+		}
+		print_field(stp_state_txt, (sfr_data[3 - (stp_st_of >> 2)] >> ((stp_st_of << 1) & 0x7)) & 0x3, 5);
 		write_char(' ');
 		print_field(stp_role_txt, stp_i == stp_root_port ? 1 : 0, 4);
 		write_char(' ');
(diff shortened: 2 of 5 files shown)


logicog:
Good call to close #325. This looks much better. Still needs a bit of reviewing (saw a couple of hard-coded things like the CPU-port or the number of LAGs, already), I will do that and testing later today. Also need to think about how to test the LAG part... Maybe with 2 RTL switches? One can also set up LAGs in Linux, but that is really cumbersome, Want to make sure all lag members can be blocked.

DrDoof:
You don't need a second switch, or Linux, for the blocking part — one switch and one cable does it, and the check you want falls out of a register read.

Take two ports with nothing plugged into them and make a **static** LAG of them. No LACP needed, `lag 2 5 6` is enough: STP only cares that the group exists. Then run a single cable from one of those two members to an ordinary port on the same switch. That is a loop through the aggregate, entirely inside the switch, so nothing else on the network is involved and there is nothing to set up on a host.

Then read the MSTP state register, two bits per port from the LSB, `3` forwarding and `1` blocking:

```
regget 5310  ->  0x000ff577

  log 4 (phys 5)  BLOCK    <- lag member, this one has the cable
  log 5 (phys 6)  BLOCK    <- lag member, no cable in it at all
  log 2 (phys 3)  forward  <- the ordinary port, stays up
```

Both members are blocked in silicon with only one of them cabled, which is the thing you wanted to be sure of: the decision is taken once for the entity and written to every member, not to the port the BPDU happened to arrive on. `stp status` agrees (` L2  block`, ` 03  fwd`), but the register is the one I would trust, since it sits on the far side of the whole display path.

I also saved and rebooted with the loop still in place, so it exists from the first second rather than being introduced into an already converged tree. It latches immediately and held over 40 s, register `0x000ff5f7`. Traffic on the looped pair stayed at BPDU rate throughout, so the loop is extinguished rather than merely labelled.

On the hard-coded values, you are right on both. The port count is the bare `10` the file already used before this change, and it is really `CPU_PORT + 1` from `rtl837x_common.h`; `STP_LAG_BASE` is that same number written once more. The LAG count is the worse of the two — `4` appears here, as `group > 3` in `parse_lag`, and again in the LACP branch, with no shared define anywhere.

I would rather not move the code while you are reading it. Say when you are through and I will fold both into named constants, or tell me to keep it inside this file and leave `parse_lag` for a separate change.

doc/stp.md
```diff
@@ -58,6 +58,28 @@ port set to admit tagged frames only (`ingress <port>t`) never delivers one
 to the CPU. `stp_setup()` prints a warning for every STP-enabled port in that
 state.
 
+## Link aggregation
+
+A LAG is a port to the protocol, the way the vendor firmware presents it as
```
logicog: Suggestion, just write: A LAG (Link Aggregation Group) is handled like an additional port by the STP protocol with its own timers and states. The switch-devices support up to 4 LAGs which are shown alongside physical ports in the STP status.

I.e.: No reference to OEM firmware, which may change the name or have different names on different versions. Also, mixing trunk and LAG is kind of confusing.

doc/stp.md
```diff
@@ -58,6 +58,28 @@ port set to admit tagged frames only (`ingress <port>t`) never delivers one
...
+Trk1 to Trk4. The four groups are entities alongside the nine physical ports:
+they carry their own path cost, priority, edge and guard settings, they hold
+their own timers, and they appear as their own rows on the Spanning Tree page.
+
+```
+stp lag 1 cost 10000    # the group decides, not its members
+stp lag 1 edge off
+```
+
+Membership comes from the aggregation registers the `lag` command writes, re-read
```
logicog: Suggestion: LAGs are configure via the `lag` command. Once a port is configured as a member of a LAG, it can no longer be configured individually for STP, e.g. via the `stp port <n>` command. The switch hardware itself does not handle STP state for a LAG as a whole, instead STP-status changes for the members of a LAG must be handled by the firmware by updating all member ports, which is possible using a single register write.

html/stp.js
```diff
@@ -38,45 +38,58 @@ function num(id, min, max, onch) {
   return n;
 }
 
+function members(mask) {
+  const a = [];
+  for (let i = 0; i < 10; i++)
```
logicog: Hard-coded port 10?

logicog:
I tried creating a LAG via the web-interface, on a 4+2 device. I want to put both STP-ports into LAG 1 (I have one of those dirt-cheap Intel 82599ES dual SFP+ PCIe-cards on the Linux-PC side.). However, this does not work, there is no reaction on the web-page and on the console I see:
```
> Error: lag <1-4> [port]...

```
It looks like the web-page sends an incorrectly configured lag command. Indeed, Wireshark says, the command is `lag 0 5 6`, and 0 is not between 1 and 4.

logicog (commented):
Not yet through, as I was testing in parallel and got stuck, see above.

html/stp.js
```diff
@@ -38,45 +38,58 @@ function num(id, min, max, onch) {
...
   const tbl = document.getElementById("stpPortsTbl");
   const stat = document.getElementById("stpStatTbl");
   for (const p of [...ports].sort((a, b) => a.p - b.p)) {
     const tr = tbl.insertRow();
-    tr.insertCell().textContent = p.p;                     // Port
-    tr.insertCell().appendChild(sel("en_" + p.p,
+    p.n = p.lag ? "lag " + p.lag : "port " + p.p;
+    p.k = p.lag ? "L" + p.lag : p.p;
+    tr.insertCell().textContent = p.lag
+      ? "Trk" + p.lag + " (" + members(p.mbr) + ")"
```
logicog: LAG or Trunk? Usually a trunk is the central backbone of a network carrying multiple VLANs, and LAG is very specific about combining multiple Links/Ports. I would prefer to stay with LAG, which is also used by Linux and Windows. Also, the menu entry is called LAG, and not "Trunking".

html/stp.js
```diff
@@ -38,45 +38,58 @@ function num(id, min, max, onch) {
...
+      e => stpCmd("stp " + p.n + " guard " + e.target.value)));
+    tr.insertCell().appendChild(sel("p2p_" + p.k,
       [["auto","Auto"],["on","Enable"],["off","Disable"]],
-      e => stpCmd("stp port " + p.p + " p2p " + e.target.value)));
+      e => stpCmd("stp " + p.n + " p2p " + e.target.value)));
 
     const sr = stat.insertRow();
-    sr.insertCell().textContent = p.p;
+    sr.insertCell().textContent = p.lag
+      ? "Trk" + p.lag + " (" + members(p.mbr) + ")"
```
logicog: Trk?

logicog:
Created LAG 1 (2 in the web-interface, but later is correctly shown as 1), and then went to STP. This looks like this, now:
STP_lags
I then logged out and back in and got:
stp_lag2
which looks OK. (except for Trk).

logicog:
In the above image, Trk(4, 9) is confusing. These are the logical ports. Needs to be translated to LAG1(5, 6).

logicog:
Dissolving the lag works nicely, although the lag command documentation is incorrect, `show` is missing. And the result from `lag 1` is confusing, we should have `lag 1 delete/d`:
```
> stp status
STP on, RSTP
bridge  8000/061e04b94f1a
root    8000/061e04b94f1a (this switch)
changes 0x0000
port state role edge tx bpdu
 01  fwd   desg yes  36 255
 02  fwd   desg yes  36 255
 03  fwd   desg yes  36 255
 04  fwd   desg yes  36 255
 L1  fwd   desg yes  36 255 <<<<<<<<<<< LAG from Web.

> lag
Error: lag <1-4> [port]...

> lag 1
port_lag_members_set, lag: 00, members: 0x0000

> STP: link down, port blocking L1
STP: edge port forwarding 05
STP: edge port forwarding 06


> 
tcp: found no unused connections.

> lag 1 5 6
port_lag_members_set, lag: 00, members: 0x0108

> STP: edge port forwarding L1

> stp status
STP on, RSTP
bridge  8000/061e04b94f1a
root    8000/061e04b94f1a (this switch)
changes 0x0001
port state role edge tx bpdu
 01  fwd   desg yes  d5 255
 02  fwd   desg yes  d5 255
 03  block desg yes  d5 255
 04  block desg yes  d5 255
 L1  fwd   desg yes  5e 36

> lag show
LAG status:
 1 member ports: 5 6  (hash: 0x7e)
 2 disabled
 3 disabled
 4 disabled

> 

```

html/stp.js
```diff
@@ -38,45 +38,58 @@ function num(id, min, max, onch) {
   return n;
 }
 
+function members(mask) {
+  const a = [];
+  for (let i = 0; i < 10; i++)
+    if (mask & (1 << i)) a.push(i + 1);
```
logicog: Here, you need to push machine.log_to_phys_port[i]);

rtl837x_stp.c
```diff
@@ -266,26 +298,110 @@ int8_t cmpBytes(__xdata uint8_t *m1, __xdata uint8_t *m2, uint8_t n) __reentrant
...
+static void stp_lag_map(void)
+{
+	for (stp_ss_i = 0; stp_ss_i < 10; stp_ss_i++)
+		stp_ent_of[stp_ss_i] = stp_ss_i;
+	for (stp_scratch = 0; stp_scratch < STP_LAG_COUNT; stp_scratch++) {
+		stp_scratch16 = port_lag_members_get(stp_scratch);
+		if (stp_lag_mask[stp_scratch] != stp_scratch16)
+			stp_map_dirty = 1;
+		stp_lag_mask[stp_scratch] = stp_scratch16;
+		for (stp_ss_i = 0; stp_ss_i < 10; stp_ss_i++)
```
logicog: 10 hard-coded.

rtl837x_stp.c
```diff
@@ -217,11 +234,26 @@ static void stp_status(void)
...
+			print_byte(machine.log_to_phys_port[stp_i]);
+		}
 		print_string("  ");
-		print_field(stp_state_txt, (sfr_data[3 - (stp_i >> 2)] >> ((stp_i << 1) & 0x7)) & 0x3, 5);
+		stp_st_of = stp_i;
+		if (stp_i >= STP_LAG_BASE) {
+			stp_st_of = 0;
+			while (stp_st_of < 10 && !((stp_lag_mask[stp_i - STP_LAG_BASE] >> stp_st_of) & 1))
+				stp_st_of++;
+			if (stp_st_of >= 10)
```
logicog: 10 hardcoded.

rtl837x_stp.c
```diff
@@ -217,11 +234,26 @@ static void stp_status(void)
...
+			write_char('1' + stp_i - STP_LAG_BASE);
+		} else {
+			print_byte(machine.log_to_phys_port[stp_i]);
+		}
 		print_string("  ");
-		print_field(stp_state_txt, (sfr_data[3 - (stp_i >> 2)] >> ((stp_i << 1) & 0x7)) & 0x3, 5);
+		stp_st_of = stp_i;
+		if (stp_i >= STP_LAG_BASE) {
+			stp_st_of = 0;
+			while (stp_st_of < 10 && !((stp_lag_mask[stp_i - STP_LAG_BASE] >> stp_st_of) & 1))
```
logicog: 10 hardcoded.

rtl837x_stp.c
```diff
@@ -266,26 +298,110 @@ int8_t cmpBytes(__xdata uint8_t *m1, __xdata uint8_t *m2, uint8_t n) __reentrant
 }
 
 
-/* Write one port's 2-bit state into the ASIC's MSTP register.
- * 00 disable, 01 blocking, 10 learning, 11 forwarding. */
+static void stp_lag_map(void)
+{
+	for (stp_ss_i = 0; stp_ss_i < 10; stp_ss_i++)
```
logicog: 10 Hardcoded.

rtl837x_stp.c
```diff
@@ -266,26 +298,110 @@ int8_t cmpBytes(__xdata uint8_t *m1, __xdata uint8_t *m2, uint8_t n) __reentrant
...
+		return ent == p;
+	return (stp_lag_mask[ent - STP_LAG_BASE] >> p) & 1;
+}
+
+
+static uint8_t stp_state_port(uint8_t e) __reentrant
+{
+	if (e < STP_LAG_BASE)
+		return e;
+	for (stp_ss_i = 0; stp_ss_i < 10; stp_ss_i++)
```
logicog: 10 hardcoded.

rtl837x_stp.c
```diff
@@ -266,26 +298,110 @@ int8_t cmpBytes(__xdata uint8_t *m1, __xdata uint8_t *m2, uint8_t n) __reentrant
...
+		if ((stp_lag_mask[e - STP_LAG_BASE] >> stp_ss_i) & 1)
+			return stp_ss_i;
+	return 0;
+}
+
+
+static void stp_state_bits(uint8_t port, uint8_t state) __reentrant
+{
+	for (stp_ss_i = 0; stp_ss_i < 10; stp_ss_i++)
```
logicog: same here.

rtl837x_stp.c
```diff
@@ -266,26 +298,110 @@ int8_t cmpBytes(__xdata uint8_t *m1, __xdata uint8_t *m2, uint8_t n) __reentrant
...
+
+
+/* MSTP register encoding: 00 disable, 01 blocking, 10 learning, 11 forwarding. */
 static void stp_state_set(uint8_t port, uint8_t state) __reentrant
 {
 	reg_read_m(RTL837X_MSTP_STATES);
-	stp_scratch = 3 - (port >> 2);
-	sfr_data[stp_scratch] &= ~(uint8_t)(0b11 << ((port << 1) & 0x7));
-	sfr_data[stp_scratch] |= (uint8_t)(state << ((port << 1) & 0x7));
+	for (stp_ss_i = 0; stp_ss_i < 10; stp_ss_i++) {
```
logicog: 10...

rtl837x_stp.c
```diff
@@ -348,7 +464,18 @@ void stp_cnf_send(uint8_t port) __reentrant
 	STP_O->rtl_tag.version = RTL_FRAME_TAG_VERSION;
 	STP_O->rtl_tag.reason = 0x00;
 	STP_O->rtl_tag.flags = HTONS(RTL_TAG_LEARN_DIS);
-	STP_O->rtl_tag.pmask = HTONS(((uint16_t)1) << port);
+	if (port >= STP_LAG_BASE) {
+		stp_scratch = 0;
+		while (stp_scratch < 10 && !stp_ent_has(port, stp_scratch))
```
logicog: 10...

rtl837x_stp.c
```diff
@@ -348,7 +464,18 @@ void stp_cnf_send(uint8_t port) __reentrant
...
+			stp_scratch++;
+		if (stp_scratch >= 10) {
```
logicog: 10...

rtl837x_stp.c
```diff
@@ -714,25 +886,34 @@ static void stp_fdb_update(__xdata uint16_t pmask)
...
+	for (stp_i = 0; stp_i < STP_ENTITIES; stp_i++) {
 		stp_pflags[stp_i] &= ~(STP_PF_OPEREDGE | STP_PF_TRIPPED);
 		stp_loop_held[stp_i] = 0;
 		stp_bpdu_age[stp_i] = 0;
 		stp_tx_budget[stp_i] = stp_txhold;
 		stp_tx_count[stp_i] = 0;
-		if (!(stp_pflags[stp_i] & STP_PF_ENABLED) || (stp_pflags[stp_i] & STP_PF_ADMEDGE)) {
+		port_timers[stp_i] = 0;
+		port_hello[stp_i] = (uint16_t)stp_hello_s * STP_HZ;
+		if (stp_i < 10 && (stp_i < machine.min_port || stp_i > machine.max_port))
```
logicog: 10...

rtl837x_stp.c
```diff
@@ -667,7 +834,12 @@ void stp_defaults(void) __banked
 	stp_fwddelay_s = 15;
 	stp_rstp = 1;
 	stp_txhold = 6;
-	for (stp_i = 0; stp_i < 10; stp_i++) {
+	for (stp_i = 0; stp_i < 10; stp_i++)
```
logicog: 10...

logicog (commented):
Just details.

logicog:
> Then read the MSTP state register, two bits per port from the LSB, `3` forwarding and `1` blocking:

I want to test this through with practical examples. It is often surprising what kind of errors and "undocumented hardware features" one finds when actually doing real-life tests. I also do lots of other driver development and skipping this step can become very embarrassing when users start using your code. Just reading the register, we could overlook that other registers are also necessary to be correctly set. Or one would need to first disable a port before taking it out of a LAG or other funny things one would not expect. BTW, the most thorough tester I have ever met is @vDorst ;-)

DrDoof:
You are right about the testing, and this is a good demonstration of why. My register read was true and it was true about one board. `regget 5310` tells you what the silicon decided, which is the part I most wanted evidence for, but it says nothing about the layer above it, and that is exactly where you found the bug. On the switch I have, logical and physical numbering differ by one, so the members list came out right by accident and would have kept coming out right forever.

Fixed in `b802f6a`: the list goes through `logToPhysPort` now, the same map the LAG page, the mirror page and the L2 table already use. Your `Trk (4, 9)` is mask `0x0108`, so with the translation it reads `LAG1 (5, 6)`. The label says LAG rather than Trk while I was in there, since that is what the rest of the tree calls it.

The other three you hit are not from this branch, so I put them in #399 rather than widen this one:

* the web page cannot create the first group at all, because the buttons pass a 0-based index and the parser wants 1 to 4 — that is the `lag 0 5 6` Wireshark showed you, and why the page did nothing
* `show` missing from the usage line
* a bare `lag 1` wiping the group, now `lag 1 delete` or `lag 1 d`

I have not clicked through the GUI half of #399 myself for the same reason as above: on my board the numbering bug is invisible. It builds and `machine_check` is clean, but it wants your hardware.

DrDoof:
Thanks, all fifteen of them are gone. The 10 is the width of the port field
in the MSTP register, so it is CPU_PORT + 1 now and has a name. The assembly
comes out unchanged apart from the version string, so it really is just the
rename.

stp.js counted to ten as well when listing members; that uses numPorts now.

For the doc I took your wording almost verbatim: no OEM firmware, no trunk,
and the point about the hardware keeping no per-group state while the
firmware writes all the members in one register write. I kept the LACP
re-read and the BPDU going out through the lowest member, since neither is
obvious from the code.

hello and txhold also say 10, but those are seconds, so they stayed.

DrDoof:
Rebased on current main after #315. The stp.js and config.js changes moved into main.js, since the web UI is a single bundle now.

logicog:
I am wondering a bit about the use of  `__reentrant`: You use it for e.g `stp_ent_active(), spt_ent_has()` or `stp_ent_port()` At least the first two ones do not seem to need O/DSEG for variable hand-off, since the single u8 parameter should be handed over in the A-register by the compiler, the rest of the functions are just logic operations on the variable, so should not need much more register use. Is the  `__reentrant`, which uses up stack, really necessary? In the Linux kernel, functions of this kind would either be defined as a macro or expected to be inlined by the compiler. Unfortunately, sdcc does not do that. There is a certain inflation of `__reentrant` in the code and I am wondering whether it will not bite us later because of putting too much on the stack.

rtl837x_stp.c
```diff
@@ -714,25 +886,34 @@ static void stp_fdb_update(__xdata uint16_t pmask)
...
+			continue;
+		if (stp_i < STP_PORTS && stp_ent_of[stp_i] != stp_i)
+			continue;
+		if (!(stp_pflags[stp_i] & STP_PF_ENABLED)
+		    || (stp_pflags[stp_i] & STP_PF_ADMEDGE)) {
 			/* not participating, or admin edge: forwarding immediately */
 			if (stp_pflags[stp_i] & STP_PF_ADMEDGE)
 				stp_pflags[stp_i] |= STP_PF_OPEREDGE;
-			sfr_data[3 - (stp_i >> 2)] |= (uint8_t)(0b11 << ((stp_i << 1) & 0x7));
+			stp_state_bits(stp_i, 0b11);
```
logicog: Maybe use a define to name the state bits 0b11?

rtl837x_stp.c
```diff
@@ -714,25 +886,34 @@ static void stp_fdb_update(__xdata uint16_t pmask)
...
 			port_timers[stp_i] = 0;
 		} else {
 			/* listen first: blocking until the forward-delay expires */
-			sfr_data[3 - (stp_i >> 2)] |= (uint8_t)(0b01 << ((stp_i << 1) & 0x7));
+			stp_state_bits(stp_i, 0b01);
```
logicog: define?

rtl837x_stp.c
```diff
@@ -803,12 +990,27 @@ void stp_parse(void) __banked __reentrant
...
 			goto err;
-		if (!cmd_parse_port_separator(cmd_words_b[2]))
-			goto err;
-		port = atoi_results_u8;
+		if (cmd_compare(1, "lag")) {
+			if (!atoi_byte(cmd_words_b[2]))
+				goto err;
+			if (!atoi_results_u8 || atoi_results_u8 > STP_LAG_COUNT)
+				goto err;
+			port = STP_LAG_BASE + atoi_results_u8 - 1;
```
logicog: Is the port not a physical port and the STP_LAG_BASE a logical port equivalent?

logicog (commented):
Some observations.

rtl837x_stp.c
DrDoof: Both done, thanks.

The state bits have names now. There were eighteen of these literals in the file, sixteen of them older than this PR, so I converted the lot rather than leave it written half one way and half the other. Naming them turned up something the literals were hiding: 0b11 appears in two roles, once as the forwarding state and once as the two bit field mask, so those became STP_ST_FORWARDING and STP_ST_MASK. The comment above stp_state_set() went with them, since it said exactly what the defines say.

On the port question: it is already a logical index by the time it gets here. cmd_parse_port() does the translation before it returns:

```c
uint8_t port = cmd_buffer[idx] - '0' - 1;
port = machine.phys_to_log_port[port];
atoi_results_u8 = port;
```

STP_LAG_BASE is STP_PORTS, which is CPU_PORT + 1, so the aggregates sit just above the CPU port in that same logical space and the state arrays are all STP_ENTITIES long. The way back out is symmetric: log_to_phys_port[] for a port, '1' + i - STP_LAG_BASE for a group.

That said, you had to go two functions deep to answer your own question, which is fair comment on the code. The variable holds an entity index rather than a port, so it is called ent now. Same generated code either way, 5331 instructions before and after.

logicog:
Thanks for fixing all the issues.

PR #399, opened 2026-09-05
lag: fix the web page's group numbering, and the command's syntax line

DrDoof:
Three small things @logicog ran into while testing #397 on a 4+2 board. None of them come from that branch, so they are here instead.

**The web page cannot create the first group.** The LAG page draws four sections headed LAG 1 to LAG 4 and wires their buttons to `lagSub(0)` through `lagSub(3)`, then sends that index straight through as the group number. The first button therefore asks for `lag 0`, the parser rejects it, and nothing appears on the page because the error only reaches the console. The rest are off by one, so the section headed LAG 2 is the one that creates group 1. The index stays 0-based for the element ids; only the command gets the 1-based number.

**The usage line leaves out `show`.** `lag show` is the only way to see the groups, and it is the one subcommand the error message does not mention.

**A group number with no ports silently wiped the group.** Worth having, but not one fat-fingered return away from a bare `lag 1`. It takes `lag 1 delete` or `lag 1 d` now, and a bare group number prints the usage.

Nothing changes for a saved configuration: `lag 1 5 6` is still what gets written and replayed, and neither the old bare form nor the new delete matches the pattern that decides what is persisted.

Builds for SWTGW218AS and `machine_check` is clean. The GUI half I have not clicked through myself, since the board I have puts logical and physical numbering one apart and the bug is invisible there; @logicog, you have the hardware that shows it.

### DIFF
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -406,10 +406,16 @@ void parse_lag(void)
 	if (group > 3)
 		goto err;
 
+	if (cmd_words_len == 2)
+		goto err;
+
+	if (cmd_compare(2, "delete") || cmd_compare(2, "d")) {
+		port_lag_members_set(group, 0);
+		return;
+	}
+
 	uint8_t w = 2;
 	while (w < cmd_words_len) {
-//		write_char('|'); print_byte(w); write_char(':'); write_char(cmd_buffer[cmd_words_b[w]]); write_char('-');
-
 		// Parse port.
 		if (cmd_parse_port_separator(cmd_words_b[w++]) == 0)
 			goto err;
@@ -419,7 +425,7 @@ void parse_lag(void)
 	port_lag_members_set(group, members);
 	return;
 err:
-	print_string("Error: lag <1-4> [port]...\n");
+	print_string("Error: lag (show | <1-4> (delete | <port>...))\n");
 }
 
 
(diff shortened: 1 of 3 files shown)


cmd_parser.c
```diff
@@ -406,10 +406,16 @@ void parse_lag(void)
 	if (group > 3)
 		goto err;
 
+	if (cmd_words_len == 2)
+		goto err;
```
vDorst: This can be merge with line 398, we already the number of arguments.

cmd_parser.c
```diff
@@ -406,10 +406,16 @@ void parse_lag(void)
...
+
+	if (cmd_compare(2, "delete") || cmd_compare(2, "d")) {
+		port_lag_members_set(group, 0);
+		return;
+	}
```
vDorst: I would like only "d"-variant. Short and simple.
Maybe update the docs.

cmd_parser.c
DrDoof: Done. The bare `lag <n>` is caught by the argument-count check now (< 3), so the separate == 2 is gone.

cmd_parser.c
DrDoof: Dropped `delete`, it is just `d` now. Docs updated too (the serial-console section).

vDorst:
Thanks

PR #400, opened 2026-09-05
l2: name the group a learned address sits behind

DrDoof:
The L2 table shows a device that sits behind an aggregate on whichever member the last frame happened to arrive on, and the hash keeps moving that around. The same address is on port 7 in one refresh and on port 8 in the next. Nothing on the page or the console says those two ports are one link, so what you are looking at reads as an address flapping between ports rather than a LAG doing its job.

Watching a single entry by its table index, on a SWTGW218AS with a two port group:

```
idx=0744  c0:74:2b:fb:16:bc  port 7  ...25 s later...  port 6
```

Same index, same address, same VLAN.

I went looking for a trunk bit in the entry first, because taking this from the silicon would be better than correlating it in software. The three words the firmware reads carry a plain member port and nothing that separates a LAG member from an ordinary port, and across 180 entries the port field never goes above 9, so there is no group id tucked in there either. Whether the chip could be told to learn against the trunk instead I could not find out: our register map describes one of the 32 bits in `L2_CTRL`, so I can neither rule it in nor rule it out.

That leaves doing it where the masks already live. `port_lag_of()` is already there and `port_isolate()` already uses it, so both views ask it instead of growing a second copy of the same formula. The console prints `LAG1` where it used to print the member, and `/l2.json` gains a `"lag"` field next to the port, zero when the port is on its own, which is the shape `/stp.json` already uses. The L2 page reads that field for the port column, and for the filter and the sort as well: showing `LAG1` while a search for `7` still pulled half of one device's addresses out of the group would have been worse than leaving it alone. Groups sort after the physical ports, CPU stays last.

The port stays in the JSON deliberately. Which member a frame came in on is genuinely useful when you are chasing a miscabled link, so the page can show the group without the firmware having thrown the detail away.

On hardware, with the group up:

```
c0:74:2b:fb:16:bc   0x001e   learned   LAG1
c0:74:2b:fb:16:bd   0x0002   learned   LAG1
dc:a6:32:a3:53:af   0x0002   learned   9      <- not in a group
```

The page shows the same thing, and a device that had been moving between two ports now stays on one row.

151 bytes of BANK1, which leaves 1139 free.

### DIFF
--- a/html/l2.js
+++ b/html/l2.js
@@ -68,7 +68,11 @@ var l2SortCol = 'port';
 var l2SortDir = 1;
 
 function l2Key(e, col) {
-  if (col === 'port') return e.port === 'CPU' ? Number.MAX_SAFE_INTEGER : Number(e.port);
+  if (col === 'port') {
+    if (e.port === 'CPU') return Number.MAX_SAFE_INTEGER;
+    if (e.lag) return 100 + e.lag;
+    return Number(e.port);
+  }
   if (col === 'vlan') return Number(e.vlan);
   return String(e[col]).toLowerCase();
 }
@@ -114,7 +118,11 @@ function fillL2(s)
     return;
   s.sort(l2CMP);
   s = uniq(s);
-  l2All = s;
+  l2All = s.map(function(e) {
+    if (e.lag)
+      e.port = 'LAG' + e.lag;
+    return e;
+  });
   renderL2();
 }
 
--- a/httpd/page_impl.c
+++ b/httpd/page_impl.c
@@ -359,6 +359,7 @@ void send_l2(uint16_t idx)
 	while (1) {
 		entries_left--;
 		uint8_t port = 0;
+		uint8_t lag;
 		reg_read_m(RTL837x_TBL_DATA_0);
 		REG_WRITE(RTL837x_TBL_DATA_0, sfr_data[0], sfr_data[1] & 0xfc, sfr_data[2] | (TBL_LUTREAD_NEXT_L2UC << 6), sfr_data[3]);
 
@@ -402,6 +403,9 @@ void send_l2(uint16_t idx)
 
 			port |= (sfr_data[3] & 0x3) << 2;
 			itoa_html(port);
+			slen += strtox(outbuf + slen, ",\"lag\":");
+			lag = port_lag_of(port);
+			itoa_html(lag == PORT_LAG_NONE ? 0 : lag + 1);
 		}
 
 		// Index
--- a/rtl837x_port.c
+++ b/rtl837x_port.c
@@ -381,6 +381,7 @@ void port_l2_learned(void) __banked
 
 	while (1) {
 		uint8_t port = 0;
+		uint8_t lag;
 		reg_read_m(RTL837x_TBL_DATA_0);
 		REG_WRITE(RTL837x_TBL_DATA_0, sfr_data[0], sfr_data[1],sfr_data[2] | 0xc0, sfr_data[3]);
 
@@ -422,7 +423,13 @@ void port_l2_learned(void) __banked
 				print_string("\tlearned\t");
 
 			port |= (sfr_data[3] & 0x3) << 2;
-			print_phys_port(port);
+			lag = port_lag_of(port);
+			if (lag == PORT_LAG_NONE) {
+				print_phys_port(port);
+			} else {
+				print_string("LAG");
+				itoa(lag + 1);
+			}
 		}
 
 		entry++;


logicog:
Interesting, that the learned port constantly switches between the ports included in a LAG. That is not what I would expect. I would have expected both ports to have learned the address, or there being an entry for the LAG as an entity. Which link a packet takes is supposed to depend only on the way the LAG has been set up, e.g. a hash based on LAG_HASH_L2_DMAC. But if it depends on the destination MAC hash, how is it supposed to work when the link does not have that same MAC learned for the selected port? Very strange...

I went through the code, btw. It looks great. Need to test...

DrDoof:
Same reaction here, which is why I went and looked before proposing anything.

The table has no way to hold both. Every (MAC, VLAN) pair has exactly one index — I checked the whole table and none of them turns up twice — and what we were watching was `idx=0744` keeping its identity while the port field underneath it changed. One entry with a port that moves, not two entries competing.

The thing moving it is the far end's hashing rather than ours. We learn from received frames, so the entry follows whichever member the bond last transmitted on.

The part that bothered you has a plainer answer than it looks. The entry's port is not what egress uses: six parallel flows to that same destination MAC left the switch two on one member and four on the other. If the learned port were the egress port, all six would have gone out one link. The destination resolves to the group, the hash picks the member, and the learned port only says which group — any member would point at the same one.

Your `LAG_HASH_L2_DMAC` example is worth keeping separate though, because with DMAC alone your question gets a lot sharper. This group runs the default mask, `0x7e`: SMAC, DMAC, SIP, DIP and both L4 ports. That is the only reason traffic to a single MAC spreads across the members at all.

I did go looking for the entity entry you expected, and did not find one. In the three words the firmware reads a LAG member's entry is bit-identical to an ordinary port's outside the port field, and across 180 entries the port field never goes above 9, so nothing is tucked in there. Whether the chip could be told to learn against the trunk instead I cannot say: our register map describes one of the 32 bits in `L2_CTRL`.

Thanks for reading it through.

logicog:
> The table has no way to hold both. Every (MAC, VLAN) pair has exactly one index — I checked the whole table and none of them turns up twice — and what we were watching was `idx=0744` keeping its identity while the port field underneath it changed. One entry with a port that moves, not two entries competing.

Have you actually tried that? My understanding is that the hash table also has buckets in order to overcome hash collisions. So, in theory, it could hold more than one entry with the same VLAN/MAC pair. Did you try putting more than one such entry in and look what can be read back?

DrDoof:
Fair, and you're right to split those two. I checked that the table never holds the same pair twice as it learns. That says nothing about whether the hardware would take it if something put it there, and I wrote the second as if I'd shown the first.

Your bucket model holds up, though. All 39 entries off a live switch, index split as `bucket * 4 + way`:

```
way 0: 36    way 1: 3    way 2: 0    way 3: 0

0x005d -> 0174 1e:2a:a3:1a:72:44   0175 52:54:00:fe:6c:cb
0x0154 -> 0550 dc:a6:32:d0:3e:35   0551 64:e8:33:60:23:ac
0x03d9 -> 0f64 c4:62:37:02:5b:73   0f65 50:02:91:1a:66:09
```

Everything lands in way 0 until something is already there. Split it the other way round, with the way in the high bits, and you get 11/11/9/8, which is what a hash looks like rather than a way selector. So 1024 buckets, 4 ways.

All three pairs are different addresses, which is the collision case you'd expect and the only one I've seen.

What I haven't done is your actual test: force the same MAC and VLAN into two ways and see what a lookup by key gives back. There's no route to it from the CLI, since `port_l2_static_mgmt()` writes with method 0 and the hardware picks the slot. It needs a build with an explicit index write. Say the word and I'll do it.

logicog:
> There's no route to it from the CLI, since `port_l2_static_mgmt()` writes with method 0 and the hardware picks the slot. It needs a build with an explicit index write. Say the word and I'll do it.

I do not think that we need this in this project, there are very limited applications and lots of ways to break things. This would be more for you yourself to test.

Anyway, I have had my moment where I finally understand that this may not be such a big problem. So, I set up a LAG on linux in different network namespaces:
```
$ sudo ip netns add client
$ sudo ip netns add server
$ sudo ip link set enp1s0f0 netns client    <<<<<<<<<<< This is a 10GBit fiber
$ sudo ip link set enp2s0 netns server      <<<<<< 2.5GBit
$ sudo ip link set enx6c6e0709ffff netns server << +2.5GBit
$ sudo ip netns exec server ip link set enp2s0 master bond0
$ sudo ip netns exec server ip link set enx6c6e0709ffff master bond0
$ sudo ip netns exec server ip link set bond0 up
$ sudo ip netns exec server ip addr add 192.168.5.1/24 dev bond0
$ sudo ip netns exec client ip addr add 192.168.5.2/24 dev enp1s0f0
$ sudo ip netns exec client ip link set enp1s0f0 up
$ sudo ip netns exec server iperf3 -s
```
The bonding is 0, round-robin:
```
$ sudo ip netns exec server cat /sys/class/net/bond0/bonding/mode
balance-rr 0
```
On the RTL-switch, the 2x2.5GBit end in ports 3 and 4 (port 1 is used for monitoring the switch):
```
> lag 1  3  4
```
And lo-and-behold, 2 parallel iperf3 streams on different(!) ports make 5GBit total throughput. We need the different ports because the port number is part of the hash being used to distribute the packets on the switch.
``` 
l2

24:5e:be:50:cb:98       0x0001  learned LAG1
06:1e:04:b9:4f:1a       0x0001  learned CPU
28:41:c6:fb:xx:xx       0x0001  learned 6
28:41:c6:fb:xx:xx       0x0001  learned 5
9c:69:d3:80:xx:xx       0x0001  learned 1
```
But this still looks strange: there is only one MAC in the LAG1. Until you notice that linux assigned a new MAC to both links of the LAG:
```
$ sudo ip netns exec server ip a
1: lo: <LOOPBACK> mtu 65536 qdisc noop state DOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
2: bond0: <BROADCAST,MULTICAST,MASTER,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default qlen 1000
    link/ether 24:5e:be:50:cb:98 brd ff:ff:ff:ff:ff:ff
    inet 192.168.5.1/24 scope global bond0
       valid_lft forever preferred_lft forever
    inet6 fe80::265e:beff:fe50:cb98/64 scope link proto kernel_ll 
       valid_lft forever preferred_lft forever
3: enp2s0: <BROADCAST,MULTICAST,SLAVE,UP,LOWER_UP> mtu 1500 qdisc mq master bond0 state UP group default qlen 1000
    link/ether 24:5e:be:50:cb:98 brd ff:ff:ff:ff:ff:ff
    altname enx245ebe50cb98
11: enx6c6e0709ffff: <BROADCAST,MULTICAST,SLAVE,UP,LOWER_UP> mtu 1500 qdisc fq_codel master bond0 state UP group default qlen 1000
    link/ether 24:5e:be:50:cb:98 brd ff:ff:ff:ff:ff:ff
    altname enp12s0u1
```
So, at least for the LAG-link to an edge device itself, everything is OK, we need only 1 L2 entry. Still, if there were other devices in the network that are not directly connected to the switch, l2 would only allow one of the links to be used, while in theory the hash should be better in distributing the load...

I tested the CLI and Web and things look great. I think it is time to merge this, even if LAGs (and LACP!) probably need to be re-visited.

logicog:
Thanks for contributing!

PR #401, opened 2026-09-05
ports: read the right link state in the port table

DrDoof:
`createPortTable()` read `pState[i]` where the array is filled by `portNum - 1`, so each row started out showing the next port's link speed until the first refresh corrected it. On the last copper port that meant showing the uplink's speed.

Running the real `status.json` of a nine port switch through both versions, for the eight copper rows the table draws:

```
expected  7  7  7  1  7  1  7  7
before    7  7  1  7  1  7  7  6
after     7  7  7  1  7  1  7  7
```

The debug line above it goes too, since it printed a third index again (`pState[i-2]`).

This originally also gave the SFP port a row of its own. Per review that half is dropped: leaving modules out is deliberate for now, and being able to force a module's speed belongs in a feature of its own rather than arriving inside a fix.

### DIFF
--- a/html/ports.js
+++ b/html/ports.js
@@ -14,15 +14,16 @@ function createPortTable() {
        + '</select>';
       const dSwitch = '<input type="checkbox" id="disable_port" onchange="portOnOff();">'
      for (let i = 1; i <= numPorts; i++) {
-      if (pIsSFP[i-1])
-        continue;
-      console.log("Table row: " + i + "pState: " + pState[i-2]);
       const tr = tbl.insertRow();
       let td = tr.insertCell(); td.appendChild(document.createTextNode(t('common_port') + i));
       let portName = portNames[physToLogPort[i-1]] || '';
       td = tr.insertCell(); td.appendChild(document.createTextNode(portName));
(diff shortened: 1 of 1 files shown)


vDorst:
> The table skipped SFP ports entirely, so the uplink had no row at all, while the MTU table on the same page has always drawn it. Only the Set Speed column is meaningless for a module, so the port now gets its row with the name, link speed and connected devices, and the three control cells stay empty. Whether speed and disable ought to work for a module is a separate question I have not tested, so this does not offer a button for it.

I think this was deliberate for now to exclude SFP-port settings.

On a 4x2-sfp device both port don't show up, even with this patch.

Screenshot_20260906_150247

I would like to see that you can force the speed of a SFP-port eventually.
But as a feature, not as a bug fix.
But for now I say leave it as it is, I want give #315 priority.

So keep only the `Separately, createPortTable() read pState[i]` part.

DrDoof:
Fair enough, dropped it (30547cc). What's left is just the pState[i] fix, plus the debug line above it that printed yet another index. The SFP row can come back with the speed control behind it, as a feature, after #315.

vDorst:
Thanks

Issue #336, opened 2026-08-14
Interest in a rewritten web UI? (single-page, themed)


bloqaudio:
I have been running a rewritten web UI on my SWTGW218AS units since mid July and before I invest in preparing it for upstream I want to ask whether a replacement UI is something you would consider at all.

What it is: a single-page app, one page plus a couple of scripts, replacing the per-page HTML files. It talks to the same endpoints the current pages use, so the firmware side is untouched. It brings four themes plus a follow-the-OS mode, a live dashboard (port grid, packets per second per port), a console tab that shows command output, dirty-config tracking with an explicit save-to-flash, a startup-config editor that highlights lines the parser will not understand before writing, and the per-port connected devices display I just sent as #335. The assets are split to stay inside the existing per-file size limits and the whole thing fits the 512 KB image.

Several pieces of it have already come to you as separate PRs, merged or open (hostname support, upload verdicts in #322, the connected devices column in #335), and #329 covers the same console-output ground my version has. What remains is the UI layer itself, which cannot be split into small PRs since it is one artifact.

Screenshots below: the dashboard in the four themes, and the System page. If this is not a direction you want for the project, no hard feelings, it simply stays in my fork.

<img width="2362" height="1047" alt="Image" src="https://github.com/user-attachments/assets/4e6269cd-d0f3-49a0-931f-21abe4ed8efb" />
<img width="2362" height="1047" alt="Image" src="https://github.com/user-attachments/assets/f69b75ad-aa2a-4045-928e-3c9c696c8dfd" />
<img width="2362" height="1047" alt="Image" src="https://github.com/user-attachments/assets/6b7fb756-ba90-4c02-9aef-830ebc68a8db" />
<img width="2362" height="1047" alt="Image" src="https://github.com/user-attachments/assets/71c7fd74-1349-492a-83ef-5328346fde1c" />
<img width="2362" height="1047" alt="Image" src="https://github.com/user-attachments/assets/02481634-1ca1-4579-821c-80949d1b58ac" />
<img width="2362" height="1047" alt="Image" src="https://github.com/user-attachments/assets/e2123a79-2262-40e9-9b5e-8ce0176e724a" />
<img width="2362" height="1047" alt="Image" src="https://github.com/user-attachments/assets/985a32f9-eec3-46da-aacd-e4a619dc4e09" />

vDorst:
Yes I am!

But I think it needs a bit more coordination to make it happen.
- I think first want #315 ready and merged.
[...]

PR #403, opened 2026-09-06
User interface reimagined

piotrkochan:
This awesome project deserves a better UI.

Totally vibe-coded, and it definitely needs a thorough review. Some changes probably go a bit beyond a UI overhaul. I’ll keep refining things in my spare time, but I’m at the point where community feedback and testing would really help—especially around compatibility with other models. On my SWTGW218AS 8+1 Managed Switch, everything seems to work correctly so far.

The frontend uses Lit, TypeScript and Vite, with a lightweight runtime and no external fonts or CDN dependencies. Production assets are bundled and minified to save flash space. The current UI uses 162,411 bytes, leaving 30,101 bytes (~29.4 KiB) available in the reserved UI flash region.

Alongside the redesign, this adds optional automatic saving after applying changes, configurable automatic sign-out, session renewal without losing drafts, and warnings before discarding unsaved edits. Requests are serialized to accommodate the switch’s single HTTP connection.

The PR also integrates frontend compilation into the existing firmware build and adds tools for memory checks, device diagnostics and remote firmware uploads. That adds some scope, but I think it’s worthwhile. The firmware’s C code remains unchanged; the new behavior uses the existing API.

image
image


vDorst:
Thanks for this PR.

For now #315 get highest priority.
This compresses all the current web-assets with gzip which reduces the size a lot and to load times of the website.
Every other PR which makes changes to the web-assets `html/*` need adapt the PR to #315.
#315 will be merge in a few days.

Currently there is already a PR #336 to replace the website, #315 is related to that PR.
Also a previous demo is shown in #220.

So I strongly advice to see if you can collaborate with the other contributors to get one solution.

vDorst:
I am going to close this, #429 is merged.

Issue #196, opened 2026-04-09
Port isolation


namezys-at-getpatch:
I want to understand if it's possible to make a port isolation without VLAN.

Assume that I have 2 devices connected to 2.5Gb (AP1 and modem for example) and switch connected to SFP+.

AP1 and modem should not be able to see each other directly but through switch on SFP+.

On other words, if AP1 send a package to MAC of modem, it should go to SFP+ and SFP+ will send it back.

Therefore, if package come from SFP+, it should be redirected according mac table.
However, if package come from any other port, it should be redirected to SFP+.

I expected that stock firmware with VLAN switch works in this way. 
Unfortunetely, according my observation, it just drop package if it should be forwarded from one port to another.

logicog:
I just had to look into the code to check, whether port isolation was implemented in the code, because I was not sure. The hardware has support for this, of course, however only the API for configuration is implemented at this point, none of the CLI or web-interface commands, which would allow you to configure it during runtime.

There are probably 2 reasons: the first is that most of the professionals prefer VLANs over port isolation for what you want to achieve: with VLANs you still have a single net, that you can monitor and control from one point in that network. The second reason is that it is so simple to implement in theory, but quite a lot of boring (for embedded developers) work to implement e.g. the web-page for configuration.

The first step would be to implement the CLI command, which can also be put into the config.txt,  that can be compiled in to configure the router during flashing. Would that be already helpful to you? This can be done in a couple of hours (which I or someone else will need to find).

namezys:
Fair enough.
My naive aproach was to put AP and modem to different VLAN. However, my Mikrotik CRS226-24G-2S+RM doesn't support join VLANs to one broadcast domain. Because it doesn't have L3 routing in ASIC, I have to put AP and modem to one boradcast domain.

You are right, it needs time which even I don't have. But because it is needed for me, the best to do it by myself.
I hope that I can do it (even I know C++ better than C).
Maybe you can give hints which API functions should I use and I will try to dig deeper.

Long story short:
I want to filter out some packages from modem using ACL om Mikrotik and it works well if I sit behaind it.
However, AP and modem on same switch and are trying to send packages directly (which is expected).
I want to force packages to go through modem.

I tryed to check chip PDF but was not found something similar to it.

logicog:
Can you try out PR #197 I added support for configuring port isolation on the CLI. I hope this is what you need. If you need web-configuration, this will need a bit more time.

namezys:
[...]

PR #405, opened 2026-09-06
isolate: configure port isolation from the web interface

DrDoof:
This adds the web page for port isolation. The firmware side has been there
since #197, but until now you needed the serial console to use it.

#196 asked for access ports that can reach the uplink but not each other.
That is one row per access port here, with only the uplink and the CPU
ticked, and no VLANs involved.

The grid reads a new /isolation.json and writes through the existing isolate
command, so there is still just the one way into the hardware. Only rows
whose mask actually changed are sent.

Tested on an SWTGW218AS:

- set an asymmetric mask from the console, the page showed the same ports
- ticked and unticked in the page, then read the masks back over the CLI
- Allow all puts every port back to 0x03ff
- an edit that has not been applied yet is not lost when the page refreshes

413 bytes of code, and nothing changed in the isolation logic itself.


vDorst:
Thanks for this feature, but I want to hold all changes, that touches the web-files, back until #315 is merge.

DrDoof:
Rebased on current main after #315. The page moved into index.html and main.js, since the web UI is a single bundle now.

Issue #404, opened 2026-09-06
Spontaneous management-plane wedge after hours of normal operation (no flash write involved)


gnacho:
**Context**
- keepLINK KP-9000-9XHML-X V3.1 (RTL8373), running a local fork of RTLPlayground.
- Base of the image: upstream `e266938` (includes the bounded ASIC/NIC wait work and the uIP MSS cap), plus a small custom UDP beacon (status every 30 s, FDB walk every 300 s) and an optional SNMP agent that is compiled out. The image runs close to the 8051 RAM limits.

**Observation**
Since early September the management plane of one switch has wedged spontaneously, hours after boot, with no flash write and no console command involved:

- ICMP, the web UI and the UDP beacon stop at the same time (beacon stops mid-interval, then silence).
- The ASIC keeps forwarding: clients behind the switch stay connected and reachable throughout.
- Nothing recovers it but a power cycle; it then boots and runs normally again (in one earlier case even a power cycle did not recover it and we re-flashed).
- We never had the serial console attached during these events, so we have no console output; we can only describe the network-visible symptoms.

This happened on two different builds, both with the bounded-wait work included. It is NOT the #367 pattern: no "Save Settings to Flash" happened anywhere near these events.

**What we tried / current mitigation**
Not knowing the cause, we added a software deadman as a safety net (no hardware watchdog appears to exist, as noted in #367):

- The 200 Hz system-tick ISR resets the chip (register 0x0024, same as the software reset) when `idle()` has not run for 10 s; `idle()` clears the counter each pass.
- Branch: https://github.com/gnacho/RTLPlayground/tree/netpulse/deadman-watchdog (one commit, ~21 lines, only `rtlplayground.c`).

It is running on the affected switch and has not mis-fired so far, but we do not know whether it will actually catch this particular wedge (if the fault also freezes interrupts it will not), and we are not claiming it is the right fix.

**Open questions (we are unsure about all of this)**
1. Has anyone seen a spontaneous wedge with no trigger like this on a build that already has the bounded waits?
2. Could it be related to running near the 8051 RAM limits, or to long-term uIP/UDP usage rather than a specific code path?
3. Is a software deadman fed from the main loop a sound approach, or is there a better pattern?
4. Is there anything in the tree that legitimately keeps the main loop from running for seconds at a time (so a 10 s budget would be wrong)?

Happy to test a debug build with extra console output or add instrumentation if that helps narrow it down. Thanks for the project and for any pointers.

[...]

PR #406, opened 2026-09-06
dhcp: fix transaction-id check so it actually compares

DrDoof:
`parse_dhcp()` guarded the transaction id with

    if (!DHCP_P->tid == HTONS(dhcp_state.transaction_id))
        return;

Unary `!` binds tighter than `==`, so this is `(!tid) == HTONS(id)`, which
is false for any non-zero id and never returns. The id is never actually
checked, so a DHCP reply for another client on the segment (the socket is
bound broadcast) is parsed as ours: a broadcast OFFER makes us chase
another host's address and a broadcast ACK makes us adopt it.

Compare the field directly. The id is network-order in the packet and
host-order in the state, matching the HTONS() used when the request is
built, so `DHCP_P->tid != HTONS(...)` is the intended test.

Found by inspection while looking at #404; not yet reproduced on hardware.
Builds for all machines (`make machine_check`).

### DIFF
--- a/dhcp.c
+++ b/dhcp.c
@@ -94,7 +94,7 @@ void dhcp_prepare_request(void)
 	DHCP_P->hw_len = 6;
 	DHCP_P->hops = 0;
 
-	DHCP_P->tid = HTONS(dhcp_state.transaction_id);
+	DHCP_P->tid = dhcp_state.transaction_id;
 	DHCP_P->delay = HTONS(0);
 	DHCP_P->flags = 0;
 	// Clear fields client_ip to bootp_file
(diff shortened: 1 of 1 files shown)


vDorst:
That is a nice find.
I also have my suspicions about by `HTONS` because that is for 16-bit numbers and both are 32-bit numbers.

```c
struct dhcp_pkt {
   ..
   uint32_t tid;
   ..
}
```

DrDoof:
Yeah, you're right. The xid just has to come back the way it left, so there's no byte order to keep, and HTONS was quietly halving it too. Gone from both sides now (97d44fb).

vDorst:
Thanks

PR #407, opened 2026-09-06
Age the ARP cache by calling uip_arp_timer()

DrDoof:
`uip_arp_timer()` is defined but never called, so the ARP cache is only
refreshed on new learning and only evicted under table pressure. A cached
entry for a peer whose MAC later changes (a gateway failover, a replaced
device) keeps being used until it is evicted, so traffic reached through it
is sent to the old MAC and lost, while on-link traffic is unaffected.

uIP expects the timer roughly every ten seconds; drive it from the
once-per-second housekeeping in `idle()`.

One candidate for the management-plane wedge in #404 (off-subnet dies,
on-subnet keeps working). Builds for all machines (`make machine_check`).

### DIFF
--- a/rtlplayground.c
+++ b/rtlplayground.c
@@ -89,6 +89,7 @@ volatile __xdata uint32_t ticks;
 volatile __xdata uint8_t sec_counter;
 volatile __xdata uint16_t sleep_ticks;
 __xdata uint8_t stp_clock;
+__xdata uint8_t arp_age_secs;
 extern __xdata struct dhcp_state dhcp_state;
 
 #define STP_TICK_DIVIDER 3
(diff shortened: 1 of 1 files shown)


vDorst:
Thanks, takes a while to test, but it works.

```
Arp TIME!
uip_arp_timer()
arp idx: 00 time 7d - 05 = 78
arp drop 
arp idx: 01 time 7d - 00 = 7d
arp idx: 02 time 7d - 00 = 7d
arp idx: 03 time 7d - 00 = 7d
arp idx: 04 time 7d - 00 = 7d
arp idx: 05 time 7d - 00 = 7d
arp idx: 06 time 7d - 00 = 7d
arp idx: 07 time 7d - 00 = 7d
```

PR #408, opened 2026-09-06
Do not call write_char() from the external interrupt handlers

DrDoof:
The EX0..EX3 handlers each emitted a debug character through `write_char()`,
which busy-waits on the serial TX-empty flag set by the serial interrupt.
If one of these runs while a byte is in flight, it waits for a flag the
equal-or-lower-priority serial interrupt cannot deliver, hanging the handler
with the system tick stopped.

The characters were only debug markers, so drop them; the handlers now just
acknowledge the interrupt. No ISR should call the blocking console path.

A latent trap noted while investigating #404 (I could not get these handlers
to fire under load on an RTL8372, so likely not that report's cause, but the
deadlock is real). Builds for all machines (`make machine_check`).

### DIFF
--- a/rtlplayground.c
+++ b/rtlplayground.c
@@ -493,7 +493,6 @@ void print_cmd_prompt(void)
 void isr_ext0(void) __interrupt(0)
 {
 	EX0 = 0;	// Disable interrupt for the moment
-	write_char('X');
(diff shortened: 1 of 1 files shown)


vDorst:
thanks

PR #409, opened 2026-09-06
feat(translate): add Spanish (es) translations

gnacho:
## Summary
Add a complete Castilian Spanish (es) translation of the web UI, following the existing `en` / `ja` / `zh` pattern in `html/i18n.js`. This also wires up two areas that were previously hardcoded in English **for every language**:

- The full **Spanning Tree** page (`stp.html` + `stp.js`): it had zero `data-i18n` attributes / i18n keys, so all its labels, states, roles, option labels and status messages were English-only. Now fully localized with new `stp_*` keys (added to `en` and `es`).
- The **detailed port counters** popup (`stat.js`): the 89 RMON/MIB counter labels were hardcoded in the `mib_counters` array. Now they resolve through `t()`, with Spanish values added to `es` (the other languages keep the English label via the existing fallback).

## What's in this PR
- `html/i18n.js`: `es` block for the base 171 keys (full 1:1 coverage), plus `stp_*` / `nav_stp` keys in both `en` and `es`, plus the 89 RMON labels in `es`.
- `html/system.html`: added `Español` to the language selector.
- `html/stp.html`: `data-i18n` wiring on all labels (header cells restructured so the Application/InnerHTML translation can't nuke embedded children; the "(0 = Auto)" hint became a table note).
- `html/stp.js`: localized states, roles, port options, `(default)` / `(guard!)`, True/False and the status messages (with a small `{placeholder}` helper).
- `html/stat.js`: the RMON label rows now resolve through `t()`.
- `html/navigation.js`: added `data-i18n="nav_stp"` to the sidebar STP link (the only sidebar entry previously not wired).

## Verification
- `node --check` passes on `i18n.js`, `stp.js`, `stat.js`, `navigation.js`.
- Programmatic check: all `data-i18n` / `t()` / `tf()` keys used exist in both `en` and `es` (0 missing either way); 0 em/en dashes in the Spanish text.
- `t()` keeps untranslated keys falling back to `en` (so `ja` / `zh` users see English, as before).


vDorst:
Thank you for contribution, this feature is hold back until #315 is merge.
Than you have to rebase it.

vDorst:
Please rebase, a lot has changes after #429.

gnacho:
Wow, that's an amazing change!! Yeah, I'll rebase it.

gnacho:
Rebased onto main after #429 and adapted the Spanish translations to the new single-page app format. Changes are now in html/app.js and html/login.html; node --check passes on app.js.

vDorst:
Did you test this?

Because you also need to add this to html/index.html line 395.          
```html
<select class="in" id="langSel">
...
</select>
```
Otherwise you can't select the language in System -> Display -> Language

gnacho:
> Did you test this? Because you also need to add this to html/index.html line 395.

Yes, tested it on my switch running this firmware, the whole UI renders in Spanish. You were right about the selector though: I had `rtl_lang=es` left in localStorage from the previous build, so the page came up in Spanish without me ever opening the dropdown, and that's how I missed it. Added the missing `es` option to `html/index.html` in 45fb382. Thanks for catching it!

vDorst:
Thanks

PR #410, opened 2026-09-07
httpd: retransmit a continued chunk from where it was sent

DrDoof:
Anything larger than the 2500-byte output buffer goes out in two phases. The first buffer is sent in MSS-sized pieces with `o_idx` walking through it, and once that is drained the rest is streamed from flash a chunk at a time into `outbuf`, always from offset 0.

The continuation branch never resets `o_idx`, so it keeps whatever the first phase left there, normally 1460. Meanwhile `uip_rexmit()` retransmits with `uip_send(outbuf + o_idx, slen)`. So when a continuation segment goes missing, what goes out in its place is the bytes 1460 further into the buffer, and the client takes those as the data it was waiting for. With `o_idx + slen` reaching 2920 against a 2500-byte buffer it also reads 420 bytes past `outbuf`, which happens to be where httpd keeps `entry`, `slen`, `o_idx`, `len_left` and `cont_len`.

Nothing complains, because the length is right. The transfer completes, the file is exactly as long as it should be, and only the contents are wrong, so it shows up as mangled JS or CSS rather than as a failed download.

Reproduced on a SWTGW218AS by dropping 15% of the switch's TCP segments at the client. The direction matters: dropping ACKs proves nothing, because then the switch retransmits data the client already has and TCP throws it away as a duplicate. Fetching the 21 kB `i18n.js` eight times each way:

```
before:  7 of 8 corrupted
after:   0 of 8 corrupted
```

Every corrupted copy was exactly 21326 bytes, and the first differing byte landed at 3625, 6545, 6545, 9465, 16765, 18225 and 19685. Those are all the same offset modulo 1460, and that fixed phase inside the chunk is what pointed at the retransmission rather than at the flash reads.

The fix is `o_idx = 0` where the chunk is read in. The `o_idx += slen` sitting right after `slen = 0` went too, since it only ever added zero.

### DIFF
--- a/httpd/httpd.c
+++ b/httpd/httpd.c
@@ -865,7 +865,6 @@ void httpd_appcall(void)
 			o_idx += uip_mss();
 		} else {
 			slen = 0;
-			o_idx += slen;
 		}
 
 		s->tstate = TSTATE_ACKED;
@@ -886,7 +885,8 @@ void httpd_appcall(void)
 			flash_region.addr = cont_addr;
 			flash_region.len = slen;
 			flash_read_bulk(outbuf);
-			uip_send(outbuf, slen);
+			o_idx = 0;
+			uip_send(outbuf + o_idx, slen);
 			cont_len -= slen;
 			cont_addr += slen;
 			s->tstate = TSTATE_TX;


httpd/httpd.c
```diff
@@ -886,6 +885,9 @@ void httpd_appcall(void)
 			flash_region.addr = cont_addr;
 			flash_region.len = slen;
 			flash_read_bulk(outbuf);
+			/* The chunk starts at outbuf[0], so a retransmission has to as
```
logicog: The comment sounds like a justification for the fix. But if you imagined that it was right from the beginning, it would not be there. Anyway, the justification is in the commit comment. So I suggest to drop this comment.

logicog (commented):
Small comment only. Otherwise good --- and great catch!

httpd/httpd.c
DrDoof: You are right about the second half of it. It only means anything to someone who already knows what the code did before, and the place for that is the commit message.

What stops me from simply deleting it is how the line reads on its own:

```c
o_idx = 0;
uip_send(outbuf, slen);
```

The send underneath it never touches o_idx, so the assignment looks like a dead store. The only thing that makes it matter is the uip_rexmit() branch at the end of the function. This same patch removes another o_idx assignment for exactly that reason, so someone tidying this one away later does not strike me as far-fetched.

How about letting the continuation send go through o_idx like every other send in httpd_appcall()? The other six are all uip_send(outbuf + o_idx, ...), and this branch is the only one passing a bare outbuf:

```c
o_idx = 0;
uip_send(outbuf + o_idx, slen);
```

Same bytes on the wire, the assignment is visibly used by the line below it, and the comment can go entirely.

If you would rather leave the send as it is, say so and I will just drop the comment.

logicog: > Same bytes on the wire, the assignment is visibly used by the line below it, and the comment can go entirely.

That would be a very nice solution! Great idea.

DrDoof: Thanks, done. The comment is gone and the send now goes through o_idx like the other six.

I was curious whether it really costs nothing, so I compiled it both ways and compared the generated assembly: 5101 instructions each, identical line for line. The only thing that changed is how the code reads, which was the whole point.

DrDoof: I have to correct something I wrote above. I said the two forms compiled to the same thing. They do not, and I got there by comparing the wrong pair of builds.

`uip_send(outbuf + o_idx, slen)` loads `o_idx` out of xdata and adds it to the buffer address, where the plain form emits a single `mov dptr,#_outbuf`. That is twelve extra instructions and 17 bytes of BANK1. SDCC does not fold it away, I assume because `o_idx` is a global that an interrupt could touch between the assignment and the use.

So the readability argument still holds, but it is not free and I told you it was. If those 17 bytes are not worth it here, I will put the send back to `uip_send(outbuf, slen)` and leave `o_idx = 0;` standing on its own without the comment, which is what you asked for to begin with. Happy either way, it is your call.

vDorst:
thanks

PR #411, opened 2026-09-07
httpd: stop a firmware upload at the end of the image area

DrDoof:
Nothing bounds `uptr`. A body whose closing boundary never turns up, or just one bigger than the image area, keeps writing page after page straight past the staging region. On a part exactly twice `FIRMWARE_UPLOAD_START` those writes run off the end of the chip, the SPI address wraps to zero, and a bad upload takes the running firmware with it. Fair warning on that last step: I have only 2 MB parts here, where the overrun lands in unused space, so that bit is address arithmetic rather than something I watched happen.

The length limit does not stand in for this. `max_upload` is 1024576 bytes, while the image area, `FIRMWARE_UPLOAD_START` to twice it, is 524288. A body half a megabyte larger than the region it writes into still passes that test.

I put the bound where `uptr` actually moves, in the page-write path, rather than on the receiving side. A receive-side test against `uip_len` gets the final segment wrong: it still carries the closing multipart boundary, which counts towards the length but never reaches flash, so a perfectly good image would be refused right at the finish line. Checking just before the page write also has the nice property that nothing has left the area by the time it fires.

The area itself runs from `FIRMWARE_UPLOAD_START` to twice that, the span `check_and_flash_update_image()` reads back and checksums. A half-written image left behind does no harm: boot rejects it on the checksum and erases it.

Tried it on a SWTGW218AS by posting 614400 bytes into a 524288-byte area:

```
HTTP/1.1 400 Bad Request
NO: upload exceeds image area
```

and the switch was none the worse for it afterwards: same version running, statics still served byte for byte.

### DIFF
--- a/httpd/httpd.c
+++ b/httpd/httpd.c
@@ -60,7 +60,6 @@ __xdata uint16_t content_length;
 // Global variables holding POST state
 __xdata uint16_t bindex; // Current index into the boundary
 __xdata uint8_t verify_crc;
-__xdata uint32_t max_upload;
 __xdata uint16_t short_parsed;
 
 #define POSTBODY_CMD	1
@@ -529,6 +528,19 @@ uint8_t stream_upload(void)
 			crc16(upload_settings.p + upload_settings.bptr);
 			flash_buf[write_len++] = upload_settings.p[upload_settings.bptr++];
 			if (write_len >= FLASH_PAGE_SIZE) {
+				/* The staged image spans FIRMWARE_UPLOAD_START to twice that,
+				 * the span check_and_flash_update_image() reads back. Nothing
+				 * else bounds uptr: a body whose closing boundary never arrives
+				 * keeps writing, and on a flash exactly this size the address
+				 * wraps onto the running image at zero. */
+				if (uptr >= (uint32_t)FIRMWARE_UPLOAD_START * 2) {
+					print_string("Upload runs past the image area! Aborting.\n");
+					slen = strtox(outbuf, "HTTP/1.1 400 Bad Request\r\nContent-Length: 30\r\n"
+						"Content-Type: text/plain\r\n\r\n"
+						"NO: upload exceeds image area\n");
+					s->tstate = TSTATE_NONE;
+					return 0;
+				}
 				dbg_string("len: "); dbg_short(write_len); dbg_char(' ');
 				dbg_string("CRC16: "); dbg_short(crc_value); dbg_char('\n');
 				if (uptr % FLASH_SECTOR_SIZE == 0) {
@@ -759,7 +771,6 @@ void handle_post(void)
 			config_upload = 0;
 			uptr = FIRMWARE_UPLOAD_START;
 			verify_crc = 1;
-			max_upload = 1024576;
 			pre_acc = 0;
 		} else if (is_word(request_path, "config")) {
 			if (!authenticated) {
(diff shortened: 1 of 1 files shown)


bloqaudio (approved):
Read, build-verified and hardware-tested on an RTL8373 board, `PCB_SWTG018AS_V2_1_0`, with the commit cherry-picked onto current main and nothing else changed. The bound is correct and I think it should go in. The description has gone stale since you opened it, and the change that made it stale is mine.

**On the description.** It says `max_upload` is set once and never decremented, and quotes a check `max_upload - uip_len > 0`. That was true on 7 September. My own `edc469f`, "httpd: make the firmware upload length limit count", merged on the 18th, replaced it with `uip_len <= max_upload` and an actual `max_upload -= uip_len`. So the dead check your first two paragraphs describe had already gone from your merge base, through a change of mine rather than anything on your side. It is also why `max_upload` is still present instead of going "along with the dead check": removing it now would drop a limit that finally does something. The diff is right, the prose in front of it is out of date.

**The bound is still needed, which the arithmetic settles.** `max_upload` is 1024576 bytes. The image area is `FIRMWARE_UPLOAD_START` to twice it, `0x80000` to `0x100000`, which is 524288 bytes. The length limit therefore admits a body 500288 bytes larger than the region it writes into, so it cannot stand in for this check. Your bound in the page-write path is what actually stops the overrun.

**Placement and the off-by-one.** Checking before the page write, with `uptr` advancing after it, makes the boundary exact rather than merely close. A full 524288-byte image writes its last page with `uptr` at `0xFFF00`, below the limit, and only then reaches `0x100000`, so nothing legitimate is refused. An oversized body finds `uptr` already at `0x100000` on the next page and stops. I confirmed both rather than taking it on reading alone.

Hardware, in order:

```
POST 700000 bytes into the 524288-byte area
  HTTP/1.1 400 Bad Request
  NO: upload exceeds image area

afterwards: same version running, stored config intact, login page served, uplink up at 2.5G

POST 524288 bytes, the exact area size, contents deliberately not a valid image
  HTTP/1.1 400 Bad Request
  NO: checksum failed, not applied
```

The second is the one I cared about, since a bound in the wrong place would have refused that at the finish line with the wrong message. It reaches the checksum stage, so it does not.

**What I could not test.** The wrap onto the running image needs a part exactly twice `FIRMWARE_UPLOAD_START`. This board has 2 MB, like yours, so the overrun lands in unused space here too. I did not deliberately overrun the area to watch it, since I do not know what occupies the space above 1 MB on this part and the unbounded write is plain enough from the source. That last step is still address arithmetic, as you said.

**One number worth having.** This costs 207 bytes of `BANK1`, 47994 to 48201 against current main, which leaves 951 free. Direct RAM and the other banks are unchanged. `BANK1` is the tight one on this part, so it is worth knowing that the error string and its print are most of the cost, not the comparison.

Approving. Worth editing the first two paragraphs of the description before this is merged, so the commit does not land describing a check that no longer exists.

DrDoof:
Thank you for going through it that carefully, and for spotting that half my description had quietly gone out of date. Your edc469f is what took the dead check out, so those paragraphs were describing a tree that had already moved on without me noticing.

I have rewritten them. The opening now starts from uptr being unbounded, and the max_upload line is gone, since taking it out would drop a limit that finally does something. I also borrowed your arithmetic for why the length limit cannot stand in for this bound, as it says in one line what my two paragraphs were circling around.

The 207 bytes of BANK1 is good to know as well, and it is reassuring that most of it is the error string rather than the comparison.

feelfree69:
Thanks!

PR #413, opened 2026-09-07
vlan: bound the name table before writing into it

DrDoof:
A VLAN name is copied into `vlan_names` with nothing comparing `vlan_ptr` against `VLAN_NAMES_SIZE`. The only limit is the 128-byte command line, so enough named VLANs simply walk off the end of the 1 kB table.

What sits right behind the table is what makes this nastier than a plain overflow. From the map:

```
008A  vlan_names   (1024)
048A  vlan_ptr     <- first thing overwritten
048C  port_names
05AC  cmd_buffer
```

The first casualty is the index driving the write, so once it is clobbered the following bytes land wherever it now points. And since the configuration is replayed at boot, a config file that overflows does it again on every start rather than once when somebody typed it.

So: measure the name first, and refuse it if the three id digits, the name, its separating space and the terminator would not fit.

Two things about where the check sits. It goes before `vlan_name_remove()`, so a name that does not fit cannot take the existing one down with it. And a rejected name does not fail the whole command, the VLAN still gets its ports: dropping the line outright would lose port membership during a config replay, which is a worse outcome than losing a name.

I have not driven a real table overflow on hardware for this one. Doing that on a live switch means writing over `port_names` and `cmd_buffer` to prove a point, and the layout above is already the argument. The build is clean and `make machine_check` passes.

### DIFF
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -504,19 +504,29 @@ void parse_vlan(void)
 		}
 
 		uint8_t w = 2;
-		if (cmd_words_len > w && isletter(cmd_buffer[cmd_words_b[w]])) {
+		uint8_t pos = cmd_words_b[w];
+		if (cmd_words_len > w && isletter(cmd_buffer[pos])) {
 			uint8_t i = 0;
-			vlan_name_remove(vlan_settings.vlan);
-			vlan_names[vlan_ptr++] = hex[(vlan_settings.vlan >> 8) & 0xf];
-			vlan_names[vlan_ptr++] = hex[(vlan_settings.vlan >> 4) & 0xf] ;
-			vlan_names[vlan_ptr++] = hex[vlan_settings.vlan & 0xf];
-			while(!cmd_is_space_or_nul(cmd_words_b[w] + i)) {
-				write_char(cmd_buffer[cmd_words_b[w] + i]);
-				vlan_names[vlan_ptr++] = cmd_buffer[cmd_words_b[w] + i++];
+			while (!cmd_is_space_or_nul(pos + i))
+				i++;
+			if (vlan_ptr + i > (VLAN_NAMES_SIZE - 5)) {
+				print_string("VLAN name table full, name ignored\n");
+			} else {
+				vlan_name_remove(vlan_settings.vlan);
+				__xdata uint8_t *vlan_str = &vlan_names[vlan_ptr];
+				*vlan_str++ = hex[(vlan_settings.vlan >> 8) & 0xf];
+				*vlan_str++ = hex[(vlan_settings.vlan >> 4) & 0xf] ;
+				*vlan_str++ = hex[vlan_settings.vlan & 0xf];
+				while(!cmd_is_space_or_nul(pos)) {
+					uint8_t c = cmd_buffer[pos++];
+					write_char(c);
+					*vlan_str++ = c;
+				}
+				*vlan_str++ = ' '; *vlan_str = NUL;
+				vlan_ptr = vlan_str - vlan_names;
+				print_string("<\n");
 			}
-			vlan_names[vlan_ptr++] = ' '; vlan_names[vlan_ptr] = NUL;
 			w++;
-			print_string("<\n");
 		}
 
 		uint8_t ret;


cmd_parser.c
```diff
@@ -506,17 +506,30 @@ void parse_vlan(void)
...
-				write_char(cmd_buffer[cmd_words_b[w] + i]);
-				vlan_names[vlan_ptr++] = cmd_buffer[cmd_words_b[w] + i++];
+			while (!cmd_is_space_or_nul(cmd_words_b[w] + i))
+				i++;
+			/* Three hex digits of the id, the name, its separating space
+			 * and the terminator. Nothing else bounds vlan_ptr, and the
+			 * first thing behind the table is vlan_ptr itself, so running
+			 * over corrupts the index that drives the write. Checked
+			 * before the removal, so a name that does not fit does not
+			 * take the existing one with it. */
```
vDorst: I am not happy this all these long AI text comments.
As a PR text, it is fine not as a comment for a bound-check.

DrDoof: Fair point, and the comment is gone. It was repeating the commit message almost word for word, so it was not carrying anything the log does not already say.

vDorst:
Sorry for hacking/modify it again.
But this code saves +100 bytes.

I am trying to lookup data in XDATA once.
Keep the pointer read and increments in one operation.
Also compiler don't simplify constant values in a if-statement.
`if (vlan_ptr + i + 5 > VLAN_NAMES_SIZE)` `+5` is added at runtime. (cost 6-bytes)

```patch
index 9d374fd..14bc426 100644
--- a/cmd_parser.c
+++ b/cmd_parser.c
@@ -504,23 +504,28 @@ void parse_vlan(void)
                }
 
                uint8_t w = 2;
-               if (cmd_words_len > w && isletter(cmd_buffer[cmd_words_b[w]])) {
+               uint8_t pos = cmd_words_b[w];
+               if (cmd_words_len > w && isletter(cmd_buffer[pos])) {
                        uint8_t i = 0;
-                       while (!cmd_is_space_or_nul(cmd_words_b[w] + i))
+                       while (!cmd_is_space_or_nul(pos + i))
                                i++;
-                       if (vlan_ptr + i + 5 > VLAN_NAMES_SIZE) {
+                       if (vlan_ptr + i > (VLAN_NAMES_SIZE - 5)) {
                                print_string("VLAN name table full, name ignored\n");
                        } else {
-                               i = 0;
                                vlan_name_remove(vlan_settings.vlan);
-                               vlan_names[vlan_ptr++] = hex[(vlan_settings.vlan >> 8) & 0xf];
-                               vlan_names[vlan_ptr++] = hex[(vlan_settings.vlan >> 4) & 0xf] ;
-                               vlan_names[vlan_ptr++] = hex[vlan_settings.vlan & 0xf];
-                               while(!cmd_is_space_or_nul(cmd_words_b[w] + i)) {
-                                       write_char(cmd_buffer[cmd_words_b[w] + i]);
-                                       vlan_names[vlan_ptr++] = cmd_buffer[cmd_words_b[w] + i++];
+                               __xdata uint8_t *vlan_str = &vlan_names[vlan_ptr];
+                               *vlan_str++ = hex[(vlan_settings.vlan >> 8) & 0xf];
+                               *vlan_str++ = hex[(vlan_settings.vlan >> 4) & 0xf] ;
+                               *vlan_str++ = hex[vlan_settings.vlan & 0xf];
+                               while(!cmd_is_space_or_nul(pos)) {
+                                       uint8_t c = cmd_buffer[pos++];
+                                       write_char(c);
+                                       *vlan_str++ = c;
                                }
-                               vlan_names[vlan_ptr++] = ' '; vlan_names[vlan_ptr] = NUL;
+                               *vlan_str++ = ' '; *vlan_str = NUL;
+
+                               // Calculate length that is written to vlan_names[].
+                               vlan_ptr = vlan_str - &vlan_names;
                                print_string("<\n");
                        }
                        w++;
```

DrDoof:
This is much better than what I had, thank you. I measured it against main with the version string pinned so it would not drift into the result: your version is 118 bytes below mine, and 25 below main as it stands today without any bounds check at all. So the check ends up paying for itself and leaving change, which I honestly did not expect.

Two small things on the way in. `vlan_str - &vlan_names` would not compile here, `&vlan_names` being a pointer to the array rather than to its first element, so it went in as `vlan_str - vlan_names`. And I left the `// Calculate length` line out, since the expression reads clearly enough on its own.

The point about the constant is the one I would not have found on my own. It had never occurred to me that `+5` on the wrong side of the comparison costs anything at all, and it is exactly the kind of thing I will be looking for from now on.

vDorst:
Thanks!

Thanks for fixing my pointer issues/notation.

> The point about the constant is the one I would not have found on my own. It had never occurred to me that +5 on the wrong side of the comparison costs anything at all, and it is exactly the kind of thing I will be looking for from now on.

For fun I try to see if I can make the code generation better.
But to see that you often have to look into the generated assembly-code.
Than this becomes visible.

That is why I do sometimes weird `__bit` tricks, try to force the compiler to use specific instruction, like 'jbc' to save memory and instruction.


`*vlan_str++` vs `vlan_names[vlan_ptr++]` is not always clear which one is the best.
I try this trick on other parts of the code, but it didn't save any code-space.

If it is just a fixed xdata memory location and a 8-bit value is added to the pointer.
Then this is the most time cheaper `vlan_names[vlan_ptr++]` then only one or two operations are preformed to that location.

But for me it is fun to test this out and measure the code/mem-size.

Issue #386, opened 2026-09-01
SDCC spills `__xdata` automatics through DSEG slots


bloqaudio:
Observed on SDCC 4.5 (`-mmcs51`, default small model) while working on #374.

A `__xdata` qualifier on an automatic variable places its home in xdata, but the register allocator still caches the value in registers, and when it has to survive a call or a loop back-edge it spills to an `sloc` temporary in DSEG. Those slots are overlaid per function, so their count follows the register pressure in the function rather than the number of variables. With direct RAM fully allocated (`No spare internal RAM space left` in the `.mem` file) every slot comes straight out of the stack.

Measured in `config_take()` (`httpd/httpd.c`), full builds, SWTGW218AS:

- the six parse cursors as plain `__xdata` automatics: stack start moves from `0x63` to `0x69` (6 bytes of direct RAM), code +173, `sloc` references in `httpd.asm` go from 197 to 271
- the same six as `static __xdata`: no spill slots, internal RAM map identical to `main`
- `cfg_bl` alone as a plain automatic: no direct RAM cost at the current register pressure, code +40

`static` avoids it because statics are never register-allocated, at the price of the extra loads and stores. The same shape shows up in `early_boot_handle_button()` (`rtlplayground.c`), whose timing locals are `__xdata` automatics.

Left to investigate:

- which SDCC options, if any, keep the spill homes out of direct RAM (`--nooverlay`, `--max-allocs-per-node`, `--opt-code-size` are the candidates I have not measured)
- whether the allocator can be told that an `__xdata` automatic is not worth caching across a call
- a rule for when a `static` is worth its code cost, so this does not get decided per review

The comment above the statics in `config_take()` references this issue.

bloqaudio:
Coming back to the two open questions with numbers.

All figures are clean builds of `0cc09e4`, `MACHINE=SWTGW218AS`, SDCC 4.5.0. `stack` is the initial stack start from the `.mem` file, `avail` the bytes it reports available, `sloc` the count of spill references in `httpd.asm`. The variable under test is the group of parse cursors in `config_take()`.

| variant | stack | avail | ROM | sloc |
| --- | --- | --- | --- | --- |
| `static __xdata`, as main has it today | `0x6d` | 147 | 113213 | 191 |
| plain `__xdata` automatics | `0x73` | 141 | 113342 | 265 |
| `volatile __xdata` automatics | `0x6d` | 147 | 113308 | 191 |
[...]

PR #415, opened 2026-09-07
syslog: stop the log ring from lapping its reader

DrDoof:
`write_char()` masks `writeptr` and never looks at `readptr`, so once more than `LOGBUF_SIZE` bytes are written between two drains the ring wraps over content that has not been sent yet.

That is not a corner case. The ring is emptied once per main loop pass, a command runs to completion inside a single pass, and a full L2 listing prints well over twice the 512-byte buffer.

Watching it on a SWTGW218AS with a capture on the syslog server: the console printed 1227 bytes in 38 lines, and what arrived was one 174-byte datagram beginning like this:

```
<14>SWTGW218AS 29    0x0002    learned    9
6c:50:4d:50:e8:69    0x0002    learned    9
```

The first line starts at `29`, the tail of a MAC address whose beginning had already been overwritten. The rest of the dump was simply gone.

Drop the character when the ring is full rather than overwriting, so what has been logged stays intact and it is the tail that goes missing. A full ring also flags a line: otherwise a burst with no newline in it would leave the reader waiting on `line_available` forever and syslog would stop for good.

With that in place the same dump comes back as a 525-byte datagram that starts where the listing does:

```
<14>SWTGW218AS     MAC        VLAN    type    port
dc:a6:32:a3:53:af    0x0002    learned    9
```

What is missing now is the end of the listing rather than its beginning, which is the trade worth having.

The dropped bytes are silent, with no marker in the stream. A marker needs a byte of state to remember that something went missing, and that felt like a separate decision rather than part of this fix.

### DIFF
--- a/rtlplayground.c
+++ b/rtlplayground.c
@@ -277,8 +277,19 @@ void write_char(char c)
 	write_char_no_syslog(c);
 
 	if (syslog_state.enabled) {
-		logbuf[syslog_state.writeptr++] = c;
-		syslog_state.writeptr &= (LOGBUF_SIZE - 1);
+		__xdata uint16_t next = (syslog_state.writeptr + 1) & (LOGBUF_SIZE - 1);
+
+		/* The ring is drained once per main loop pass, and a single command
+		 * can print several times its size, so writing on regardless laps
+		 * the reader and the datagram starts mid record. Drop instead, and
+		 * flag a line so the reader empties what is there even if no
+		 * newline has been seen yet. */
+		if (next == syslog_state.readptr) {
+			syslog_state.line_available = 1;
+			return;
+		}
+		logbuf[syslog_state.writeptr] = c;
+		syslog_state.writeptr = next;
 		if (c == '\n')
 			syslog_state.line_available = 1;
 	}


bloqaudio (approved):
Reproduced, build-verified and hardware-tested on an RTL8373 board, `PCB_SWTG018AS_V2_1_0`. It behaves exactly as you describe, and the reasoning holds up.

The bug, on a build without this commit. A 56 entry `l2` listing produced one 158 byte datagram beginning:

```
<14>devswitch ed	8
bc:24:11:10:73:b4	0x0001	learned	8
```

`ed\t8` is the tail of `learned\t8`, so the ring had lapped and the datagram starts mid record, the same shape you saw.

The same listing with this commit cherry-picked onto current main, nothing else changed, gives one 524 byte datagram:

```
<14>devswitch 	MAC		VLAN	type	port
b8:27:eb:7b:c5:64	0x0001	learned	8
```

524 rather than your 525 because the hostname is a character shorter here. It applies to current main without conflict.

Cost, SWTGW218AS, against this branch's own merge base rather than today's main:

| | base `6488433` | this PR |
| --- | --- | --- |
| stack start | `0x68` | `0x68` |
| direct RAM available | 152 | 152 |
| `BANK1` | 47949 | 47949 |
| `CSEG` | 13469 | 13478 |

Nine bytes of `CSEG`, nothing else.

One thing I went looking for and did not find, which seemed worth checking given where this lands. `next` is a plain `__xdata` automatic and `write_char()` runs for every character the firmware prints, so it was a candidate for the spill behaviour in #386, where such an automatic can push register pressure into direct RAM. It does not here: `_write_char` has no `sloc` slots in the generated assembly either side of the change, and the internal RAM map is identical. The concern does not apply.

Worth noting that this also fixes something the description does not claim. When exactly `LOGBUF_SIZE` bytes were written between two drains, the old code wrapped `writeptr` onto `readptr`, and `syslog_callback()` computes `log_size = writeptr - readptr`, which is then zero. The reader saw an empty ring and discarded the whole buffer rather than sending a garbled one. Keeping a slot free makes that state unreachable.

One question rather than a change request. Dropping at the head means the bytes lost are the newest, so an error printed after a large dump in the same pass is lost entirely, where previously it would have survived by overwriting older content. Dropping the oldest instead would keep the recent end, at the cost of having to resync to a record boundary so the datagram does not start mid line again. Your trade is defensible and lapping is plainly worse than either, so this is not blocking, but it seemed worth having in the thread.

Approving.

feelfree69:
Thanks for fixing, reproducing and reviewing!

PR #418, opened 2026-09-07
crc16: leave the code bank as it was found

DrDoof:
The lookup tables live in BANK1, so `crc16` switches the code bank to reach them with `movc`, and then never switches it back.

The routine itself sits in the common segment, which is always mapped, so the call always works. The return is where it shows: that lands in whatever bank the caller runs from, and the window is now BANK1 regardless of what the caller had.

Both callers survive it today, by luck rather than design. One is in the common segment, where the window does not matter, and the other is in BANK1, which happens to be exactly what is left selected. Move either of them to another bank, which is an ordinary thing to do when a bank fills up, and the return goes to the right address in the wrong window: the caller carries on executing whatever happens to sit there in BANK1.

Save the bank and put it back. `push` and `pop` on a direct address cost four bytes of the common segment, and the trap stops waiting for the next reshuffle.

Verified where it counts, since both call sites run the checksum over a whole image. Uploading firmware to a switch running this answered `OK: checksum verified`, which is the BANK1 caller, and the boot that followed re-checksummed the staged image from the common segment and applied it, which is the other one. Builds clean, `make machine_check` passes.

### DIFF
--- a/crc16.asm
+++ b/crc16.asm
@@ -3,6 +3,8 @@
 ;
 	.globl 	_crc_value
 	.globl 	_crc16
+	.globl 	_crc16_bank1
+	.globl 	__sdcc_banked_ret
 	.equ	BANK, 0x96
 ;	.equ	DPS, 0x86
 ; Variable in XMEM holding current CRC16 value, being updated
@@ -18,7 +20,7 @@ _crc_value::
 	.area HOME    (CODE)
 	.area CSEG    (CODE)
 ;	.area BANK1   (CODE)
-_crc16:
+_crc16_bank1:
 	mov	BANK, #1
 	push	dph
 	push	dpl
@@ -51,6 +53,10 @@ _crc16:
 	ret
 
 	.area BANK1   (CODE)
+
+_crc16:
+	lcall	_crc16_bank1
+	ljmp	__sdcc_banked_ret
 
 crc16_table_l:
 	.byte #0x00, #0xc1, #0x81, #0x40, #0x01, #0xc0, #0x80, #0x41
--- a/httpd/httpd.c
+++ b/httpd/httpd.c
@@ -89,7 +89,7 @@ __xdata uint32_t last_session_use;
 
 extern __xdata uint16_t crc_value;
 __xdata uint16_t crc_final;
-void crc16(__xdata uint8_t *v) __naked;
+void crc16_bank1(__xdata uint8_t *v) __naked;
 
 
 inline uint8_t is_separator(uint8_t c)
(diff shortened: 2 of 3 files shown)


logicog:
> Both callers survive it today, by luck rather than design.

Sorry, but that statement is very much not OK. When this was coded, this was very much by design: you only do the things that are necessary, especially when it comes to stack usage. So, this was actually intended as also evidenced by the `__naked` declaration, which is not a `__banked`, so calls from a different BANK are not allowed. But it really was not intended to use the code anywhere else than in the firmware checks. Actually, the by-the-book-solution compared to the one you suggest would be using a proper C-wrapper in BANK1 with a  '__banked', which would also not already cost stack in its current uses.

Is there a way to tune down the arrogance/aggressiveness setting of whatever generates the PR texts?

Also: you again repeat the mistake of putting lengthy comments on obvious one-liners that do not need commenting. This does not help readability and unwillingness to learn is rather frustrating for others around,

DrDoof:
Yes, that was unfortunate writing about "by luck". `__naked` without `__banked` is the contract, the two call sites respect it, and the missing restore follows from that. I looked at the routine and not at its declaration.

Your suggestion is better than what I proposed. A `__banked` wrapper in BANK1 keeps the fast path free of the two stack bytes and puts the cost where the cross-bank call actually happens, instead of charging the existing callers for a case that does not arise. I can redo the PR that way, or close it if you would rather leave the routine as it is; the contract holds as long as nothing calls it from another bank.

The comment goes either way. Four lines on a push and a pop, and you have told me this before, so sorry for making you say it twice. I will keep these shorter.

logicog:
If you have a good idea how to implement, the `__banked` wrapper, I would think that would be the best solution. Not sure how make certain the other call is not used from the wrong places.

DrDoof:
Reworked it the way you suggested. `crc16` is now a `__banked` wrapper in BANK1, three lines sitting next to the tables it already shares a file with, and the routine itself keeps its contract under `crc16_bank1`:

```asm
	.area BANK1   (CODE)
_crc16:
	lcall	_crc16_bank1
	ljmp	__sdcc_banked_ret
```

Six bytes of BANK1 and nothing in the common segment. The two existing call sites say `crc16_bank1` now and pay exactly what they paid before; only a cross-bank caller goes near the trampoline. The four line comment is gone as well.

Your worry about the raw one being reached from the wrong place was the part I spent the most time on, and it turned out better than I expected. The wrapper cannot go just anywhere: it has to be in BANK1 or in the common area. Put it in BANK2 and the call sets PSBANK to 2, the routine sets BANK to 1, and the return comes back to a window that is not mapped any more. So a wrong placement dies on its first call instead of lying in wait, which in this code is about the best one can hope for.

The other half was already there before either of us looked at it. `crc16` has never been in a header, so both callers write out their own prototype. Reaching for the fast path now means typing a name that says where it may be called from, and the plain `crc16` anyone would reach for out of habit is the safe one.

I checked what a BANK2 caller generates and it is the ordinary banked sequence with `r2` taken from `_crc16 >> 16`, so BANK1. Builds clean.

logicog:
That is a very nice solution! Love it.

PR #422, opened 2026-09-07
Do not enter Idle mode from the external interrupt 2 handler

DrDoof:
`isr_ext2()` ends by putting the core to sleep:

```c
void isr_ext2(void) __interrupt(8)
{
	EXIF &= 0xef;	// Clear IRQ flag (bit 7) in EXIF
	PCON |= 1; // Enter Idle mode until interrupt occurs
}
```

On the DW8051 idle ends when any enabled interrupt is activated, so the core does not stay down for long. The handler naps until the next enabled interrupt request turns up, which the 200 Hz tick caps at 5 ms, and then carries on at the instruction after the sleep. That instruction is the return:

```
_isr_ext2:
	anl	_EXIF,#0xef
	orl	_PCON,#0x01
	reti
```

So the sleep gains nothing, and for as long as it lasts the handler holds the priority latch, leaving everything at its level or below waiting behind it, the tick and the console and the NIC receive interrupt included.

The other four uses of `PCON |= 1` are in task context, in the main loop and the delay helpers, which is the ordinary way to use it. This one is the only one inside a handler.

The bit numbers in the two comments beside the masks are corrected on the way past: `0xef` clears bit 4 and `0xdf` clears bit 5, not 7 and 6.

I have not seen EX2 trigger and cannot tell from the source whether it does: the enable is a bare `EX2 = 1` and nothing describes what drives it.

### DIFF
--- a/rtlplayground.c
+++ b/rtlplayground.c
@@ -515,7 +515,7 @@ void isr_ext1(void) __interrupt(2)
  */
 void isr_ext2(void) __interrupt(8)
 {
-	EXIF &= 0xef;	// Clear IRQ flag IE2 (bit 4) in EXIF
+	EXIF &= 0xef;
 }
 
 /*
@@ -524,7 +524,7 @@ void isr_ext2(void) __interrupt(8)
  */
 void isr_ext3(void) __interrupt(9)
 {
-	EXIF &= 0xdf;	// Clear IRQ flag IE3 (bit 5) in EXIF
+	EXIF &= 0xdf;
 }
 
 // Timer2: handles system tick.
--- a/rtlplayground.c
+++ b/rtlplayground.c
@@ -516,7 +516,6 @@ void isr_ext1(void) __interrupt(2)
 void isr_ext2(void) __interrupt(8)
 {
 	EXIF &= 0xef;	// Clear IRQ flag (bit 7) in EXIF
-	PCON |= 1; // Enter Idle mode until interrupt occurs
 }
 
 /*
--- a/rtlplayground.c
+++ b/rtlplayground.c
@@ -515,7 +515,7 @@ void isr_ext1(void) __interrupt(2)
  */
 void isr_ext2(void) __interrupt(8)
 {
-	EXIF &= 0xef;
+	EXIF &= 0xef;	// Clear IRQ flag (bit 4) in EXIF
 }
 
 /*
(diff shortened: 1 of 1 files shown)


vDorst:
> Idle is left when an enabled interrupt is taken, and inside a handler the priority latch blocks everything at the same level or below until RETI. PX3 = 1 in setup_external_irqs() is the only priority raised anywhere in the firmware, so the system tick, the console and the NIC receive interrupt all sit at the same level as EX2 and none of them can end that idle. Only EX3 can, and its source is as undocumented as EX2's.

This is not the case on a DW8051. According to the DW8051 manual

> There are three ways to exit idle mode: activate any enabled interrupt,
por_n, or rst_in_n. Activation of any enabled interrupt causes the
hardware to clear the IDLE bit and terminate idle mode. The CPU

Anyhow, it is a good idea not set the idle mode in a interrupt.

Comment is also wrong, probably a typo. It is bit 4 not 7.
`// Clear IRQ flag (bit 7) in EXIF`

DrDoof:
Thanks for taking the time to check this against the DW8051 manual, and sorry for the confident wording. I hadn't found that passage and reasoned from the generic 8051 priority rules instead. So the handler doesn't freeze anything, it just naps until the next enabled interrupt request comes along, one tick at most, and then returns. Still 5 ms of dead time per EX2 event for nothing, so I'd like to keep the change, and I'll rewrite the description so it says what actually happens rather than what I guessed.

And thanks for spotting the comment, that one's embarrassing. 0xef is bit 4 (IE2), and the ext3 handler has the same slip with 0xdf (bit 5, not 6). Happy to fix both here, or keep this PR to the one line if you'd rather, whichever is easier for you.

vDorst:
BTW I was not again the comment, I think they are useful here. The issue was that the bit number was incorrect. Please add them back.

You can search for `DW8051-databook.pdf`, you can find the full-version on a university-website (UTK[.]edu).

DrDoof:
Understood, and they are back with the numbers fixed: bit 4 for `0xef` and bit 5 for `0xdf`. Sorry for reading your first note as an objection to the comments themselves.

Thanks for the databook pointer as well. I had been working from the generic 8051 rules, which is how I got the wake-up behaviour wrong in the first place, so having the actual part documented is worth a lot.

vDorst:
Thanks!

PR #424, opened 2026-09-08
Move the boot and SFP handling out of the common area into BANK3

DrDoof:
The common segment is where things get tight: sixteen kilobytes shared by every bank, with roughly 2.5 kB free on current main. The startup path and the SFP handling do not really need to be in it, so this suggests moving them into BANK3.

`boot.c` takes `setup_i2c`, `init_smi`, `nic_setup`, `rtl8373_revision`, `rtl8224_enable`, `early_boot_handle_button`, `set_hostname_default` and the two SerDes helpers. `sfp.c` takes `handle_sfp`, `sfp_module_read`, `sfp_apply_quirks`, `sfp_get_info`, `sfp_read_field`, `sfp_print_info`, `setup_sfp_gpio`, the quirk table and the module state. Switching BANK3 on is a single linker flag and nothing else in the build had to change.

Everything exported is `__banked`, because every caller sits outside BANK3: `main()` and the poll loop are in the common area, `cmd_parser.c` reaches `handle_sfp()` and `sfp_print_info()` from BANK2, and `rtl837x_init.c` calls the SerDes helpers from BANK2. The calls going the other way were already banked, so nothing needed doing there.

The quirk table is the part I was least comfortable with. Its entries are `__code` pointers and those carry no bank, so handing one to a function in another bank would give it an offset read through the wrong window. It works out because the consumers all stay in the common area: `strcmp` at 0x570, `print_string` at 0x38f, `write_char` at 0x1ff. I checked it in the built image as well, where the table at 0x35580 holds 0x563c and 0x5644, resolving to "QSFPTEK" and "QT-SFP+-T" inside BANK3.

Sizes, with the version string pinned so it does not drift into the comparison: common segment 13471 to 8510 bytes, CONST 1999 to 1441, BANK3 at 5710, BANK1 unchanged, BANK2 up by 31 for the trampolines. All 34 machines in machine.h build and link. I also walked the generated assembly for plain lcalls crossing a bank boundary, since those compile happily and only bite at run time; there are none among 2787 calls, and I pointed the check at a build with `__banked` taken off `nic_setup` first, to be sure it could still find one. The function bodies are unchanged byte for byte, and the diff adds no comments of its own: 161 comment lines move across, 161 leave, none are new.

I ran it on a SWTGW218AS, though not from this branch exactly, since the image carries these two commits on top of our working branch, which has more on it than main. Twelve reboots came out the same each time: CPU revision read, RTL8224 brought up, SerDes configured from BANK3, saved config replayed. The module is picked up at boot and also survives being pulled and put back while the switch is forwarding, which takes `handle_sfp`, `sfp_module_read`, `sfp_get_info`, `sfp_read_field`, `sfp_apply_quirks` and `sds_config` through in one go. `sfp`, `hostname` and `information.json` all come back right, the last one reading the module strings from `page_impl.c` over in BANK1.

Happy to build it from plain main and try that instead, if you would rather see it that way first.

### DIFF
--- a/boot.c
+++ b/boot.c
@@ -1,0 +1,480 @@
+#pragma codeseg BANK3
+#pragma constseg BANK3
+
+#include <8051.h>
+#include <stdint.h>
+#include <stdbool.h>
+
+#include "rtl837x_sfr.h"
+#include "rtl837x_regs.h"
+#include "rtl837x_common.h"
+#include "rtl837x_flash.h"
+#include "rtl837x_pins.h"
+#include "rtl837x_phy.h"
+#include "rtl837x_port.h"
+#include "rtl837x_init.h"
+#include "machine.h"
+#include "phy.h"
+#include "boot.h"
+
+extern __code const struct machine machine;
+extern __xdata struct machine_runtime machine_detected;
+extern __xdata uint8_t sfr_data[4];
+extern volatile __xdata uint32_t ticks;
+extern __xdata uint8_t tx_seq;
+extern __code uint8_t * __code hex;
+
+
+
+void sds_config_mac(uint8_t sds, uint8_t mode) __banked
+{
+	reg_read_m(RTL837X_REG_SDS_MODES);
+	sfr_data[0] = 0;
+	sfr_data[1] = 0;
+	switch (sds) {
+	case 0:
+		sfr_mask_data(0, 0x1f, mode);
+		break;
+	case 1:
+		sfr_mask_data(0, 0xe0, mode << 5);
+		sfr_mask_data(1, 0x03, mode >> 3);
+		break;
+	case 2:
+		sfr_mask_data(1, 0xfc, 0x02 << 2);
+	}
+	if (machine_detected.isRTL8373) // Set 3rd SERDES Mode to 0x2 for RTL8224
+		sfr_mask_data(1, 0xfc, 0x02 << 2);
+	else
+		sfr_data[2] &= 0x03;
+	reg_write_m(RTL837X_REG_SDS_MODES);
+	print_string("\nRTL837X_REG_SDS_MODES: ");
+	print_reg(RTL837X_REG_SDS_MODES);
+	print_string("\n");
+}
+
+
+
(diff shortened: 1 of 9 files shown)


boot.c
```diff
@@ -0,0 +1,480 @@
...
+
+		PCON |= 1;
+	}
+
+	set_sys_led_state(SYS_LED_ON);
+
+	if ((ticks - start_ticks) >= min_hold_ticks) {
+		print_string("[Button held 10s-30s at boot; restoring default config]\n");
+		set_sys_led_state(SYS_LED_FAST);
+		flash_default_config();
```
vDorst, line 111: Getting a compiler warning here.
```
boot.c:111: warning 112: function 'flash_default_config' implicit declaration
boot.c:111: warning 84: 'auto' variable 'flash_default_config' may be used before initialization
```

```
#rg -tc flash_default_config
rtlplayground.c
38:void flash_default_config(void);
1078:void flash_default_config(void)
1121:                           flash_default_config();

boot.c
111:            flash_default_config();
```

`flash_default_config()` is located in `rtlplayground.c` but not exported via `rtl837x_common.h`.

This solves the warning.
```patch
index d062cd7..97c2a44 100644
--- a/rtl837x_common.h
+++ b/rtl837x_common.h
@@ -188,4 +188,5 @@ bool gpio_pin_test(uint8_t pin);
 void set_sys_led_state(uint8_t state);
 void sds_read(uint8_t sds_id, uint8_t page, uint8_t reg);
 void sds_write_v(uint8_t sds_id, uint8_t page, uint8_t reg, uint16_t v);
+void flash_default_config(void);
```

vDorst (commented):
Only one issue.

boot.c
DrDoof: Good catch, thank you! I applied your patch in e66a177 and dropped the
now duplicate prototype from rtlplayground.c. The call was already safe,
since the function lives in the common area, but the warning is gone and
the image comes out the same size.

vDorst:
Thanks

PR #427, opened 2026-09-09
Temporary revert to SingleIO for Flash operations (except data read)

bennydiamond:
Result of a discussion I started: https://github.com/logicog/RTLPlayground/discussions/425

I understand the change I suggest might be controversial.

I wanted to convert a currently unsupported unmanaged unit. I swaped the stock SPI flash (512KB) for a 1MB Winbond W25Q80DV I had laying around.

It would seem the W25Q80DV will hold in busy state when device tries to read status register while MMIO is enabled.

The change introduced in this very PR fixed the issue for me.

### DIFF
--- a/rtl837x_flash.c
+++ b/rtl837x_flash.c
@@ -28,11 +28,16 @@ __xdata uint8_t flash_capacity_code;
 #define CMD_READ_JEDEC_ID	0x9f
 #define CMD_FREAD_DIO		0xbb
 
+#define STATUS_REG_BUSY_MASK	0x01
+#define STATUS_REG_WEL_MASK		0x02
+
 /*
  * Configure Memory Managed IO
  */
-void flash_configure_mmio(void)
+static void flash_configure_mmio()
 {
+	while(SFR_FLASH_EXEC_BUSY);
+
 	// Set configuration for MMIO access by controller
 	if (dio_enabled) {
 		SFR_FLASH_MODEB = 0x18;
@@ -44,6 +49,58 @@ void flash_configure_mmio(void)
 	SFR_FLASH_MODEB = 0x0;
 	SFR_FLASH_CMD_R = CMD_FREAD; // Default is Single IO
 	SFR_FLASH_DUMMYCYCLES = 8;
+}
+
+static void flash_configure_sio()
+{
+	while(SFR_FLASH_EXEC_BUSY);
+
+	// Set configuration for SIO access by controller
+	SFR_FLASH_MODEB = 0x0;
+	SFR_FLASH_DUMMYCYCLES = 0;
+}
+
+static uint8_t flash_read_status(void)
+{
+    uint8_t status;
+    uint8_t old_cmd_r;
+    uint8_t old_tconf;
+
+    while(SFR_FLASH_EXEC_BUSY);
+
+    // SAVE SPI peripheral state at entry
+    old_cmd_r = SFR_FLASH_CMD_R;
+    old_tconf = SFR_FLASH_TCONF;
+
+    SFR_FLASH_TCONF = 0x11;
+    SFR_FLASH_CMD_R = CMD_READ_STATUS; 
+    
+    SFR_FLASH_EXEC_GO = 1;
+    while(SFR_FLASH_EXEC_BUSY);
+    status = SFR_FLASH_DATA0;
+
+    // RESTORE SPI peripheral state
+    SFR_FLASH_CMD_R = old_cmd_r;
+    SFR_FLASH_TCONF = old_tconf;
+
+    return status;
+}
+
+static void flash_write_enable(void)
+{
+	short status;
+
+	while (flash_read_status() & STATUS_REG_BUSY_MASK);
+	
+	while (!(status & STATUS_REG_WEL_MASK)) {	
+		while(SFR_FLASH_EXEC_BUSY);
+
+		SFR_FLASH_TCONF = 0x18;
+		SFR_FLASH_CMD = CMD_WRITE_ENABLE;
+		
+		SFR_FLASH_EXEC_GO = 1;
+		status = flash_read_status();
+	}
 }
 
 
@@ -82,30 +139,12 @@ void flash_init(uint8_t enable_dio)
 	flash_configure_mmio();
 }
 
-
-uint8_t flash_read_status(void)
-{
-	// Test Controller Busy (we might call this directly after executing a command)
-	while(SFR_FLASH_EXEC_BUSY);
-
-	// setup status read command
-	SFR_FLASH_TCONF = 0x11;
-	SFR_FLASH_CMD_R = CMD_READ_STATUS;
-
-	// execute and wait for controller done
-	SFR_FLASH_EXEC_GO = 1;
-	while(SFR_FLASH_EXEC_BUSY);
-
-	return SFR_FLASH_DATA0;
-}
-
-
 void flash_read_uid(void)
 {
-	while (flash_read_status() & 0x1);
+	flash_configure_sio();
+	while (flash_read_status() & STATUS_REG_BUSY_MASK);
 
 	// Set slow read mode for UID
-	SFR_FLASH_MODEB = 0x0;
 	SFR_FLASH_CMD_R = CMD_READ_UNIQUE_ID;
 	SFR_FLASH_DUMMYCYCLES = 8;
 
@@ -152,12 +191,12 @@ __code char* get_flash_size_str(void)
 
 void flash_read_jedecid(void)
 {
-	while (flash_read_status() & 0x1);
+	flash_configure_sio();
+
+	while (flash_read_status() & STATUS_REG_BUSY_MASK);
 
 	// Set read mode for JEDEC ID
-	SFR_FLASH_MODEB = 0x0;
 	SFR_FLASH_CMD_R = CMD_READ_JEDEC_ID;
-	SFR_FLASH_DUMMYCYCLES = 0;
 
 	// Transfer 3 bytes back
 	SFR_FLASH_TCONF = 0x13;
@@ -180,32 +219,6 @@ void flash_read_jedecid(void)
 }
 
 
-void flash_write_enable(void)
-{
-	short status;
-
-	// Wait until busy bit clear
-	do {
-		status = flash_read_status();
-	} while (status & 0x1);
-
-	SFR_FLASH_TCONF = 0x18;
-	SFR_FLASH_CMD = CMD_WRITE_ENABLE;
-
-	/* The following makes sure that the PAGE_PROGRAM command,
-	 * where the data to be written follows the command word directly
-	 * works properly
-	 */
-	SFR_FLASH_DUMMYCYCLES = 0;
-	SFR_FLASH_MODEB = 0;
-
-	SFR_FLASH_EXEC_GO = 1;
-	// Wait for write status enabled
-	do {
-		status = flash_read_status();
-	} while (!(status & 0x2));
-}
-
 /*
  * Reads bulk data of length len from the flash memory starging at address src
  * and writes the data into a buffer pointed to by dst in XMEM
@@ -213,20 +226,9 @@ void flash_write_enable(void)
 void flash_read_bulk(__xdata uint8_t *dst)
 {
 	short status;
-	do {
-		status = flash_read_status();
-	} while (status & 0x1);
-
-	// Set fast read mode
-	if (dio_enabled) {
-		SFR_FLASH_MODEB = 0x18;
-		SFR_FLASH_CMD_R = CMD_FREAD_DIO;
-		SFR_FLASH_DUMMYCYCLES = 4;
-	} else {
-		SFR_FLASH_MODEB = 0x0;
-		SFR_FLASH_CMD_R = CMD_FREAD;	// Fast read
-		SFR_FLASH_DUMMYCYCLES = 8;	// Add 8 dummy clocks
-	}
+	flash_configure_sio();
+	while (flash_read_status() & STATUS_REG_BUSY_MASK);
+	flash_configure_mmio();
 
 
 	// Read 4 bytes
@@ -260,12 +262,12 @@ void flash_read_bulk(__xdata uint8_t *dst)
 
 void flash_read_security(void)
 {
-	while (flash_read_status() & 0x1);
+	flash_configure_sio();
+
+	while (flash_read_status() & STATUS_REG_BUSY_MASK);
 
 	// Set slow read mode
-	SFR_FLASH_MODEB = 0x0;
 	SFR_FLASH_CMD_R = CMD_READ_SECURITY_REGS;		// read security register
-	SFR_FLASH_DUMMYCYCLES = 8;	// Add 8 dummy clocks as for fast read
 
 	// Transfer 4 bytes (command + 3byte address)
 	SFR_FLASH_TCONF = 4;
@@ -298,6 +300,7 @@ void flash_read_security(void)
 
 void flash_sector_erase(void)
 {
+	flash_configure_sio();
 	flash_write_enable();
 	SFR_FLASH_TCONF = 8;
 	SFR_FLASH_CMD = CMD_SECTOR_ERASE;
@@ -307,7 +310,7 @@ void flash_sector_erase(void)
 	SFR_FLASH_ADDR0 = flash_region.addr;
 
 	SFR_FLASH_EXEC_GO = 1;
-	while (flash_read_status() & 0x1);
+	while (flash_read_status() & STATUS_REG_BUSY_MASK);
 
 	flash_configure_mmio();
 }
@@ -315,33 +318,39 @@ void flash_sector_erase(void)
 
 void flash_write_bytes(__xdata uint8_t *ptr)
 {
-	// write_char('\n'); write_char('>'); print_long(flash_region.addr); write_char(':'); print_short(flash_region.len); write_char('-'); print_byte(*ptr); // write_char('\n');
-	while(1) {
-		flash_write_enable();
-		SFR_FLASH_CMD = CMD_PAGE_PROGRAM;
-		SFR_FLASH_TCONF = 0x40 | 8 | 4; // Bytes written is 4, 8 enables write, 0x40 is unknown
+    flash_configure_sio();
+    
+    while(1) {
+        flash_write_enable();
+        SFR_FLASH_CMD = CMD_PAGE_PROGRAM;
+        
 		// Last transfer?
-		if (flash_region.len < 5) {
-			SFR_FLASH_TCONF = 8 | flash_region.len;
-		}
+        if (flash_region.len < 5) {
+            SFR_FLASH_TCONF = 8 | flash_region.len;
+        } else {
+            SFR_FLASH_TCONF = 0x40 | 8 | 4;  // Bytes written is 4, 8 enables write, 0x40 is unknown
+        }
 
-		SFR_FLASH_ADDR16 = flash_region.addr >> 16;
-		SFR_FLASH_ADDR8 = flash_region.addr >> 8;
-		SFR_FLASH_ADDR0 = flash_region.addr;
-		SFR_FLASH_DATA0 = *ptr++;
-		SFR_FLASH_DATA8 = *ptr++;
-		SFR_FLASH_DATA16 = *ptr++;
-		SFR_FLASH_DATA24 = *ptr++;
+        SFR_FLASH_ADDR16 = flash_region.addr >> 16;
+        SFR_FLASH_ADDR8 = flash_region.addr >> 8;
+        SFR_FLASH_ADDR0 = flash_region.addr;
+        
+        // Safely load only the valid bytes remaining in the buffer
+        SFR_FLASH_DATA0 = *ptr++;
+        if (flash_region.len > 1) SFR_FLASH_DATA8 = *ptr++;
+        if (flash_region.len > 2) SFR_FLASH_DATA16 = *ptr++;
+        if (flash_region.len > 3) SFR_FLASH_DATA24 = *ptr++;
 
-		// Execute transfer, we wait for completion at top of loop
-		SFR_FLASH_EXEC_GO = 1;
+        // Execute transfer, wait for completion at top of loop
+        SFR_FLASH_EXEC_GO = 1;
 
-		if (flash_region.len < 5)
-			break;
+        if (flash_region.len < 5)
+            break;
 
-		flash_region.len -= 4;
-		flash_region.addr += 4;
-	};
-	while (flash_read_status() & 0x1);
-	flash_configure_mmio();
+        flash_region.len -= 4;
+        flash_region.addr += 4;
+    };
+    
+    while (flash_read_status() & STATUS_REG_BUSY_MASK);
+    flash_configure_mmio();
 }
--- a/rtl837x_flash.h
+++ b/rtl837x_flash.h
@@ -13,7 +13,6 @@
 
 void flash_init(uint8_t enable_dio);
 void flash_read_uid(void);
-void flash_write_enable(void);
 void flash_dump(uint8_t len);
 void flash_read_jedecid(void);
 void flash_read_security(void);


logicog:
What about adding a preprocessor flag such as `FLASH_STATUS_SIO` which can be defined in `rtl837x_flash.c` and define 2 versions of `flash_read_status(void)` for the 2 cases where that flag is either set or not set. That flag could be defined based on the MACHINE variable. It would still require you to come up with a different machine variable for your device.

On the other hand it is not really optimal to allow for custom machines which someone has hacked, as this may multiply the machines supported quickly. So, the other solution would be if you were able to identify the flash without using MMIO status. That way a flag could be set that chooses either an SIO or MMIO path in `flash_read_status(void)`. Or, set say a `status_mmio = false` flag at boot time for everyone and only when the flash is positively identified to support status with mmio, switch that flag to `true` and take a different code-path in `flash_read_status(void)`.

feelfree69:
I agree with @logicog .
We should not change the way of working for the vast majority of devices which have no issue so far.

vDorst:
May this is not helpful, but I look sometime in the github network graphs to see that other forks are doing. I have seen one commit about that delaycycle maybe left on the previous state by an other command. But I can't find it in the graph. Maybe it is not the issue.

vDorst:
Found it https://github.com/xlon72/RTLPlayground/commit/5f814ebd0d09b2e3e4d53bf9e05375fc0610a8fa
Related, early commit on the [main-banch](https://github.com/xlon72/RTLPlayground/commits/main/) https://github.com/xlon72/RTLPlayground/commit/86effc9e64d0f8250e8d1d45adcaa2f1567e2ade

bennydiamond:
> Found it [xlon72@5f814eb](https://github.com/xlon72/RTLPlayground/commit/5f814ebd0d09b2e3e4d53bf9e05375fc0610a8fa) Related, early commit on the [main-banch](https://github.com/xlon72/RTLPlayground/commits/main/) [xlon72@86effc9](https://github.com/xlon72/RTLPlayground/commit/86effc9e64d0f8250e8d1d45adcaa2f1567e2ade)

That's pretty much what I'm doing if his call to `flash_configure_mmio` at the end do indeed reinstate MMIO correctly.

> What about adding a preprocessor flag such as `FLASH_STATUS_SIO` which can be defined in `rtl837x_flash.c` and define 2 versions of `flash_read_status(void)` for the 2 cases where that flag is either set or not set. That flag could be defined based on the MACHINE variable. It would still require you to come up with a different machine variable for your device.

This isn't exclusively a machine-related issue as it's closely tied to the SPI NOR flash device used. While it might be true that factory production runs of each hardware variants are probably all single runs and thus a specific machine variant would always be populated with the same flash chip, we cannot exclude the possibility of a BOM revision with a flash chip device that will exhibit this issue.

I had previously considered going in with a compile time flag or specific identifier in machine struct population but decided against for the reason I just explained.

My rationale is that all SPI NOR flash should support Read Status command in SingleIO mode and performance hit of downgrading from MMIO to SIO for this single operation wasn't all that bad for normal day to day operation. Favorising compatibility over performance.


> On the other hand it is not really optimal to allow for custom machines which someone has hacked, as this may multiply the machines supported quickly. So, the other solution would be if you were able to identify the flash without using MMIO status. That way a flag could be set that chooses either an SIO or MMIO path in `flash_read_status(void)`. Or, set say a `status_mmio = false` flag at boot time for everyone and only when the flash is positively identified to support status with mmio, switch that flag to `true` and take a different code-path in `flash_read_status(void)`.

I admit my case is pretty rare at the moment. I do like the auto detect approach where Read status is tried in MMIO first but with a thight timeout. If it fails, it reverts to SIO mode for read status operation by raising a runtime flag.

bennydiamond:
I have something that starts trying in MMIO mode and will fallback to SIO on failure.
Stick to SIO from that point on, until reboot/power cycle.

rtl837x_flash.c
```diff
@@ -28,6 +29,90 @@ __xdata uint8_t flash_capacity_code;
...
+	else
+		return flash_read_status_mmio();
+}
+
+static void flash_wait_busy(void)
+{
+	uint16_t i = 0;
+	while (flash_read_status() & 0x1) {
+		// Wait until busy bit clear
+		if(++i > 5000) {
```
logicog: I have a feeling that i++ may generate better code on 8051. Could you also say something on the choice of 5000? Have you tested that, and when does it usually timeout?

rtl837x_flash.c
```diff
@@ -28,6 +29,90 @@ __xdata uint8_t flash_capacity_code;
...
+		}
+	}
+}
+
+static void flash_wait_write_latch_enable(void)
+{
+	uint8_t status;
+	do {
+		status = flash_read_status();
+	} while (!(status & 0x2));
```
logicog: We could use the chance and give the bit 0x2 a name, something like FLASH_STATUS_WEL from Write Enable Latch (WEL).

rtl837x_flash.c
```diff
@@ -28,6 +29,90 @@ __xdata uint8_t flash_capacity_code;
...
+	if(sio_compat_busy_flag_read)
+		return flash_read_status_sio();
+	else
+		return flash_read_status_mmio();
+}
+
+static void flash_wait_busy(void)
+{
+	uint16_t i = 0;
+	while (flash_read_status() & 0x1) {
```
logicog: We could use the chance and give the bit 0x1 a name, something like FLASH_STATUS_BUSY or so.

logicog (commented):
Details...

logicog:
Tested on a KP_9000_6XHML_X2 and KP_9000_9XH_X_EU, and it works.

rtl837x_flash.c
bennydiamond: I was curious. One isn't better than the other. I changed it anyway.

```
                                   358 ;	rtl837x_flash.c:84: if(i++ > 5000) {
      002FAB 8E 04            [24]  359 	mov	ar4,r6
      002FAD 8F 05            [24]  360 	mov	ar5,r7
      002FAF 0E               [12]  361 	inc	r6
      002FB0 BE 00 01         [24]  362 	cjne	r6,#0x00,00130$
      002FB3 0F               [12]  363 	inc	r7
      002FB4                        364 00130$:
      002FB4 C3               [12]  365 	clr	c
      002FB5 74 88            [12]  366 	mov	a,#0x88
      002FB7 9C               [12]  367 	subb	a,r4
      002FB8 74 13            [12]  368 	mov	a,#0x13
      002FBA 9D               [12]  369 	subb	a,r5
      002FBB 50 DE            [24]  370 	jnc	00103$
                                    371 ;	free result
```

and:

```
                                    358 ;	rtl837x_flash.c:84: if(++i > 5000) {
      002FAB 0E               [12]  359 	inc	r6
      002FAC BE 00 01         [24]  360 	cjne	r6,#0x00,00130$
      002FAF 0F               [12]  361 	inc	r7
      002FB0                        362 00130$:
      002FB0 8E 04            [24]  363 	mov	ar4,r6
      002FB2 8F 05            [24]  364 	mov	ar5,r7
      002FB4 C3               [12]  365 	clr	c
      002FB5 74 88            [12]  366 	mov	a,#0x88
      002FB7 9C               [12]  367 	subb	a,r4
      002FB8 74 13            [12]  368 	mov	a,#0x13
      002FBA 9D               [12]  369 	subb	a,r5
      002FBB 50 DE            [24]  370 	jnc	00103$
                                    371 ;	free result
```

Basically all the same instructions but `mov	ar4,r6` and `mov	ar5,r7` are called at different times.

bennydiamond:
Serial log output for a unit that falls back to SIO

> 
> Initializing Flash controller
> Detecting CPU: RTL8373
> Software version: v0.1.0-0c0ae42-dirty
> Build date: 2026-09-10 04:43:35
> Hardware: Lianguo HYWS-SGT0108S
> NIC reset
> Setting MAC to: 06:09:15:25:df:2b
> 
> init_switch called
> CPU revision: 6868
> leds_setup called
> leds_setup done
> RTL837X_PIN_MUX_0: 0x00000000
> RTL837X_REG_LED_GLB_IO_EN: 0x47ffffff
> RTL837X_REG_LED1_0_SET0: 0x00410175
> RTL837X_REG_LED3_2_SET0: 0x00000000
> RTL837X_REG_LED1_0_SET1: 0x00000175
> RTL837X_REG_LED3_2_SET1: 0x00000000
> RTL837X_REG_LED1_0_SET2: 0x00000000
> RTL837X_REG_LED3_2_SET2: 0x00000000
> RTL837X_REG_LED1_0_SET3: 0x00000000
> RTL837X_REG_LED3_2_SET3: 0x00000000
> RTL837X_REG_LED3_0_SET1: 0x00010000
> RTL837X_REG_LED3_0_SET3: 0x00000000
> RTL837X_LED_PORT_SET_SEL: 0x00010000
> RTL837X_REG_LED_GLB_MUX_1: 0x05102040
> RTL837X_REG_LED_GLB_MUX_2: 0x0c289206
> RTL837X_REG_LED_GLB_MUX_3: 0x1245038d
> RTL837X_REG_LED_GLB_MUX_4: 0x19616554
> RTL837X_REG_LED_GLB_MUX_5: 0x2079d71a
> RTL837X_REG_LED_GLB_MUX_6: 0x000238a1
> RTL837X_REG_LED_GLB_ACTIVE: 0x3ffb6dff
> RTL837X_REG_LED_MODE: 0x0021e6b0
> LED pad Configuration:
> 00 01 02 03 04 05 06 07 08 09 0a 0b 0c 0d 0e 0f 10 11 12 13 14 15 16 17 18 19 1a 1b 
> 00 01 02 04 05 06 08 09 0a 0c 0d 0e 10 11 12 14 15 16 18 19 1a 1c 1d 1e 20 21 22 23 
> LED-set Configuration:
> LED-ID          0               1               2               3
> SET 0:     000175          000041          000000          000000
> SET 1:     010175          000000          000000          000000
> SET 2:     000000          000000          000000          000000
> SET 3:     000000          000000          000000          000000
> Port 0: SET 0: ( 2G5 1G 100M 10M LINK ACT), ( 2G5 LINK), (), (), 
> Port 1: SET 0: ( 2G5 1G 100M 10M LINK ACT), ( 2G5 LINK), (), (), 
> Port 2: SET 0: ( 2G5 1G 100M 10M LINK ACT), ( 2G5 LINK), (), (), 
> Port 3: SET 0: ( 2G5 1G 100M 10M LINK ACT), ( 2G5 LINK), (), (), 
> Port 4: SET 0: ( 2G5 1G 100M 10M LINK ACT), ( 2G5 LINK), (), (), 
> Port 5: SET 0: ( 2G5 1G 100M 10M LINK ACT), ( 2G5 LINK), (), (), 
> Port 6: SET 0: ( 2G5 1G 100M 10M LINK ACT), ( 2G5 LINK), (), (), 
> Port 7: SET 0: ( 2G5 1G 100M 10M LINK ACT), ( 2G5 LINK), (), (), 
> Port 8: SET 1: ( 2G5 1G 100M 10M LINK ACT 10G), (), (), (), 
> 
> rtl8373_init called
> 
> sds_init done
> 
> phy_config_8224 called
> RTL8224 ID: 0x82240000
> R02f8-00000bff R02f4-00000bed 
> phy_config_8224 done
> 
> RTL837X_REG_SDS_MODES: 0x00000bff
> 
> RTL837X_REG_SDS_MODES: 0x00000bff
> sds_config sds: 00, mode: 0d
> 
> RTL837X_REG_SDS_MODES: 0x00000bed
> 
> rtl8224_phy_enable called
> R02f8-00000003 R02f4-0000000c 
> rtl8224_phy_enable done
> EEE on for 00 speed 2g5
> EEE on for 01 speed 2g5
> EEE on for 02 speed 2g5
> EEE on for 03 speed 2g5
> EEE on for 04 speed 2g5
> EEE on for 05 speed 2g5
> EEE on for 06 speed 2g5
> EEE on for 07 speed 2g5
> EEE can't be enabled for SFP port 08
> 
> rtl8373_init done
> WARNING: SPI Flash MMIO failure detected. Falling back to SPI Flash SIO Mode: 01 00
> Flash information:
>   Manufacturer ID: 0xef
>   Memory Type:     0x40
>   Capacity:        0x14 = 1 MB
> 1 MB flash size detected. (1 MB is needed for image updating)
> Checking for update image in flash... no update image found.
> 
> vlan_setup called 
> vlan_setup done 
> 
> port_l2_setup called
> 
> port_l2_forget called
> port_l2_forget done
> 
> port_l2_setup done
> igmp_setup called
> bandwidth_setup called
> RTL837X_IGBW_CTRL: 0x0000003f
> RTL837X_EGBW_CTRL: 0x00000001
> bandwidth_setup done
> uip_arp_init called
> Configuring I2C for SFP idx=00 SCL=03, SDA=04
> 
> A minimal prompt to explore the RTL8372:
> 
> Clock register: 0x00001101
> Register 0x7b20/RTL837X_REG_SDS_MODES: 0x00000bed
> Verifying PHY settings:
> 
> Port    State   Link    TxGood          TxBad           RxGood          RxBad
> 1       On      Down    0x00000000      0x00000000      0x00000000      0x00000000
> 2       On      1000M   0x00000000      0x00000000      0x00000003      0x00000000
> 3       On      Down    0x00000000      0x00000000      0x00000000      0x00000000
> 4       On      Down    0x00000000      0x00000000      0x00000000      0x00000000
> 5       On      Down    0x00000000      0x00000000      0x00000000      0x00000000
> 6       On      Down    0x00000000      0x00000000      0x00000000      0x00000000
> 7       On      Down    0x00000000      0x00000000      0x00000000      0x00000000
> 8       On      Down    0x00000000      0x00000000      0x00000000      0x00000000
> 9       NO SFP  Down    0x00000000      0x00000000      0x00000000      0x00000000
> Setting ip: 192.168.0.7
> Setting gw: 192.168.0.3
> Setting netmask: 255.255.254.0
> 
> 
> > 
> <new link: 2000000020, was 0000000000>
> 

Notice the line `WARNING: SPI Flash MMIO failure detected. Falling back to SPI Flash SIO Mode: 01 00
` close to middle.

bennydiamond:
Log from a device that do not fallback to SIO mode

> Initializing Flash controller
> Detecting CPU: RTL8372
> Software version: v0.1.0-0c0ae42-dirty
> Build date: 2026-09-10 10:16:14
> Hardware: Lianguo ZX-SWTGW215AS
> NIC reset
> Setting MAC to: 06:1b:1a:15:78:23
> 
> init_switch called
> CPU revision: 6868
> leds_setup called
> leds_setup done
> RTL837X_PIN_MUX_0: 0x38000000
> RTL837X_REG_LED_GLB_IO_EN: 0x77ffff00
> RTL837X_REG_LED1_0_SET0: 0x00000141
> RTL837X_REG_LED3_2_SET0: 0x00000174
> RTL837X_REG_LED1_0_SET1: 0x00000155
> RTL837X_REG_LED3_2_SET1: 0x00000000
> RTL837X_REG_LED1_0_SET2: 0x00000000
> RTL837X_REG_LED3_2_SET2: 0x00000000
> RTL837X_REG_LED1_0_SET3: 0x00000000
> RTL837X_REG_LED3_2_SET3: 0x00000000
> RTL837X_REG_LED3_0_SET1: 0x00050000
> RTL837X_REG_LED3_0_SET3: 0x00000000
> RTL837X_LED_PORT_SET_SEL: 0x00010000
> RTL837X_REG_LED_GLB_MUX_1: 0x3fffffff
> RTL837X_REG_LED_GLB_MUX_2: 0x0c3fffff
> RTL837X_REG_LED_GLB_MUX_3: 0x1245038d
> RTL837X_REG_LED_GLB_MUX_4: 0x19616554
> RTL837X_REG_LED_GLB_MUX_5: 0x2079d71a
> RTL837X_REG_LED_GLB_MUX_6: 0x000238a1
> RTL837X_REG_LED_GLB_ACTIVE: 0x3ffbedff
> RTL837X_REG_LED_MODE: 0x0021e6b0
> LED pad Configuration:
> 00 01 02 03 04 05 06 07 08 09 0a 0b 0c 0d 0e 0f 10 11 12 13 14 15 16 17 18 19 1a 1b 
> 3f 3f 3f 3f 3f 3f 3f 3f 0f 0c 0d 0e 10 11 12 14 15 16 18 19 1a 1c 1d 1e 20 21 22 23 
> LED-set Configuration:
> LED-ID          0               1               2               3
> SET 0:     000141          000000          000174          000000
> SET 1:     050155          000000          000000          000000
> SET 2:     000000          000000          000000          000000
> SET 3:     000000          000000          000000          000000
> Port 3: SET 0: ( 2G5 LINK ACT), (), ( 1G 100M 10M LINK ACT), (), 
> Port 4: SET 0: ( 2G5 LINK ACT), (), ( 1G 100M 10M LINK ACT), (), 
> Port 5: SET 0: ( 2G5 LINK ACT), (), ( 1G 100M 10M LINK ACT), (), 
> Port 6: SET 0: ( 2G5 LINK ACT), (), ( 1G 100M 10M LINK ACT), (), 
> Port 7: SET 0: ( 2G5 LINK ACT), (), ( 1G 100M 10M LINK ACT), (), 
> Port 8: SET 1: ( 2G5 1G 100M LINK ACT 10G 5G), (), (), (), 
> 
> rtl8372_init called
> 
> sds_init done
> 
> phy_config: 8
>   phy config done
> 
> phy_config: 3
>   phy config done
> EEE on for 03 speed 2g5
> EEE on for 04 speed 2g5
> EEE on for 05 speed 2g5
> EEE on for 06 speed 2g5
> EEE on for 07 speed 2g5
> EEE can't be enabled for SFP port 08
> 
> rtl8372_init done
> Flash information:
>   Manufacturer ID: 0xef
>   Memory Type:     0x40
>   Capacity:        0x14 = 1 MB
> 1 MB flash size detected. (1 MB is needed for image updating)
> Checking for update image in flash... no update image found.
> 
> vlan_setup called 
> vlan_setup done 
> 
> port_l2_setup called
> 
> port_l2_forget called
> port_l2_forget done
> 
> port_l2_setup done
> igmp_setup called
> bandwidth_setup called
> RTL837X_IGBW_CTRL: 0x0000003f
> RTL837X_EGBW_CTRL: 0x00000001
> bandwidth_setup done
> uip_arp_init called
> Configuring I2C for SFP idx=00 SCL=03, SDA=04
> 
> A minimal prompt to explore the RTL8372:
> 
> Clock register: 0x00001101
> Register 0x7b20/RTL837X_REG_SDS_MODES: 0x00000bff
> Verifying PHY settings:
> 
> Port    State   Link    TxGood          TxBad           RxGood          RxBad
> 5       On      1000M   0x00000000      0x00000000      0x00000000      0x00000000
> 1       On      2.5G    0x00000062      0x00000000      0x00000042      0x00000000
> 2       On      1000M   0x00000041      0x00000000      0x0000003f      0x00000000
> 3       On      1000M   0x00000082      0x00000000      0x0000002d      0x00000000
> 4       On      Down    0x00000000      0x00000000      0x00000000      0x00000000
> 6       SFP IN  Down    0x00000000      0x00000000      0x00000000      0x00000000
> Setting ip: 192.168.0.4
> Setting gw: 192.168.0.3
> Setting netmask: 255.255.254.0
> 
> Guests<
> 
> vlan_create called
> vlan: 0x03e9, members: 0x0300, tagged: 0x0300
> vlan_create done 
> vlan_ptr 0x000a:>3e9Guests <
> IotCloud<
> 
> vlan_create called
> vlan: 0x03ea, members: 0x0300, tagged: 0x0300
> vlan_create done 
> vlan_ptr 0x0016:>3e9Guests 3eaIotCloud <
> IotMixed<
> 
> vlan_create called
> vlan: 0x03eb, members: 0x0370, tagged: 0x0370
> vlan_create done 
> vlan_ptr 0x0022:>3e9Guests 3eaIotCloud 3ebIotMixed <
> 
> > 
> <new link: 2002252000, was 0000000000>
> sds_config sds: 00, mode: 02
> 
> RTL837X_REG_SDS_MODES: 0x000003e2
> 
> <MODULE INSERTED>  Slot: 1  Rate: 67  Encoding: 06  Module: FINISAR CORP.   FCBG110SD1C05-1RA   
> 
> sds_config sds: 01, mode: 1a
> 
> RTL837X_REG_SDS_MODES: 0x00000342
> 
> <SFP-RX OK>  Slot: 1
> 
> <new link: 2402252000, was 2002252000>

No warning line above `Flash information:`

bennydiamond:
I'd also like to point out  both device use the same SPI Flash chip, both from the same reel.

Flash information:
Manufacturer ID: 0xef
Memory Type: 0x40
Capacity: 0x14 = 1 MB

Silicon revision issue in the RTL chip?
The Lianguo HYWS-SGT0108S I have (that exhibit the issue) was purchased Jun 9, 2023. 
The Lianguo ZX-SWTGW215AS switch (which works fine) was purchased on Nov 27, 2023.

feelfree69:
> Silicon revision issue in the RTL chip?

I guess the difference comes from the external circuitry between SoC and Flash (serial, pull-up or pull-down-resistors)

Maybe you can even measure them?

I have quickly measured the resistance between GND/VCC and the IOs on my SWTGW218AS: 
IO0 / IO1 have no pullups or pulldowns
IO2 5kOhms PullUp
IO3 10kOhms PullUp

Pure guessing: On the failing switch, there's also a Pullup on IO0.

bennydiamond:
> > Silicon revision issue in the RTL chip?
> 
> I guess the difference comes from the external circuitry between SoC and Flash (serial, pull-up or pull-down-resistors)
> 
> Maybe you can even measure them?
> 
> I have quickly measured the resistance between GND/VCC and the IOs on my SWTGW218AS: IO0 / IO1 have no pullups or pulldowns IO2 5kOhms PullUp IO3 10kOhms PullUp
> 
> Pure guessing: On the failing switch, there's also a Pullup on IO0.

All 6 logic pins have 10K pullups on them. While not ideal I admit, especially on CLK, it does not explain why I read status reg at 0x01 in DIO mode and when I switch to SIO, I get 0x00. I would understand the pullup would be an issue if I had a recurring pattern on like 0x55 or 0xAA.

feelfree69:
Maybe the situation without your patch is even worse:

There's no such thing as a STATUS_READ cmd for Dual IO operation. But the existing code assumes there is, and reads the data on both lines, IO0 and IO1, with 4 dummycycles. Bit0 lands on IO0, which is left floating by the flash, and with the Pullup the SoC reads 1.

That doesn't explain why bits 2, 4 and 6 do not also read a 1....

But at the moment, I think that your initial patch, without fallback, but unconditional reading in sio mode, is the correct solution.

Will do some measurements at the weekend.

bennydiamond:
> But at the moment, I think that your initial patch, without fallback, but unconditional reading in sio mode, is the correct solution.
> 
> Will do some measurements at the weekend.

Honestly, I think so too. I was surprised that the current implementation was reading status reg in MMIO configuration. Did not question it but now that this whole thing blew up, it makes sense why it seemingly worked with MMIO enabled but is technically improperly implemented.

feelfree69:
Even better: I think your change additionally eliminates the need for an additional hack: Before writing the update image or the config, `flash_init(0)` is called, which disables DIO operation. 
Reason for that was, with DIO operation flashing did not work but firmware got stuck. I bet it got stuck in `flash_write_enable`, which itself calls `flash_read_status` and waits probably forever for the WriteEnable bit to be set.

feelfree69:
Did a quick test by adding a few commandline cmds:

```
> flash s

Flash Status: 00

> flash t

Flash Status (Single IO): 00

> flash w

Setting Flash Write Enable

> flash s

Flash Status: 08

> flash t

Flash Status (Single IO): 02
```

`flash s` performs is the old command in dio operation
`flash t` does the same, but in single io operation. - it uses the implementation of your first commit and restores dio operation after performing the read status
`flash w` calls a modified 'flash_write_enable: it does not wait for the WREN status, and also restores the dio mode after performing the command

It can be clearly seen that your existing, first implementation for the `flash_read_status()` is the correct one.
And also `flash_write_enable()` needs to be fixed the same way.

bennydiamond:
Write enable bit set function might be affected by the same MMIO/SIO issue.

One of the cleaner approach would be to set SPI peripheral in SIO by default and set in MMIO only when about to read data off the flash. But this only an optimal solution is the RTL8372 MCU does not do XIP. Which I don't know if its the case since there's no documentation on this chip.

The next best thing would be the opposite. Set SPI peripheral in n DIO mode by default and temporary set in SIO when doing any operations other than data read.

There's more to be done for the moment so I'm removing this PR of the review stack.

bennydiamond:
Ok, I reworked the whole driver file quite a bit. But the gist of it is to keep the DIO state as much as possible as I don't have any info on if the RTL MCU is XIP or not. So better be safe than sorry and ensure fastest code execution possible.

When any manual SPI flash operation is requested, other than data read, the driver temporarily switches to SingleIO mode for the duration of the operation and then is set back to DIO.

rtl837x_flash.c
```diff
@@ -46,6 +51,58 @@ void flash_configure_mmio(void)
...
+    // RESTORE SPI peripheral state
+    SFR_FLASH_CMD_R = old_cmd_r;
+    SFR_FLASH_TCONF = old_tconf;
+
+    return status;
+}
+
+static void flash_write_enable(void)
+{
+	short status;
```
feelfree69: uint8_t ?

rtl837x_flash.c
```diff
@@ -46,6 +51,58 @@ void flash_configure_mmio(void)
...
+
+	while (flash_read_status() & STATUS_REG_BUSY_MASK);
+	
+	while (!(status & STATUS_REG_WEL_MASK)) {
```
feelfree69: status read before written?

feelfree69 (commented):
Thanks, I think this is the right way to go.
Will test later.
2 small comments, just from reading with my mobile, so I may be wrong.

rtl837x_flash.c
bennydiamond: I don't get it. Status bit is validated cleared before trying to set Write Enable Latch bit.

feelfree69: status is not initialized in line 91.
And read in line 95.

vDorst:
Also, when a function doesn't has a argument, then use `void` to prevent compiler warnings like
```
rtl837x_flash.c:37: warning 283: function declarator with no prototype
rtl837x_flash.c:54: warning 283: function declarator with no prototype
```

```patch
diff --git a/rtl837x_flash.c b/rtl837x_flash.c
index 4845eeb..e96230f 100644
--- a/rtl837x_flash.c
+++ b/rtl837x_flash.c
@@ -34,7 +34,7 @@ __xdata uint8_t flash_capacity_code;
 /*
  * Configure Memory Managed IO
  */
-static void flash_configure_mmio()
+static void flash_configure_mmio(void)
 {
        while(SFR_FLASH_EXEC_BUSY);
 
@@ -51,7 +51,7 @@ static void flash_configure_mmio()
        SFR_FLASH_DUMMYCYCLES = 8;
 }
 
-static void flash_configure_sio()
+static void flash_configure_sio(void)
 {
        while(SFR_FLASH_EXEC_BUSY);
```

rtl837x_flash.c
```diff
@@ -46,6 +51,59 @@ void flash_configure_mmio(void)
...
+
+    while ((status = flash_read_status()) & STATUS_REG_BUSY_MASK);
+
+	while (SFR_FLASH_EXEC_BUSY);
+    while (!(status & STATUS_REG_WEL_MASK)) {
+
+        SFR_FLASH_TCONF = 0x18;
+        SFR_FLASH_CMD = CMD_WRITE_ENABLE;
+
+        SFR_FLASH_EXEC_GO = 1;
```
feelfree69: I don't think it is needed to repeat the CMD_WRITE_ENABLE. Just fire it once and poll for the WEL bit.

bennydiamond: It's debatable.
If for some reason the SPI flash does not take in the command, you're locked in the while loop waiting for WEL bit to raise.

I can remove it if you want. To be fair, I don't see why the spi flash wouldn't raise the WEL bit on first try. But you never know...

feelfree69: > I don't see why the spi flash wouldn't raise the WEL bit on first try.

Exactly. It is debatable if polling the WEL bit improves anything, since it will be set probably immediately anyway. 

But ignoring a command (here CMD_WRITE_ENABLE) would be clearly out of spec. If that could be the case, you could also argue to read the contents of an erased sector to see if it's really empty, and repeat the CMD_SECTOR_ERASE if not.

bennydiamond: > But ignoring a command (here CMD_WRITE_ENABLE) would be clearly out of spec. If that could be the case, you could also argue to read the contents of an erased sector to see if it's really empty, and repeat the CMD_SECTOR_ERASE if not.

I would do that honestly. Data validation is crucial. I would implement readback on write/erase operations and retry as long as necessary because in anycase, if the firmware data you just wrote isn't right, you end up with a bricked device.

But I think implementing readback capabilities is out of scope for this PR.

I'll do what is asked of me. If you want me to remove the while loop on write enable set I'll do it no problem. But I think it's a safe way to acheive it without noticeable negative impact.

feelfree69: Then let some other reviewers ask for their opinion.
For me it doesn't make sense to repeat a command that is not supposed to fail, as long as the device is not defective.
And if it is defective, it's not up to the firmware to find a workaround.

bennydiamond: When you're putting it that way it makes sense.
Removed the while loop that would send multiple times WEL bit set.

feelfree69 (approved):
Code looks good now. Tested on KP_9000_6XH_X2, saving config and update image worked without issue.

vDorst:
Thanks

PR #428, opened 2026-09-09
httpd: count every authenticated request as session activity

bloqaudio:
The session timer was refreshed in two places only: at login, and when a page from the file table was served. Every JSON endpoint and `/cmd` left it alone. With the old multi-page UI that was invisible, because each navigation loaded a page. With the single-page UI from #315 the pages are fetched once and everything after that is JSON, so the session expires `SESSION_TIMEOUT` (200) seconds after the page load however busy the user is, and the next request is answered with 401. A firmware upload started after that point gets the 401 instead of a checksum verdict, which is how I ran into it.

The change refreshes the timer wherever the session cookie validates in `scan_header()`, which covers every authenticated request, and drops the page-serving refresh that this makes redundant, together with the `timeptr` scratch pointer it used. Three lines in, seven out.

Measured on `PCB_SWTG018AS_V2_1_0`: internal RAM map identical to `main`, BANK1 66 bytes smaller, xdata 2 bytes smaller, `make machine_check` clean. On the switch: after login, polling `/information.json` every 30 s stays authenticated past 240 s; before the change the same probe was logged out at 210 s. Unauthenticated requests still get their 401.

### DIFF
--- a/httpd/httpd.c
+++ b/httpd/httpd.c
@@ -76,7 +76,6 @@ __xdata uint8_t fw_reset_pending;
 __xdata char session_id[SESSION_ID_LENGTH + 1];
 __xdata uint8_t authenticated;
 __xdata uint32_t now;
-__xdata uint8_t * __xdata timeptr;
 __xdata uint32_t last_session_use;
 
 #define TSTATE_NONE		0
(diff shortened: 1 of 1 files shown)


vDorst:
So the website never expires when it stays open, because the `status.json` fetch resets the timer every time.
That seems not ideal to me.

bloqaudio:
Agreed, an open tab should not keep the session alive forever. This PR's commit is already in main (it went in as the first commit of #429, ec1cb41), so I am closing this one; the fix for the never-expiring tab is #450: the page stops polling after ten minutes without input, the firmware session then times out as before, and the next input resumes polling and lands on the login page. That keeps the firmware rule simple (any authenticated request is activity, which the SPA needs so that a user who is clicking around is not logged out mid-upload) and puts the idle decision where the input events are.

PR #429, opened 2026-09-09
web: replace the multi-page UI with a themed single-page app

bloqaudio:
This is the UI from #336, rebased onto the #315 pipeline, on top of the session-timer fix in #428 (a single-page UI is logged out 200 s after load without it). One `index.html` with every section, one `app.js`, a self-contained `login.html` and the favicon; the stylesheet and the port images are gone (the CSS is inline, the port strip is drawn with CSS). The firmware side is a two-line change in `httpd.c`: the pre-login allowlist becomes `login.html` plus the favicon, because nothing else is served before authentication any more. `tools/httpd_sim.c` gets the same allowlist.

**What the UI does differently**

- One request queue that waits for the previous response body, not just its headers. The firmware shares one output buffer across connections and delimits responses by closing, so a second request in flight corrupts the first. Every fetch in the UI goes through this queue.
- "Save to flash" merges `/cmd_log` into `/config` with the same grammar the console accepts (the regex list in `app.js` mirrors `cmd_parser.c`), shows the merged file before writing, writes it, reads the sector back and only clears the command log if the read-back matches. The header shows an "unsaved changes" badge while the running configuration differs from flash, also across page loads (it reads the command log at start).
- The console section shows the command output the firmware returns, and the parser's explanation on a rejected command.
- The firmware update section checks size, bank header and CRC16 in the browser before uploading, shows upload progress, and waits for the switch to come back after the reset with the outcome (applied / rejected / no reboot seen).
- The ports table shows the one learned MAC behind a port, or the count when there are several (this is #335 re-homed). The MAC table sorts per column, filters, and names the LAG a learned address sits behind; no delete button on the CPU row (#393 behaviour kept).
- The spanning tree section is DrDoof's `/stp.json` model: bridge settings and per-port settings are edited in place and Apply sends only what changed, instead of one command per control event. Enabling and disabling asks first.
- LAG groups are numbered 1 to 4 as on the command line; applying an empty group sends `lag N d`.
- Confirmation dialogs sit in front of everything that can lock the operator out: management VLAN, IP change, DHCP, reboot, STP on/off, VLAN delete, LAG clear, MAC flush.
- Four themes (light, dark, Selenized light, Selenized dark) plus follow-the-OS, stored in `localStorage`.

**Translations**

English, Japanese and Chinese, as a `LANG` table at the top of `app.js` with the same `t()` pattern as before. The Japanese and Chinese strings that existed for the old pages were reused where the meaning matched; the rest is new and I would welcome corrections from native speakers. Static markup uses `data-i18n`, `data-i18n-t` (title) and `data-i18n-p` (placeholder); the language selector sits in the page header and reloads the page so every table is rebuilt in the new language. `doc/support-multi-language.md` describes it.

**Sizes** (`make MACHINE=PCB_SWTG018AS_V2_1_0`, compressed as embedded)

| file | main | this PR |
|---|---:|---:|
| `login.html` | 499 | 1816 |
| `index.html` | 3984 | 6953 |
| `app.js` / `main.js` | 19860 | 28990 |
| `favicon.ico` | 766 | 766 |
| `style.css` + 3 SVG | 3422 | 0 |
| total | 28531 | 38525 |

The html block ends at `0x4967d`, far below the config sector at `0x6f000`. Internal RAM map identical to `main` on all machines (`make machine_check` clean, 34 machines); BANK1 is 125 bytes smaller from the shorter allowlist; xdata unchanged.

**Verification**

- Every source file goes through `tools/minify.py` unchanged in behaviour: the browser test suite passes on the raw `html/` and on the minified output, and the embedded gzip entries decompress byte-identical to `output/html_min`.
- A mock of the httpd endpoints (JSON shapes from `page_impl.c`) plus a headless Chromium walk: login, every section, one action per section with the emitted `/cmd` bodies asserted, the save-to-flash merge (toggle pairs, `vlan N d`, `lag N d`, `bw` limit and mode lines kept apart), firmware image checks, three languages, four themes. 70 checks, no page errors.
- On the switch (PCB_SWTG018AS_V2_1_0): every section exercised against the command line on a bench unit; a firmware update started from the page itself (browser-side image check, upload, reboot wait, back at the login page); with #428 in the build, a session polled every 30 s stays logged in past 240 s where it was gone at 210 s before.

**Depends on** the session-refresh fix in `httpd.c` (separate PR / commit): without it any single-page client is logged out 200 s after loading, because only static page hits refreshed the session timer.

**Not included, on purpose**

- Forcing an SFP module's speed uses the existing `sfp <slot> <speed>` command from the ports table. If that is unwanted for now (#401 left it out of the old table), the row is easy to drop.
- Ingress filtering is write-only (the state is not readable from the switch); the section says so.
- No per-page files remain, so the old pages are not kept as a fallback build option.

Screenshots are the ones from #336 (the flow-control entry visible in that sidebar is a fork feature and not part of this PR):

Image
Image
Image
Image
Image
Image
Image


feelfree69:
Just tried it out and played around a bit - it's just AMAZING!
Thank you so much for your work!

vDorst:
I didn't to expect that this PR is already ready.
Although I had no real expectations.

But we have many pull request that adds or fixed things to the current website.
So I'm not sure what needs to be done first.
I see real fixes but I can't tell if they are still an issue for this PR.

@bloqaudio can you may check if the other PRs are still needed and have to be merge before or after this PR?

The points I found.

Tested in Firefox
- Left menu
  - The Default-hostname doesn't fit.
- Dashboard
  - I do miss the tooltip for the SFP-ports.
  - The SFP info windows doesn't update as before
- Ports
  - MTU field is to small 
- BW
  MBit/s are also to small
- Save to flash
  I would love to see that you can edit it, so you can remove/correct lines.
- Theme, autoselect darktheme when the browser signals it.  

Add a entry in the Changelog.md.

eraiza0816:
Some UI suggestions from testing on real hardware:

The header always shows "Save to flash", the language selector, and the theme selector, but these are rarely changed. Would it be better to show them only where system operations happen (e.g. the System tab), keeping the header minimal during daily use?

Also, on a ~24" screen the layout does not expand to fill the window, so scrolling is needed in places. A layout that adapts to wider screens without scrolling would feel much better.

Found a display-order bug on a remapped board (SWGT024, `phys_to_log = {4,5,6,7,8,3,...}`): 

the Bandwidth table lists ports as 6, 1, 2, 3, 4, 5, and the EEE table likewise starts with 6 (SFP).

bloqaudio:
@feelfree69 thank you, that is good to hear.

@vDorst Thanks for the pass. Everything below is on the branch now, verified against the endpoint mock and retested on my bench unit.

- **Default hostname in the sidebar**: clipped with an ellipsis, full name on hover.
- **SFP tooltip / SFP window not updating**: every port in the strip carries the detail rows as a hover tooltip again (name, state, counters, and for SFP ports the module and DOM readings), and the detail dialog is re-rendered on every status poll while it is open, so it ticks like the old one.
- **MTU and Mbit/s fields too small**: widened from 70 px to 7 em; five digits plus the Firefox spinner fit.
- **Save to flash editable**: the preview is a textarea now. Size and unknown-line warnings follow what you type, the write button greys out above 2048 bytes, and what is written and read back is the edited text.
- **Auto dark theme**: `Theme: System` (now `System` under System > Display) is the default and follows `prefers-color-scheme`, also on the login page and before first paint, and the mock suite now asserts both directions. If your switch came up light with the OS dark, two things I can think of: Firefox Settings > General > Website appearance set to Light overrides the OS for `prefers-color-scheme`, or the browser still had a `theme` pick in `localStorage` for the switch's address from an earlier build of #336. Picking `System` once clears the second.
- **Changelog**: entry added under Added / Web UI.

On the open web PRs, against what `app.js` does today:

| PR | status with this branch |
|---|---|
| #430 stat: no double status fetch | not applicable, the SPA polls `/status.json` once per tick |
| #431 l2: stop the FDB walk on leave | the same defect existed in `app.js`; fixed here (generation guard bumped by the leave hooks) |
| #432 vlan: keep every VLAN, fetch list once | `/vlanlist` was already fetched once per refresh; a failed `/vlan.json` now keeps the row with `?` in the port columns instead of dropping the rest of the table |
| #433 web: one request at a time | covered by the request queue in `app.js`, which waits for the previous body, not just its headers |
| #434 web: fetch port icons once | not applicable, no SVGs are served any more |
| #405 isolation page, #397 STP LAG entity (the `main.js` half), #299 LACP page, #265 PoE page, #409 Spanish | the firmware halves are unaffected; the `html/` halves target `main.js` and need redoing against `app.js` whichever order they merge in. If this goes in first I will port each of those pages as its own PR once its firmware half is merged; #409's table drops into `LANG` as a fourth language. If they go in first, I rebase and port them here. My preference is this one first so nothing is written twice. |

@eraiza0816 Thanks, and the SWGT024 report was the important one: the firmware emits the per-port arrays in logical order and the SPA drew them as received, so the ports, spanning tree, EEE and bandwidth tables came out as 6, 1, 2, 3, 4, 5 and the dashboard strip put the SFP marker on the wrong port. Every array is now sorted by `portNum` on receipt, and the mock suite runs once more with `MOCK_ORDER=6,1,2,3,4,5,7,8,9` to keep it that way. I have no remapped board here; if you can confirm on the SWGT024 that would close it properly.

Header: language and theme moved to a Display card in System; the header keeps the title, the unsaved-changes badge and Save to flash, since that one is the operational action. Wide screens: the main column now stops at 1400 px instead of 1100 px, so the dashboard cards sit side by side on a 24" screen. The tables are as long as the port count, so some scrolling stays.

Bench unit after the flash (`v0.1.0-6cb1e73`): the served `app.js`, `index.html` and `login.html` are byte-identical to the build output, all five per-port tables and the strip come out 1 to 9, the detail dialog ticks while open, the save preview is editable, the selectors sit in System, the theme follows the browser preference both ways, and leaving the MAC table stops the walk mid-way with no further `/l2.json` requests. With a module in the slot, the port 9 tooltip and dialog carry vendor, model, serial, temperature, Vcc, TX bias, TX/RX power and the LOS flags, and the temperature and TX power readings move between polls while the dialog is open.

One more small one from that check: a dark receiver reports 0 mW and printed `-Infinity dBm` (the old page did the same); it is floored at -40 dBm now, the lowest value the DOM field can express.

feelfree69:
Thanks for sorting that out and for comparing against the other Web-related PRs.

I don't know if it's only me, but I feel I lose track due to the number of PRs.
I would prefer to have this one merged first, and reject any new PRs which have a web part in it.
Solving the same issue within different PRs is just waste.

vDorst:
@bloqaudio that for going through the many issues.

I did tested it for a bit and the basic features are working.
I would like to propose just to merge this PR as is.
Iron out the bugs later.

Also we have a lot of other PRs that are rely on adding things to the web-component.

vDorst:
@bloqaudio can you rebase?
@logicog did merge an issue that touches the web-component.
Currently I am holding off all the issues that touches the web-components, until it is clear what we doing with this PR.

bloqaudio:
@vDorst rebased onto main 62c6c65 and force-pushed. The only conflict was `html/main.js`: this branch deletes it and #397 edited it, so the resolution is the deletion plus one new commit that ports the #397 GUI half to `app.js`. A LAG gets its own row in both spanning tree tables, labelled with its members (`LAG1 (7,8)`), the cells are keyed by entity rather than port number, the member ports leave the port rows as the firmware now emits them, and the tables rebuild when a group appears or changes, so a LAG created on the LAG tab shows up in the spanning tree section without a reload. `stp lag <n> ...` is recognised next to `stp port <n> ...` for the unsaved-changes badge and the startup config merge. The mock suite is at 93 checks: it now creates a LAG, checks both tables, applies a cost and edge change on the group row and asserts `stp lag 1 cost 10000` and `stp lag 1 edge off` go out, then clears the group; it passes raw, minified and with `MOCK_ORDER=6,1,2,3,4,5,7,8,9`. Bench unit (SWTG018AS, `v0.1.0-47d1725`) after the flash: served `app.js`, `index.html` and `login.html` byte-identical to the build, read-only walk 22/22, and with STP on and a group made of two idle ports the firmware lists the entity (`p` 0, `lag` 1, `mbr` 3), both tables read `3,4,5,6,7,8,9,LAG1 (1,2)`, the group row's apply sends `stp lag 1 edge off` and `stp lag 1 cost 12345` and both read back on the next poll; clearing the group brings the nine port rows back without a reload.

Two things on main I hit while building the rebased branch for the bench unit, both from #397 and both independent of this PR, each with its own PR: c1ad800 changed the parser keyword from `stp port` to `stp ent` while the doc, the web UI and saved startup configs still say `stp port` (#448); and BANK2 is 30 bytes over its 48 KB window on `PCB_SWTG018AS_V2_1_0`, so the image builder refuses to link main for that machine, which CI does not build (#449).

vDorst:
@bloqaudio  Thanks for rebasing.

I merge the two issues, you mentioned.
Is it ready to merge? Or do you still need to fix the `stp` commands support changes?

bloqaudio:
@vDorst yes, ready to merge as is. The `stp lag` handling is already in the branch, in the last commit, and the only firmware-side `stp` problem was the parser keyword, which was #448. Thanks for merging both. Merged with current main (b5b5277) the branch builds for the SWTG018AS, and that merged tree is the same content the bench unit has been running since this morning.

vDorst:
Thanks for all the work!

mintsoft:
@bloqaudio Where did you get the SVG icons from? I need a new one for lldp but I can't figure out where I can source one

bloqaudio:
@mintsoft They are not from a set. Each one is a hand-written `path` on a 24 by 24 grid, `stroke-width` 2, no fill, which is the convention Feather and Lucide use, so an icon from either of those (MIT and ISC) drops in as long as it is flattened to one `d` string: the tab renderer in `html/app.js` only emits `<path d="...">`, so a `rect`, `circle` or `polyline` from a library icon has to be rewritten as path commands. The tab icons sit in the `TABS` table near the top of `app.js`, and the sidebar colours them through `#navlist li svg{stroke:currentColor}` in `index.html`.

For LLDP a radio shape reads as "announcing to the neighbours", and this one renders cleanly at 16 px with the same stroke settings as the others:

```js
{id:"lldp", icon:"M11 12a1 1 0 002 0 1 1 0 00-2 0M8.5 8.5a5 5 0 000 7M15.5 8.5a5 5 0 010 7M5.6 5.6a9 9 0 000 12.8M18.4 5.6a9 9 0 010 12.8"},
```

mintsoft:
> For LLDP a radio shape reads as "announcing to the neighbours", and this one renders cleanly at 16 px with the same stroke settings as the others:

Yep that's perfect, thank you!

PR #430, opened 2026-09-10
stat: do not fetch the port status twice when opening the page

DrDoof:
Opening the statistics page asks for /status.json twice: `sectionInits.stat`
calls `update()`, and its callback calls `update()` again. The second reply is
not used.

Dropping the inner call is enough, since the callback already runs once the
first reply has been parsed.

### DIFF
--- a/html/main.js
+++ b/html/main.js
@@ -1746,7 +1746,6 @@ document.addEventListener('DOMContentLoaded', function() {
 
 sectionInits.stat = function() {
   update( () => {
-    update();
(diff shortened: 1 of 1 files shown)


feelfree69:
I guess #430, #431, #432, #433 and #434 can all be closed, as they all have been resolved with #429, right?

DrDoof:
Closing: #429 replaced main.js, and the new page polls /status.json from a single Poller, so the double fetch is gone.

PR #431, opened 2026-09-10
l2: stop the FDB walk when its section is left

DrDoof:
`walkL2()` reads the forwarding database a page at a time, asking for the next
page from `setTimeout(page, 1000)`. If the user moves to another section while
a walk is in progress, the walk carries on. `devWalk()` does check that the
ports section is still visible, but that check sits on the 15 s restart of the
whole walk rather than on the paging inside one.

This is an easy one to miss: `clearSectionIntervals()` clears
`sectionIntervals`, `l2Timer` and `devTimer`, and reads as though it covers
everything. The paging timer simply is not registered with it.

The patch gives `walkL2` a generation number that `clearSectionIntervals()`
bumps, and both timers check it before asking for the next page.

It matters more on a larger table than on mine: the paging accepts up to 4096
entries, so a walk can run to about 146 requests at one per second on a page
that has no use for them.

### DIFF
--- a/html/main.js
+++ b/html/main.js
@@ -932,6 +932,7 @@ document.addEventListener('DOMContentLoaded', function() {
  * shown.  Each section initialises on first display, and its polling
  * intervals run only while the section is visible. */
 var sectionIntervals = [];
+var l2WalkGen = 0;
 var sectionInits = {};
 
 function setSectionInterval(fn, ms) {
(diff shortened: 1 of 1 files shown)


DrDoof:
Closing: #429 replaced main.js, and the MAC table walk in app.js already stops when the section is left (l2Gen).

PR #432, opened 2026-09-10
vlan: keep every VLAN in the table and fetch the list once

DrDoof:
A few things I ran into on the VLAN page. They all sit in the same three
functions, so they come together here rather than as separate changes that
would conflict.

1. `loadVlanTable()` builds the table from one /vlanlist plus one /vlan.json
   per VLAN, and uses `continue` when one of those does not come back. The row
   is then simply absent, which is hard to tell apart from a VLAN that was
   never configured, and since the per-VLAN requests go in order it tends to be
   the end of the list that is missing. A VLAN that cannot be read now keeps
   its row, using the id and name /vlanlist already returned, with `?` in the
   port columns.

2. A 401 took that same path. The rest of the file sends the browser to
   login.html when it sees one, so these two places now do the same.

3. `refreshVlanViews()` calls `loadVlanList()` and `loadVlanTable()`, and each
   fetches /vlanlist for itself. It is fetched once now and passed to both.

One more in `loadVlanList()`: the selector is hidden on a non-200 reply and
there is no path that brings it back, so a single failed reply leaves it hidden
for the rest of the page's life. It is shown again on success.

The table is also assembled in a document fragment and swapped in at the end.
Clearing tbody first and appending row by row meant that leaving the page and
coming back during a load could let the second pass clear the table under the
first.

### DIFF
--- a/html/main.js
+++ b/html/main.js
@@ -2013,25 +2013,45 @@ function portsToRange(mask, nPorts) {
   return parts.length ? parts.join(',') : '-';
 }
 
-async function loadVlanTable() {
+var vlanTableGen = 0;
+
+async function loadVlanTable(vlans) {
   var tbody = document.getElementById('vlanTableBody');
   if (!tbody) return;
-  tbody.innerHTML = '';
-  var resp;
-  try { resp = await fetch('/vlanlist'); } catch(e) { return; }
-  if (!resp.ok) return;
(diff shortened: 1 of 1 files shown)


DrDoof:
Closing: #429 replaced main.js, and app.js keeps a VLAN's row when its details cannot be read.

PR #433, opened 2026-09-10
web: send one request at a time

DrDoof:
The switch has a single TCP connection slot, so a second connection opened
while the first is still up gets no SYN-ACK and the browser waits out a
retransmission before retrying: about a second the first time, two the next.

`sendXHTTP()` is aiming at exactly this, and the shape of it is right. The
subtlety is that `onreadystatechange` fires for readyState 2 and 3 as well as
4, so `currentRequests` empties as soon as the first header of a reply arrives
and the next request goes out while the previous one is still on the wire. It
is the sort of detail that only shows itself when a timer happens to fire two
requests at once.

This replaces the counter with a promise chain, and routes `fetch()` through
the same chain via `qfetch()` so that XHR and fetch queue against each other
rather than only among themselves. The chain is released after 15 s if a
request never settles, so one request that hangs cannot hold up the rest.

Measured in a browser on the spanning tree page, which polls two endpoints on
the same 2 s timer. Three requests issued together:

```
before   42 / 1100 / 2130 ms
after    55 /  110 /  173 ms
```

Sending the same three outside the chain on the same build still gives
59 / 305 / 1114 ms, with `tcp: found no unused connections` on the console.

Two requests that called `send()` directly now go through the queue as well.

### DIFF
--- a/html/main.js
+++ b/html/main.js
@@ -627,8 +627,6 @@ function linkText(idx) { var v = linkS[idx]; return typeof v === 'function' ? v(
 var logToPhysPort = new Int8Array(10);
 var physToLogPort = new Int8Array(10);
 var portNames = new Array(10);
-var currentRequests = [];
-var currentCallback;
 function drawPorts() {
   var f = document.getElementById('ports');
   console.log("DRAWING PORTS: ", numPorts);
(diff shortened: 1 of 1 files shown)


DrDoof:
Closing: #429 replaced main.js, and every request in app.js goes through the api() queue, which waits for the previous body.

PR #434, opened 2026-09-10
web: fetch the port icons once and reuse them

DrDoof:
The VLAN, link aggregation and mirroring pages draw an icon per port in each
selector grid, up to 36 `` elements pointing at port.svg and sfp.svg.
Static files go out with `max-age=60` and no validator, so a minute after the
page has loaded those are all fetched again.

This fetches the two files once and keeps them as `data:` URLs for the ``
elements to use.

Worth flagging for anyone touching this later: the elements need to stay
``. The checkbox in each grid is invisible, and style.css hangs the
pointer, the checked state and the spacing off `.psel + img`, so putting an
inline `<svg>` there quietly takes all three away. That is how I found out.

### DIFF
--- a/html/main.js
+++ b/html/main.js
@@ -966,8 +966,34 @@ window.addEventListener('hashchange', function() {
   showSection((location.hash || '#/overview').replace(/^#\//, ''));
 });
 
+var svgWarm = null;
+var svgData = {};
+
+function warmPortSvgs() {
+  if (svgWarm) return svgWarm;
+  svgWarm = Promise.all(['port.svg', 'sfp.svg'].map(function(u) {
+    return fetch(u)
+      .then(function(r) { return r.ok ? r.text() : null; })
+      .then(function(txt) {
+        if (txt)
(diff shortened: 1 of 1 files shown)


tofurky:
Could we just change Cache-Control for `mime_SVG` (and others, as applicable e.g. `mime_PNG`) and instead rely on the browser to handle caching? e.g. `Cache-Control: max-age=2592000, immutable` though it'd require splitting this up into chunks with conditions in between:

```
slen += strtox(outbuf + slen, "Cache-Control: max-age=60, must-revalidate\r\nConnection: close\r\nAccess-Control-Allow-Origin: *\r\nContent-Security-Policy: default-src 'self'; script-src 'self' 'unsafe-inline'; style-src 'self' 'unsafe-inline'; img-src 'self' data:; connect-src 'self'; form-action 'self'\r\n\r\n");
```

This also seems to be missing the ports at the top of the page and the MTU section.

An aside, dev tools network tab can be misleading regarding caching.

DrDoof:
Closing: #429 replaced main.js, and the new page does not load port.svg/sfp.svg images any more.

PR #435, opened 2026-09-11
Support for Lianguo HYWS-SGT0108S 8 + 1 unmanaged

bennydiamond:
https://github.com/logicog/RTLPlayground/pull/427 needs to be merged first.

This unit is an oldie I had laying around. I figured I might as well tinker with it. Form factor is interesting although the port representation in WebUI is not suited, yet.

I will try to come up with configurable UI layout for the ports on main page in a future PR.


vDorst:
Please edit the pictures so that all the non-device is reduced to minimum.
PCB-bot.jpg rotate 180 deg.
Remove the flash changes which is a already separate PR.

bennydiamond:
> Please edit the pictures so that all the non-device is reduced to minimum. PCB-bot.jpg rotate 180 deg. Remove the flash changes which is a already separate PR.

Done.

Is there a dedicated discussion space for dev talk? Obviously, the current UI layout for ports putting all network ports in a single row isn't ideal for this particular unit.

I'm trying to come up with a way to configure ports layout in UI without resorting to increasing the size of `struct machine` _too much_ with extra coord arrays and such that won't benefit any other device in here an simply contribute to code bloat. So far a pointer to custom struct containing sets of coordinates seems the right idea. Minimal bloat.

I know there is supposedly a refactor of the Web UI coming at some point. Is this a point already addressed? 
Is it even worth it for me to start working on it at this point?
Should I just work on what I think is an acceptable solution and blindly submit a PR?

machine.c
```diff
@@ -1145,6 +1144,45 @@ __code const struct machine machine = {
...
 
+#elif defined MACHINE_LIANGUO_HYWS_SGT0108S
+__code const struct machine machine = {
+    .machine_name = "Lianguo HYWS-SGT0108S",
+    .isRTL8373 = 1,
+    .min_port = 0,
+    .max_port = 8,
+    .n_sfp = 1,
+    .log_to_phys_port = {2, 1, 4, 3, 5, 6, 7, 8, 9},
+    .phys_to_log_port = {0, 1, 2, 3, 4, 5, 6, 7, 8},
```
vDorst: phys_to_log_port should have the value that is equal idx into log_to_phys_port.

So
```
.log_to_phys_port = {2, 1, 4, 3, 5, 6, 7, 8, 9},
// Idx               0, 1, 2, 3, 4, 5, 6, 7, 8

.phys_to_log_port = {1, 0, 3, 2, 4, 5, 6, 7, 8}
```

bennydiamond: Done

vDorst:
Thanks

PR #439, opened 2026-09-12
I2CBUS: Calculate I2C-settings as compiletime instead at runtime.

vDorst:
We can reduce the i2c-settings to just one byte, and this new const value can directly be used by the I2C-peripheral without additional computation. This saves 1-xdata load and a lot of bit-shifting.

I see a codesize savings of +200 bytes.


DrDoof:
Nice one, thanks! Moving this to compile time makes the machine definitions much easier to read, and I like that the bus numbers can now be seen right where the pins are listed.

I wanted to be sure the new byte really matches what the old code worked out at runtime, so I put the old `i2c_bus_from_sda_pin()` and `i2c_bus_from_scl_pin()` next to your `I2CBUS()` in a small host program and ran every pin pair that machine.c uses through both. They agree on all of them, and I could not find anything still reaching for the old `.sda` / `.scl` fields. All machines also build fine with it.

There is one small thing I noticed along the way. No board hits it today, so please treat it as a suggestion rather than something that has to hold up the merge.

In `setup_i2c()` the SCL bus is read back as `(i2c >> 5) & 0x03`, but SCL takes up three bits of that byte. If someone ever puts a pin in there that is not an SCL pin, the macro gives -1, that ends up as `111`, and the 0x03 mask turns it into bus 3. The switch would then happily set up the pin mux for bus 3, while the old code landed in `default` and printed "Invalid SCL bus number", which is a much friendlier way to find out about a typo in a new machine definition. SDA still catches it, since that side is masked with 0x07.

While looking at that I also saw that shifting -1 is undefined behaviour in C (clang points it out with -Wshift-negative-value), and that the SCL macro does not have the `(uint8_t)` casts the SDA one has. Maybe something like this would take care of both:

```c
#define I2C_BUS_FROM_SCL_PIN(scl_pin) \
	((scl_pin) == GPIO46_I2C_SCL0 ? (uint8_t)0 : \
	 (scl_pin) == GPIO48_I2C_SCL1 ? (uint8_t)1 : \
	 (scl_pin) == GPIO50_I2C_SCL2_UART1_TX ? (uint8_t)2 : \
	 (scl_pin) == GPIO40_I2C_SCL3_MDC1 ? (uint8_t)3 : (uint8_t)7)
```

together with `(uint8_t)7` as the fallback in `I2C_BUS_FROM_SDA_PIN` and `& 0x07` for `scl_bus` in `setup_i2c()`. Since 7 is not a valid bus on either side, both default branches would speak up again. As a small bonus, an unknown SDA pin would then stay inside its own three bits. Right now -1 shifted by 2 also sets bits 7:5 and overwrites the SCL selection in the control register, and the old 0xFF return did the same, so that part would even be an improvement over before.

Apart from that, this looks good to me.

vDorst:
> In setup_i2c() the SCL bus is read back as (i2c >> 5) & 0x03, but SCL takes up three bits of that byte. If someone ever puts a pin in there that is not an SCL pin, the macro gives -1, that ends up as 111, and the 0x03 mask turns it into bus 3. The switch would then happily set up the pin mux for bus 3, while the old code landed in default and printed "Invalid SCL bus number", which is a much friendlier way to find out about a typo in a new machine definition. SDA still catches it, since that side is masked with 0x07.

What I was hoping that I can just put a `#error()` in the macro.
So the compiler just stops at a invalid value.
But that seems not possible according AI.

I hope someone has a way to make it happen, because that seems the only proper way to solve it.

I am using you suggestion to return 7 as invalid value.

Rebased the code and de-magic some values.

DrDoof:
On the compile-time check: maybe give this a try, it does work with SDCC. #error can't see the value, and _Static_assert is only a warning here, but a zero-length array is a hard error and it is fine inside the const initializer:

```c
#define I2CBUS_VALID(bus) (0 * sizeof(char[(bus) != I2CBUS_INVALID ? 1 : 0]))
```

OR it into I2CBUS() for both the SDA and the SCL half. With good pins the output is byte for byte the same; with a wrong pin the build stops with "error 176" on the line of the machine definition. Tried it on your branch and it behaves.

vDorst:
@DrDoof Do you have a full code snippet how to implement this?
I am failing to make it work.

DrDoof:
Of course, and sorry that my first note was a bit too short to be useful on its own! Here is the whole thing. It only replaces the `I2CBUS()` definition in machine.c, so `I2CBUS_INVALID` and both `I2C_BUS_FROM_*_PIN` macros can stay exactly as you have them now:

```c
#define I2CBUS_VALID(bus) (0 * sizeof(char[(bus) != I2CBUS_INVALID ? 1 : 0]))

#define I2CBUS(sda, scl) \
	((uint8_t)((I2C_BUS_FROM_SCL_PIN(scl) << RTL837X_REG_I2C_SCL_SHIFT) | \
		   (I2C_BUS_FROM_SDA_PIN(sda) << RTL837X_REG_I2C_SDA_SHIFT) | \
		   I2CBUS_VALID(I2C_BUS_FROM_SCL_PIN(scl)) | \
		   I2CBUS_VALID(I2C_BUS_FROM_SDA_PIN(sda))))
```

You can also forget about the `(uint8_t)` casts on the SCL branches that I suggested earlier. They were only there because of the -1 being shifted, and now that you return 7 nothing negative gets shifted anymore, so your version is fine as it is.

There is one little trap, and I would not be surprised if that is where you got stuck: the size for the bad case really has to be `0`. The well known trick with `-1` simply does nothing in SDCC, it accepts a negative array size without even a warning. `_Static_assert` is not much help either, since it only warns and the build happily carries on.

I checked it again on your branch (2c68b9f) with SDCC 4.5, just to be sure. With the pins as they are, machine.c comes out identical for every machine. And when I put a wrong pin into a definition on purpose, for example `I2CBUS( GPIO37, GPIO40_I2C_SCL3_MDC1 )`, the build stops with

```
machine.c:<line>: error 176: sizeof applied to an incomplete type
```

A wrong SCL pin gets caught the same way. The message itself is a bit cryptic, but the line number points straight at the machine definition with the typo, which is exactly what you were hoping for with the `#error` idea.

Fingers crossed it works for you now, and thanks for sticking with this!

vDorst:
@DrDoof Thanks for helping me.
I made it work and tested some cases, which errors out as intended.

vDorst:
rebased it so the two new devices are also up-to-date.

vDorst:
@DrDoof Can you review this again?
Rebased, and refactor the newly added devices.

DrDoof (approved):
Had another look, and the compile-time check works nicely now. Thanks for
sticking with it through all the back and forth.

I went over the current state rather than the parts we already settled:

- Every pin pair that machine.c uses (four distinct ones) still comes out byte
  for byte the same as the old runtime code worked out: 0x70, 0x6c, 0x00 and
  0x24. No differences.
- The zero-length-array check really does bite. I put a non-I2C pin into the
  definitions on purpose and the build stops with
  "machine.c:60: error 176: sizeof applied to an incomplete type", pointing
  straight at the line with the bad pin.
- All 38 machines build, each one producing an image. Worth saying because CI
  only builds one of them, and this branch was rebased onto the devices that
  were added recently.
- On SWTGW218AS the saving is a little better than you wrote: measured against
  the branch point, BANK2 goes 43323 -> 43150 and BANK3 11166 -> 11031, so 308
  bytes rather than 200, with CSEG and BANK1 unchanged.

One thing I would still change, and it is a small one.

In setup_i2c() the SCL bus is masked in the switch:

```c
	uint8_t scl_bus = (i2c >> RTL837X_REG_I2C_SCL_SHIFT);
	...
	switch (scl_bus & 0x03) {
```

With valid pins scl_bus is 0..3, so the mask never changes anything and is not
needed. With an invalid pin the byte is 0xe0, scl_bus reads 7, and `& 0x03`
turns that into 3, so the code would quietly set up the pin mux for bus 3.
Now that the default branches are gone, nothing would say a word about it.

The compile-time check does catch this today, and I confirmed that it does, so
this is not a bug you can reach from any machine definition. It is more that
the mask silently converts a wrong value into a plausible one, which is the
opposite of what the rest of the change is doing. Dropping `& 0x03` costs
nothing and keeps a bad value looking bad.

Same place, a detail: the print above the switch shows the unmasked scl_bus,
so if this ever did happen the log would say 7 while the switch acted on 3.

Apart from that this looks good to me.

vDorst:
I used the `& 0x03` in the switch statement to fix a compiler warning `warning 110: conditional flow changed by optimizer: so said EVELYN the modified DOG`.

This line still needs `0x03` mask and a define to de-magic it.
```c
uint8_t scl_bus = (i2c >> RTL837X_REG_I2C_SCL_SHIFT) & 0x03;
```

The warning can be fixed in at least two ways.
1. At `& 0x03` in the switch statement.
2. Replace the last case with a default.

What to you think is better?

DrDoof:
Good question, and I was curious enough that I went and built the series commit
by commit to find out where that warning actually lives.

You were right that it was there. It turns up at f65d14f, and the reason is
rather neat once you see it: with the read masked & 0x03 the value can only be
0..3, the four cases cover exactly that, so the default: branch can never be
reached. The optimizer throws it away and tells us so.

What surprised me is that it had already gone by 19ccc40, as soon as the read
mask became & 0x07 - the default was reachable again and the warning went away
by itself. So by the time the mask moved into the switch there was nothing left
for it to silence. Easy to miss, since both changes live in the same stretch of
the series.

I tried the three shapes on the code as it stands, on SWTGW218AS and on the
machine CI builds, SDCC 4.6.0, same answer both times:

  switch (scl_bus)                        no warning
  scl_bus = (...) & 0x03; switch (...)    warning 110 at boot.c:404
  default: instead of the last case       no warning

They also come out the same size, CSEG 8612 and BANK2 43148, so nothing is
really at stake here beyond how it reads.

So if I may offer a third option: drop the & 0x03 from the switch and leave the
read as it is. No warning, nothing gained or lost in size, and a value that
should never occur stays visibly wrong instead of quietly turning into bus 3,
which is all I was after in the first place. It keeps the print above the
switch honest too, since it would then show the same number the switch acts on.

The one I would steer away from is putting the mask back on the read line. Of
the three, that is the only one that brings the warning back.

Either way, thanks for chasing this down properly instead of just papering over
it.

vDorst:
I keep the `& 0x03` on both places and added a comment why it is there.
That feels the right way.

PR #443, opened 2026-09-14
Makefile: let the linker check external RAM against 0xBFFF

DrDoof:
This is a small one, but I think it is worth having. Right now the linker does not know how much external RAM the 8051 actually has, so it quietly assumes 16 MB. If XDATA ever grew past the end of the real memory, the build would still come out green and we would only find out once the switch started behaving strangely. Adding `--xram-size` to the link step lets the linker tell us instead.

I wanted the number to come from the hardware rather than from a datasheet guess, so I tried it on my SWTGW218AS. I wrote a byte to a high XDATA address and read it back from the same address, and also looked at 0x0000 to make sure nothing was wrapping around. Everything up to and including 0xBFFF kept its value, and a byte written to 0xC000 came back as 0x00. So we have 48 kB, from 0x0000 to 0xBFFF. That holds with `SFR_97` at 0, which is what `main()` sets; as ranma found in #51, other values map a flash page in at 0x4000.

You might wonder why the value is 49151 and not the rounder 49152. I was curious how the linker counts, so I relinked with limits right around what we use today. XSEG is 14404 bytes starting at 0x0001 and XISEG adds another 20, which puts the last byte at 0x3858. A limit of 14424 still links, while 14423 stops with `?ASlink-Error-Insufficient EXTERNAL RAM memory.` It turns out the linker just adds up the sizes and does not count the fact that allocation starts at 0x0001. With 49152 the very last byte could land on 0xC000, so 49151 is the value that really stops at 0xBFFF.

To be sure the flag is not just decoration, I also tried a limit that is far too small, 4096, and the link fails as it should. With the flag in place all 34 machines in machine.h build fine, and every map file is identical to the one without it, so the images do not change at all. I did that run with 49152, but at today's usage both values behave exactly the same. Usage is the same on every board too, 14424 bytes, so there is still lots of room.

The one thing I could not try is an RTL8372, since I only have an RTL8373 here. If someone has an RTL8372 at hand and it turns out to have a different amount of RAM, I would be happy to turn this into a per chip value instead.


logicog:
In PR #207 there is an absolute allocation of an xram structure, because the assembly requires 256 byte alignment:
```
__xdata struct chacha20_t __at(0x7000) chacha20;
``` 
I assume this would get caught if the 0x7000 were outside of the ram size, but it would not prevent any other xram user to overwrite the absolutely allocated structure, or?

vDorst:
Do you have code to test? I do have RTL8372 (SWGT024_V2_0_MANAGED) and RTL8383N (KP_9000_9XHML_X_V3_1).

With `SFR_97` toggle we can make serving the website faster by just mapping the html-pages and use memcpy the fetch the data.
The CPU should support Double data-pointer which makes copying data even more faster with a assembly written memcpy that makes use of both data pointers.

DrDoof:
@logicog I went and tried it, and the absolute block turns out not to be checked at all. With sdcc 4.6.0, a variable placed with `__at` lands in the absolute area, while the linker only adds up XSEG, which starts at 0x0001 either way. Give XSEG 40000 bytes and the `__at(0x7000)` block sits right in the middle of it, with no complaint. `__at(0xd000)` with `--xram-size 49151` links too, so even an address past the end of the RAM goes through. The flag isn't asleep, though: `--xram-size 8` fails the way it should.

So it doesn't make #207 safe. If the 256 byte alignment is what forces the absolute address, the structure could sit in ordinary xdata and be aligned in the assembly instead, or the top of the RAM could be kept for it and the linker told a smaller size.

@vDorst Sure, here you go. Paste this in front of another `} else if` branch in cmd_parser.c, so the next branch closes it, and run `xprobe`. It prints the address, what comes back after writing 0xa5, and a 1 if a low address moved with it, so aliasing shows up too. It puts the old byte back and stays above 0x8000, clear of the flash page that `SFR_97` maps at 0x4000.

```c
} else if (cmd_compare(0, "xprobe")) {
	static __xdata uint8_t * __xdata p;
	static __xdata uint8_t * __xdata q;
	static __xdata uint16_t a;
	static __xdata uint8_t old;
	static __xdata uint8_t low;
	for (a = 0x8000; a; a += 0x800) {
		p = (__xdata uint8_t *)a;
		q = (__xdata uint8_t *)(a & 0x3fff);
		old = *p;
		low = *q;
		*p = 0xa5;
		print_short(a);
		write_char(' ');
		print_byte(*p);
		write_char(' ');
		print_byte(*q == low ? 0 : 1);
		write_char('\n');
		*p = old;
	}
```

On my RTL8373 everything up to 0xBFFF keeps the value and 0xC000 and up read back as 00. If the RTL8372 or the RTL8383N say something else, I'll happily make this a per chip value instead of one number for all boards.

The SFR_97 idea sounds lovely, and the single TCP slot would certainly notice. Glad to help test it once this is out of the way.

vDorst:
`RTL8372 (SWGT024_V2_0_MANAGED)`

```
> xprobe 
0x8000 a5 00
0x8800 a5 00
0x9000 a5 00
0x9800 a5 00
0xa000 a5 00
0xa800 a5 00
0xb000 a5 00
0xb800 a5 00
0xc000 c0 00
0xc800 c8 00
0xd000 d0 00
0xd800 d8 00
0xe000 e0 00
0xe800 e8 00
0xf000 f0 00
0xf800 f8 00
```

`RTL8383N (KP_9000_9XHML_X_V3_1)`
```
> xprobe
0x8000 a5 00
0x8800 a5 00
0x9000 a5 00
0x9800 a5 00
0xa000 a5 00
0xa800 a5 00
0xb000 a5 00
0xb800 a5 00
0xc000 c0 00
0xc800 c8 00
0xd000 d0 00
0xd800 d8 00
0xe000 e0 00
0xe800 e8 00
0xf000 f0 00
0xf800 f8 00
```

Both have the same result.

DrDoof:
Thanks for running it on both. Same picture as here: the writes stick up to 0xB800, nothing above 0xC000 holds, and no aliasing, so the 48 kB isn't specific to the RTL8373. That answers the open question in the description, the flag can stay one value for all boards.

One small difference: above the RAM your chips read back the high byte of the address, mine gives the same byte masked to six bits, so 0xC000 comes out as 00. Nothing attached either way.

Happy to look at the SFR_97 mapping once this one is out of the way.

DrDoof:
@logicog A small follow-up, since I went through the linker and compiler sources after my last comment and ran a few more links to see why it slips through.

It turns out to be simpler than I made it sound. A variable with `__at` and no initializer takes up no space at all as far as the linker is concerned, SDCC just emits `_chacha20 = 0x7000` as a symbol, so there is nothing for the check to count. The check itself adds up the sizes of all the xdata areas, XISEG as well and not only XSEG as I wrote, and compares the total with `--xram-size`, but it never looks at addresses. When I gave a test variable at 0xF000 an initializer, its 16 bytes did get counted, and it still linked fine even though 0xF000 is well past the end of the RAM. This part of the code is the same in 4.5 and 4.6.

For #207 the top of the RAM would work well: `__at(0xBF00)` together with `--xram-size 48895`. Everything relocatable starts at 0x0001, so the linker then keeps it below 0xBF00. I tried the boundary, 48895 bytes link and end at 0xBEFF, and one byte more fails. The assembly takes the page only from `mov dptr, #_chacha20` and sets DPL to offsets up to 127, and the structure is 134 bytes, so everything stays inside that page and the assembly should not need any changes. I am happy to test it when #207 gets there.

bloqaudio (approved):
Checked the three things this rests on, on current main with SDCC 4.5.0. All hold.

**The flag is inert on the output.** This needs care, because a plain rebuild-and-compare does not show it. The generated `html_data` table is not reproducible byte for byte between runs on the same host, so two builds of the same tree drift by a few bytes in `BANK1` and `BANK2` regardless of the flag, and I first saw a six byte difference that had nothing to do with your change. Linking the same object files twice, once with `--xram-size 49151` and once without, gives byte-identical `.ihx` files. That is the comparison that answers it, and it confirms what you claim.

**It fires.** With today's usage, `--xram-size 4096` stops the link with `?ASlink-Error-Insufficient EXTERNAL RAM memory.`

**The counting behaviour, and why 49151.** Confirmed exactly, though the absolute numbers have moved since you wrote this. On current main `XSEG` is 14552 bytes at `0x0001` and `XISEG` adds 20, so the last byte sits at `0x38EC` and the total is 14572. A limit of 14572 links and 14571 does not, so the linker is adding up segment sizes and taking no account of allocation starting at `0x0001`, exactly as you found with 14424 and 14423. That is what makes 49152 wrong in principle: it would let the last byte land on `0xC000`. 49151 is the value that really stops at `0xBFFF`. Both link today, since usage is nowhere near either, but the reasoning for preferring the odd-looking number is sound and worth keeping in the commit message.

Usage is identical on the two machines I built, `PCB_SWTG018AS_V2_1_0` and `SWTGW218AS`, at 14552 plus 20, which matches your observation that it does not vary by board.

**What I did not verify.** I did not independently measure the RAM size. That would need a debug build to write and read back high XDATA, and it would not have answered your open question anyway: this board is an RTL8373, the same part you have, so a second measurement on the same silicon adds little. I do not have an RTL8372 either, so the question you raised at the end stays open.

One observation on the risk of getting the number wrong, since that is the only thing this flag can break. Too low and a legitimate build is refused. Too high and it simply fails to catch an overflow, which is where the tree already is today with the linker assuming 16 MB. At 14572 bytes of usage against a 49151 limit there is about 34 kB of headroom, so even if some part turned out to have more RAM than this, nothing would be refused for a long time. The downside is bounded and the upside is a build that fails loudly instead of a switch that behaves strangely.

Approving.

feelfree69:
Thanks here as well!

bloqaudio:
A correction to my review above. I said the `html_data` generation drifts by a few bytes between builds on the same host. That was wrong. The six bytes were the `-dirty` suffix on the version string: adding the flag to the `Makefile` dirtied the tree. Reproduced on `9499256`: an uncommitted `Makefile` edit turns `VERSION_SW` from `v0.1.0-9499256` into `v0.1.0-9499256-dirty`, which moves `l_BANK1` and `l_BANK2` by exactly 6 each. Two clean builds of main on one host are byte-identical. The relink comparison and the conclusion that the flag does not change the output both still hold.

PR #445, opened 2026-09-14
sflow: send port counter samples to a collector

DrDoof:
This adds an sFlow version 5 agent that sends counter samples, so the traffic and error counters of every port can be graphed by any sFlow collector. Each datagram carries one port's generic and Ethernet interface counters, read from the MIB counters of the ASIC, and the ports take turns within the interval. Flow samples aren't part of it.

It's set up with "sflow collector <ip> [port]", "sflow interval <seconds>" and "sflow on|off", or on a new sFlow page in the web interface. The agent sits in BANK3 and its command line in BANK2.

I checked the datagrams with a host build of sflow.c against a decoder, and on a SWTGW218AS with a collector: unicast, broadcast and multicast bursts show up on the right ports and in the right direction. The error, discard and collision counters stayed at zero there, so for those the mapping only follows the counter table in the web interface.


sflow.c
```diff
@@ -0,0 +1,197 @@
...
+
+
+static void sflow_sample(void) __reentrant
+{
+	uint8_t speed, up;
+
+	reg_read_m(RTL837X_REG_LINKS_STS);
+	up = ((sfr_data[1] | ((uint16_t)sfr_data[2] << 8)) >> sflow_port) & 1;
+	reg_read_m(sflow_port >= 8 ? RTL837X_REG_LINKS_89 : RTL837X_REG_LINKS);
+	speed = sfr_data[3 - ((sflow_port & 7) >> 1)];
```
vDorst: Just use `reg_read()` and use the `SFR_DATA_X` directly.

sflow.c
```diff
@@ -0,0 +1,197 @@
...
+
+void sflow_init(void) __banked
+{
+	sflow_state.enabled = 0;
+	sflow_state.collector[0] = sflow_state.collector[1] = 0;
+	sflow_state.collector[2] = sflow_state.collector[3] = 0;
+	sflow_state.port = SFLOW_PORT_DEFAULT;
+	sflow_state.seq = 0;
+	sflow_state.conn = 0;
+	sflow_interval(SFLOW_INTERVAL_DEFAULT);
```
vDorst: Maybe cheaper to just use memset for the hold struct.

sflow.c
```diff
@@ -0,0 +1,197 @@
...
+	speed = up ? (speed & 7) << 1 : 6;
+
+	sflow_len = 0;
+	sflow_put32(5);
+	sflow_put32(1);
+	memcpy(SFLOW_P + sflow_len, uip_hostaddr, 4);
+	sflow_len += 4;
+	sflow_put32(0);
+	sflow_put32(++sflow_state.seq);
+	sflow_put32(ticks * (1000 / SYS_TICK_HZ));
```
vDorst, line 115: This doesn't seems y2038 ready.

sflow.c
```diff
@@ -0,0 +1,197 @@
...
+	reg_read_m(sflow_port >= 8 ? RTL837X_REG_LINKS_89 : RTL837X_REG_LINKS);
+	speed = sfr_data[3 - ((sflow_port & 7) >> 1)];
+	speed = (sflow_port & 1) ? speed >> 4 : speed & 0xf;
+	speed = up ? (speed & 7) << 1 : 6;
+
+	sflow_len = 0;
+	sflow_put32(5);
+	sflow_put32(1);
+	memcpy(SFLOW_P + sflow_len, uip_hostaddr, 4);
+	sflow_len += 4;
```
vDorst: I think it is way cheaper to make SFLOW_P a pointer and drop sflow_len.
Length can be recovered by `len = pointer - SFLOW_P`.
And use `*ptr++ = value` were you can.
Every function that modifies the pointer just return length appended.
```c
ptr += some_func(ptr, value);
```

Also `sflow_put32(<constvalue>)` can be replaced with memcpyc?
Passing around a 32-bit values isn't cheap.
Or make it a separate function that just takes a uint8_t for the small values.
Like `sflow_put8(ptr, uint8_t)`.

sflow.c
```diff
@@ -0,0 +1,197 @@
...
+	}
+}
+
+
+static void sflow_sample(void) __reentrant
+{
+	uint8_t speed, up;
+
+	reg_read_m(RTL837X_REG_LINKS_STS);
+	up = ((sfr_data[1] | ((uint16_t)sfr_data[2] << 8)) >> sflow_port) & 1;
```
vDorst: Just use `reg_read()` and use the `SFR_DATA_X` directly.

sflow_cli.c
```diff
@@ -0,0 +1,68 @@
...
+		return;
+	}
+	if (cmd_compare(1, "collector") && cmd_words_len >= 3 && parse_ip(cmd_words_b[2])) {
+		if (cmd_words_len >= 4 && (!atoi_short(cmd_words_b[3]) || !atoi_results_short))
+			goto err;
+		sflow_stop();
+		sflow_state.collector[0] = ip[0];
+		sflow_state.collector[1] = ip[1];
+		sflow_state.collector[2] = ip[2];
+		sflow_state.collector[3] = ip[3];
```
vDorst: memcpy

sflow_cli.c
```diff
@@ -0,0 +1,68 @@
...
+#pragma constseg BANK2
+
+extern __xdata uint8_t cmd_words_len;
+extern __xdata uint8_t cmd_words_b[15];
+extern __xdata uint8_t ip[4];
+extern __xdata uint16_t atoi_results_short;
+uint8_t cmd_compare(uint8_t start, __code uint8_t * cmd);
+uint8_t atoi_short(uint8_t idx);
+uint8_t parse_ip(uint8_t idx);
+
```
vDorst: Are these prototype needed?

sflow.h
```diff
@@ -0,0 +1,31 @@
...
+
+#include <stdint.h>
+
+#define SFLOW_PORT_DEFAULT	6343
+#define SFLOW_INTERVAL_DEFAULT	20
+
+struct uip_udp_conn;
+
+struct sflow_state {
+	uint8_t enabled;
```
vDorst: make it a bool if it had only state 0 or 1.

vDorst (commented):
First quick review.

DrDoof:
Thanks, all in. Measured on SWTGW218AS: memset -26 B, pointer + put8 -35 B, SFR read -5 B, memcpy and prototypes -13 B of bank 2. The prototypes stay because main has no header for cmd_compare and friends; they live in cmd_parser.h now.

I compared old and new sflow.c on the host against the same fake registers: datagrams identical. Worth doing, my first SFR read had the two link-state bytes swapped.

Uptime is milliseconds since boot per sFlow v5, wrapping at 49.7 days like sysUpTime, so 2038 does not apply.

PR #446, opened 2026-09-14
dns: resolve host names

DrDoof:
This adds a small resolver for IPv4 address records, mostly so that the NTP client in the next PR can take a host name. "dns server <ip> [<ip>]" sets up to two servers, without them the one from DHCP is used, and "dns lookup <name>" resolves a name on the console. The servers can also be set on the System tab, below the gateway.

Queries go to the servers in turn every two seconds, six tries in total, and compressed names and a CNAME ahead of the address are handled. I tested the parser on the host against crafted answers (A, CNAME then A, wrong id, NXDOMAIN, AAAA only), and on a SWTGW218AS it resolves pool.ntp.org.


dns.c
```diff
@@ -0,0 +1,226 @@
...
+	if (dns_used(dhcp_state.dns))
+		dns_cand[k++] = dhcp_state.dns;
+	if (!k || dns_state.tries >= DNS_TRIES) {
+		dns_state.status = DNS_FAILED;
+		if (dns_state.verbose)
+			print_string(k ? "DNS: no answer\n" : "DNS: no server\n");
+		return;
+	}
+	k = dns_state.tries % k;
+	uip_ipaddr(dns_ip, dns_cand[k][0], dns_cand[k][1], dns_cand[k][2], dns_cand[k][3]);
```
vDorst:
```patch
diff --git a/dns.c b/dns.c
index 2dc7e57..babeb3a 100644
--- a/dns.c
+++ b/dns.c
@@ -76,7 +76,7 @@ static void dns_next(void) __reentrant
                return;
        }
        k = dns_state.tries % k;
-       uip_ipaddr(dns_ip, dns_cand[k][0], dns_cand[k][1], dns_cand[k][2], dns_cand[k][3]);
+       memcpy(dns_ip, &dns_cand[k], 4);
        dns_state.conn = uip_udp_new(&dns_ip, HTONS(DNS_PORT));
        if (!dns_state.conn) {
                dns_state.status = DNS_FAILED;
```
This changes saves +140 bytes. So it seems that the `uip_ipaddr` is very inefficient to set IP.

dns.c
```diff
@@ -0,0 +1,226 @@
...
+	}
+	dns_state.tries = DNS_TRIES;
+	dns_next();
+}
+
+
+void dns_init(void) __banked
+{
+	memset(dns_state.server, 0, sizeof(dns_state.server));
+	dns_state.name[0] = 0;
```
vDorst:
```c
dns_state.name[0] = NUL;
```

PR #447, opened 2026-09-14
ntp: keep the time from an NTP server

DrDoof:
These switches have no battery backed clock, so this adds an SNTP client. It builds on the resolver from #446 for server names, so it includes those two commits until that one is in.

"ntp server", "ntp interval", "ntp timezone <+hh:mm>", "ntp dst off|eu|us" and "ntp on|off" set it up, and "time" prints the local time. There's also a Time tab in System Settings with a running clock and the last synchronisation.

The date is split with a shift and subtract division, so no 32-bit division from the library ends up in the common area. I compared the local time with Python's zoneinfo for Europe/Warsaw, America/New_York and Asia/Kolkata, including a second on either side of every DST change up to 2037, and on a SWTGW218AS it matched a synced host to the second.


vDorst:
I am currently missing the purpose for this feature. We add a lot of code for only 1 number and maybe a lot of computation time to make it human friendly.
I think for now a event timestamp is just a snapshot of the uptime-timer.
When shown in the browser, let the browser do the heavy lifting.

DrDoof:
Thanks for having a look. At the moment the time only shows up in the `time` command and the clock on the System page, so I can see why it looks like a lot of code for one number.

What I was really after is syslog. RFC 3164 has a timestamp field that we don't fill in, so the collector just uses the time a message arrives. Most of the time that's close enough, but lines buffered during boot arrive later with the wrong time, and that's exactly when you want it right.

There are a few other places it would help. When you look at the last topology change, a LACP change or a link flap from the CLI or from a script, there's no browser to do the conversion, and an uptime snapshot alone doesn't tell you when the switch booted. Further down the line, port or PoE schedules would need it too.

The syslog timestamp is the part that actually needs the date on the switch, and I haven't added it yet. I'm happy to add it here so the code has a real user. If you'd rather keep the firmware small, I can also cut this back to just the sync and let the browser handle time zones and DST. Whatever works better for you.

vDorst:
Thanks for pointing me to other features were it is also needed or useful.
I didn't think about these other options.
I am putting this PR on my review list, this may take a while because you created a lot of PRs.
So I hope other people can also help reviewing.

Are we hitting the year 2038 issue? Or it this not an issue when using the full 32-bit for the timer?

DrDoof:
No worries about the queue, I know it's a long list, and thanks for putting this one on it.

On 2038: we should be fine. The time is kept as an unsigned 32-bit count of seconds since 1970, so it runs until 2106 rather than stopping in 2038. NTP sends 32-bit seconds since 1900, which roll over in February 2036. Subtracting the 1970 offset in unsigned 32-bit arithmetic carries straight across that, so no special case is needed. Uptime comes from the 32-bit seconds counter, and the calendar code treats 2100 as a common year.

I ran the conversion and date code on the host against Python at the 2036 rollover, at 2038, around February 2100 and at the last second of 2106, and the results matched.

logicog:
> I am putting this PR on my review list, this may take a while because you created a lot of PRs.
> So I hope other people can also help reviewing.

I am on a trip to Asia with very little connectivity and no way to test anything, but will be back next week. I wanted to look into RLDP and storm control, then.

PR #448, opened 2026-09-14
stp: accept the documented port keyword again

bloqaudio:
c1ad800 (#397) turned `stp port <n>` into `stp ent <n>` in the parser only. `doc/stp.md`, `html/main.js` and every saved startup config still spell it `stp port`, so per-port settings are rejected on the console and silently dropped when the startup config is replayed at boot. This restores the keyword; `stp lag <n>` is unchanged. Two lines in `rtl837x_stp.c`.


vDorst:
Thanks

PR #449, opened 2026-09-14
Move DHCP out of BANK2 into BANK3

bloqaudio:
Since #397 BANK2 is 49182 bytes on `PCB_SWTG018AS_V2_1_0`, 30 over the 48 KB window, and the image builder refuses to link (`Error: Bank 2: code segment too large at 0x30000!`). CI only builds `KP_9000_6XHML_X2`, so it did not show. `dhcp.c` is self-contained: its three exports are `__banked`, the rest of the module is only reached from inside it, it holds no `__code` pointers, and its uip calls already crossed into BANK1, so moving the pragma is the whole change. On `PCB_SWTG018AS_V2_1_0` BANK2 goes 49182 to 43691, BANK3 5709 to 11206, BANK1 +6 for the trampolines; the generated assembly shows no plain `lcall` leaving the module's bank. All 35 machines in `machine.h` build and link; BANK2 lands between 43580 and 43686 bytes on every one of them.


feelfree69 (approved):

vDorst:
Thanks for fixing

PR #450, opened 2026-09-14
web: stop polling when the page has had no input for ten minutes

bloqaudio:
Follow-up to the session change that went in with #429 (raised by @vDorst on #428). The firmware counts every authenticated request as activity and the page polls the status every 2.5 s, so an open tab never let the session expire. This tracks the last mouse, key, touch or wheel event and treats the page as hidden once ten minutes have passed without one: the pollers keep ticking but send nothing, the firmware session times out as it did before, and the next input resumes polling, which meets the 401 and returns to the login page. `html/app.js` only. The mock suite gains two checks (polling pauses when idle, resumes on input) and passes raw and minified.


vDorst:
Thanks for fixing

PR #451, opened 2026-09-14
HASIVO_S600W_4GT_2XGT_SE machine + sysled control

SkullyDev:
Added config for HASIVO_S600W_4GT_2XGT_SE (4x2.5G, 2x10G)
SysLed behaviour logic added so it can mimic any original style


machine_init.c
```diff
@@ -117,6 +117,13 @@ void machine_custom_init(void) __banked
 	REG_SET(RTL837X_REG_LED_GLB_IO_EN, 0x7624155b);
 }
 
+#elif defined MACHINE_HASIVO_S600W_4GT_2XGT_SE
+void machine_custom_init(void) __banked
+{
+    // For this device, the reset button triggers this input
+    gpio_input_setup(GPIO54_ACL_BIT2_EN);
+}
+
```
vDorst, line 126: This is not needed.
In early_boot_handle_button() this pin is set to input.

machine.c
```diff
@@ -1226,6 +1226,47 @@ __code const struct machine machine = {
...
+            LEDS_2G5 | LEDS_LINK,
+            LEDS_1G | LEDS_100M | LEDS_10M | LEDS_LINK,
+            0,
+        },
+        {
+            LEDS_10G | LEDS_5G | LEDS_2G5 | LEDS_1G | LEDS_100M | LEDS_10M | LEDS_LINK | LEDS_ACT,
+            LEDS_10G | LEDS_LINK,
+            LEDS_5G | LEDS_2G5 | LEDS_LINK,
+            0,
+        },
```
vDorst, line 1256: Add comment which ledset controls which port-type and also colors.

rtlplayground.c
```diff
@@ -1747,7 +1751,11 @@ void main(void)
 	print_cmd_prompt();
 	idle_ready = 1;
 
-	set_sys_led_state(SYS_LED_ON);
+	if (machine.sys_led_mode) {
+		set_sys_led_state(machine.sys_led_mode);
+	} else {
+		set_sys_led_state(SYS_LED_ON);
```
vDorst, line 1757: Are you sure that stock software is setting this?
I can imagine that the software is controlling the led as a hearth-beat led, if the software is stuck, it stops blinking.
I personally I don't like this feature in this way for now.

vDorst (commented):
See comments

vDorst:
Thanks

PR #453, opened 2026-09-15
stp: MSTP with up to 15 instances

DrDoof:
This adds MSTP (802.1Q) on top of the RSTP work in #444, so it includes those commits until that one is in.

"stp version mstp" turns it on. "stp region", "stp revision" and "stp msti <n> vlan <list>" set up the region, and "stp msti <n> prio" and "stp port <n> msti <m> cost|prio" set the instances. "stp mstp" and "stp msti <n>" show the state. The Spanning Tree page gets the region fields with the bridge settings and an MST instances card.

The switch has a port state register per instance (0x5310 + 4 * instance) and keeps the instance of a VLAN in bits 20-23 of the VLAN table entry; both read back what is written on a SWTGW218AS. Inside a region the instances run their own trees; at a boundary port they follow the CIST. The configuration digest is worked out one MD5 block per timer tick, so a table change never stalls the main loop.

Commands that only show state no longer go into the command history, otherwise a page polling them pushes unsaved settings out of it.

Tested with the host simulator in test/stp (75 scenarios, the old RSTP ones unchanged) and the digest against Python's hmac. In ucsim a tick with 15 instances takes about 3.5 ms. The BANK3 version ran on my SWTGW218AS in RSTP mode, where the page showed the default digest worked out on the switch, but not against another MSTP bridge.

BANK1 has under 500 bytes left on main, so the protocol, the MSTP module and the Spanning Tree JSON handlers now live in BANK4 (`httpd/page_stp.c`, with its own copies of the output helpers and the JSON header, so nothing there calls into BANK1 directly or passes it a pointer into its code space). On KP_9000_6XHML_X2: BANK1 45539, BANK2 38444, BANK3 11154, BANK4 33487 bytes. The bank check finds no plain call across banks and the simulator passes. On a SWTGW218AS, built together with the LACP branch and running in RSTP mode, it came up with every port forwarding, the root behind the uplink and the LAG as one port of the tree; `/stp.json`, `/stpcnt.json` and `stp status` all answer from BANK4, and traffic goes through the LAG and a second port group.


DrDoof:
Rebased on main and pushed ten more commits. Before asking anyone to read five thousand lines I wanted to go through them once more myself, slowly, with the standard open next to the code, and I am glad I did: the simulator was green and still hiding a handful of real bugs. Sorry for the size of this update; I would rather it lands here than in someone's network.

What changed:

- A port that leaves the protocol now leaves the region's trees as well. The BPDU guard, the root guard and `stp port N off` only touched the CIST on a port internal to the region, so the instances kept forwarding through a port the status page called tripped, and a port switched off could get stuck discarding in an instance started later. Same root cause, three faces.
- A topology change inside an instance now reaches the boundary ports too. They forward that instance's VLANs, but were never flushed and never passed the change on, so stale entries lived until L2 aging.
- Port times start as the bridge's own, the way 802.1Q initialises them. Before, a port that had only heard the neighbour's CIST root port never had a hello time to age instance information with, and the instance never took that neighbour as its regional root.
- The digest is now worked out in full before it goes into a BPDU, at `stp on` and on every table change. I had been proud of the one-block-per-tick engine, but it left 2.6 s where BPDUs described a table the switch did not have, and neighbours saw the port hop between regions. It is about a third of a second once per configuration change now; the incremental steps only pick up a change made while STP was off. The description above still praises the old behaviour; I will fix that wording once this settles.
- `stp mstp` skips instances without VLANs; it walked the whole VLAN table fifteen times and held the main loop for a fifth of a second, and the page was polling it. The MST card is now read when the tab opens and after its own Apply.
- The startup config merge learned the MST commands, so Apply replaces the previous line instead of appending another one.
- Five new simulator scenarios, 76 to 80, each failing on the code before these commits, and the simulator now runs from `make -C test` like the other host tests.
- Two small ones: forward delay and max age of every tree come from the CIST root, and a pointer-difference warning is gone.

All machines build, no cross-bank calls, 80 of 80 scenarios pass. Still not run against another MSTP bridge; that is the next thing on my bench.

DrDoof:
One more thing I found while going back over the region handling: `stp region` and `stp revision` only wrote the variable. Both change what this bridge sends as its MST Config ID, so from that moment every neighbour we had agreed a region with is a stranger - yet the ports stayed classified as internal until the next BPDU said otherwise, and the instances kept running on a boundary that no longer existed.

Both commands now give up the boundary straight away: every port goes back to carrying its CIST state in the instances, the information and timers held for it are dropped, and a BPDU with the new identity goes out without waiting for a change of its own. Nothing stops forwarding while this happens, since a port leaving the region simply takes over the CIST state it already had.

Scenario 81 covers it. Four of its five checks fail without the change, the state one included, which is how I know it is testing something rather than passing by accident.

PR #456, opened 2026-09-15
rldp: find loops on the switch ports and block the looped port for a minute

DrDoof:
Hi! Now that storm control is in the queue, this is its natural companion: loop detection. The switch can already do RLDP in hardware, it just was never switched on, so this wires it up.

With "rldp on" the switch sends small test frames carrying its own MAC address out of its ports. If one of them comes back, that port is part of a loop, say two ports cabled to the same unmanaged switch, or a cable accidentally plugged back into the switch. The firmware then blocks that port by turning off TX and RX in its MAC control register, so the link stays up but nothing gets through. When two ports loop into each other, only the higher numbered one is blocked. After a minute the port is opened again and checked, so a loop that has been fixed clears up on its own. "rldp <port> off" leaves a port out, and "rldp" shows what is going on. There is also a small Loop detection card on the Spanning Tree page, and /rldp.json has the same data.

One thing I learned the hard way: my first version ignored STP, and on my SWTGW218AS it happily reported the uplink as looped through the redundant link that STP was keeping blocked. That would have cut the switch off for a minute. So now it stays out of STP's way: test frames only go out of ports STP forwards on, and a port is never blocked when the other end of its loop is a port STP is not forwarding on. The frames are matched on the magic value only, because the random ID changes every time the firmware asks the switch for a random number.

I tried it on my SWTGW218AS without a loop: the forwarding ports send test frames, nothing gets reported or blocked, and "rldp off" puts everything back. All 34 machines build and the stack is unchanged. I haven't had a real loop cabled up on the hardware yet, so I'd be glad about any testing, and happy to adjust anything.


DrDoof:
I finally got a cable between two ports and watched what happens. Good news first: the detection is solid. The loop showed up as the pair 2-3 and the higher port went down within seconds. My uplink was reported as looped too, since there is a second path to the neighbour switch, but it was never blocked, so the forwarding check does exactly what it should.

Two things bothered me, and both are fixed here.

A blocked port used to come back after a minute no matter what, walking straight into the loop again. A ping across the switch lost 4 to 10 seconds of traffic seven times in twelve minutes. Now a port needs a link to be blocked at all, and while it sits blocked its link is watched: pull the cable and it is back on the next tick, which is how these things usually end anyway. If the loop is still there the wait doubles, a minute, then two, then four. One reopen in two still costs a few seconds, I won't pretend otherwise, but they come far less often.

The other one was only cosmetic, but unnerving: the loop bits live in a register and stay set after the loop is gone, so `rldp` and rldp.json kept listing ports as looped with the cable out, and even after rldp off. Both now report a loop only while the detector runs and the port has a link.

One story worth passing on, because it cost me a good part of the evening. My first attempt checked the link only when reopening a port, not before blocking one. The register still held the loop bits, so the switch blocked a port with no link and reopened it a second later, over and over. The code looked perfectly fine. What gave it away was the console printing `port 3 opened` once a second, and what convinced me it was gone was fifteen seconds of complete silence after pulling the cable.

Not tested: the JSON endpoint and the web page, which would have taken the only TCP slot, and a loop on a port that STP still controls.

PR #457, opened 2026-09-15
flash: bail out of reads of zero length

DrDoof:
Hi! This is a small one, but it bit me hard on my SWTGW218AS a few days ago.

flash_read_bulk() only stops its copy loop when the remaining length hits 1, 2, 3 or 4. With a length of zero none of those ever matches, the counter wraps around and the loop writes 65536 bytes into the destination, which wipes the whole XDATA space. The switch keeps forwarding, but the management VLAN, the STP flag and the IP address are gone, so it becomes unreachable until a reboot.

The easy way to get there is send_config(): if the stored configuration starts with a NUL byte, for example after uploading an empty config file, valid_len is 0, and from then on every visit to the configuration page knocks the switch off the network. I reproduced it that way, and with the check in place the same request just returns an empty configuration and everything stays up.

The second commit adds the same check to flash_read_security(). Its only caller today is the CLI with a fixed length of 40, so nothing triggers it, but the loop has the same shape and the check belongs in the function.


feelfree69 (approved):
Thanks!

rtl837x_flash.c
```diff
@@ -222,6 +222,10 @@ void flash_read_jedecid(void)
 void flash_read_bulk(__xdata uint8_t *dst)
 {
 	short status;
```
vDorst: This is unused. Left over from #427.
I have seens it to late too correct the issue.

So I think we should do CI-build that warns on warnings.
Also the build-test from #324, also doesn't show-up in any PR.
So it seems that it doesn't build on a PR request.
I hope someone can also fix that.

DrDoof: Thanks, good catch. It shows up as warning 85 on main as well. I've pushed a third commit here that drops it, so the build has one warning less now.

About CI: build.yml only runs on `push`, so a PR opened from a fork never gets a run, and the host tests from #324 sit in the same job. Adding a `pull_request:` trigger should be enough for both. I can open a small PR for that if you like. Turning warnings into errors would need the existing 72 cleaned up first, so it might be easier to just fail when the count goes up.

vDorst: > Thanks, good catch. It shows up as warning 85 on main as well. I've pushed a third commit here that drops it, so the build has one warning less now.

Thanks

> About CI: build.yml only runs on push, so a PR opened from a fork never gets a run, and the host tests from https://github.com/logicog/RTLPlayground/pull/324 sit in the same job. Adding a pull_request: trigger should be enough for both. I can open a small PR for that if you like. 

Yes please make a PR. 

> Turning warnings into errors would need the existing 72 cleaned up first, so it might be easier to just fail when the count goes up.

Do we still have 72 warning? AFAIK we don't have any warning anymore.

DrDoof: I checked again and the 72 was from my own setup. I build locally with SDCC 4.6.0, and it adds warnings 356 and 357 about const in code space. The CI log for main on 4.5.0 has just one warning, the unused `status`, and this PR removes it. So once it's merged main builds clean, and failing CI on warnings could be switched on straight away.

Here's the PR for the trigger: #461

vDorst:
Thanks

PR #458, opened 2026-09-15
httpd: do not accept a session cookie before the first login

DrDoof:
session_id is cleared at boot and only a successful login fills it in, while the uptime counter and last_session_use both start at 0. So for the first 200 seconds after a start, a request carrying an empty `Cookie: session=` passes the check and counts as logged in, which is enough for /cmd, /config and /upload.

It also doesn't stop at 200 seconds: every accepted request moves last_session_use forward, so as long as requests keep coming, the empty cookie stays valid until someone logs in.

The fix only looks at the cookie once a session has actually been handed out.

I tested the check on the host with is_word_x and the session code taken as they are. Before the fix, `session=` and `session=;x=1` were accepted. After it, both are rejected, and a real login, a wrong id and an expired session behave as before. On SWTGW218AS this adds 6 bytes to BANK1, the bank check is clean and there are no new warnings.

This touches the line right next to the one #452 changes, so whichever goes in second needs a one-line rebase. With a configurable timeout the window would otherwise get longer, so I'd keep the two separate.


vDorst:
Good catch, thanks

PR #460, opened 2026-09-15
httpd: keep a continued upload away from the cmd and login handlers

DrDoof:
Hi! Another small one from reading through httpd.c.

handle_post() takes the request path from byte 6 of whatever segment it gets. For the first segment of a POST that really is the path, but once a multipart upload is under way the later segments are just file contents, and the path pointer lands in the middle of them. The cmd and login checks came before the check for an upload in progress, so a config or firmware upload segment that happened to have "cmd " at that spot would run the rest of the segment as CLI commands, and one with "login " there would end the upload with a 400. It depends on where the TCP segments break, so it is rare, but for a given file it happens every time.

The fix just moves the upload branch in front of cmd and login. A new POST can't match both, so nothing changes for anything that isn't an upload.

All 34 machines build and the stack is the same as main. I haven't forced the case on the hardware, since that means placing the bytes exactly at a segment boundary.


bloqaudio (approved):
Read it against current main and built it. The ordering is right and I agree it should go in.

The reasoning that it is safe for everything that is not a continuation segment holds up in the code. The moved branch matches on three things: `tstate == TSTATE_MULTIPART`, or a path of `upload`, or a path of `config`. A fresh `cmd` or `login` POST is a new request, so its `tstate` is not `TSTATE_MULTIPART` and its path is `cmd` or `login`, not `upload` or `config`. It cannot match the moved branch and falls through to the same handler it reached before. The only requests whose routing changes are continuation segments of an upload, which is exactly the set that was landing in the wrong handler.

One thing I checked for a regression rather than assuming: a genuine first segment `/upload` or `/config` POST now also enters this branch. It already did on main, where it was the third arm of the same chain, and the upload and config setup that assigns `uptr`, `verify_crc` and the rest runs earlier in `handle_post()` before this decision is reached, so moving the branch does not skip it. First segment uploads route as before.

No size cost, as expected for a reordering. On SWTGW218AS the stack, `BANK1`, `BANK2` and `CSEG` are all identical to main.

On testing: like you, I did not force the misroute on hardware. It requires the token bytes to fall at byte six of a continuation segment, which means placing content at a TCP segment boundary, and the bug is established from the routing order in `handle_post()` without needing to stage that. What I did confirm on hardware earlier this week, on a related path, is that ordinary uploads stream and are bounded correctly, so the continuation path itself is exercised in normal use; this change only removes a wrong turn out of it.

Approving.

feelfree69:
Thanks!

PR #461, opened 2026-09-15
ci: build pull requests too

DrDoof:
The workflow only runs on push, so a pull request opened from a fork never gets a build, and the host-side unit tests from #324 run in the same job, so they were skipped too. This adds a pull_request trigger next to the existing push one.

Pull requests from branches inside this repository will now build twice, once for the push and once for the PR. I left the push trigger alone so nothing changes for branch builds.

Once #457 is in, main builds without warnings on the SDCC in the workflow, so a step that fails on any warning could go in as a follow-up.


vDorst:
Thanks

PR #462, opened 2026-09-15
web: keep stored config lines and per-port ingress when saving

DrDoof:
Hi! Saving from the web page could quietly drop parts of the startup config. Three things fixed here:

- Save filtered the stored config too, so lines the page doesn't know just vanished. On my switch that was "lag 1 lacp 7 8" and a port name with a space. Stored lines are now kept as they are.
- Ingress was replaced as a whole, but the command only touches the ports it lists, so setting one port wiped the others. It's merged per port now.
- Write from the config editor cleared the command log, losing changes made on other pages. Only Save clears it now.

I also doubled the command log to 2 kB, since a few STP port settings could already fill 1 kB.

Tested the merge against my switch's config plus some edge cases, and the save and write flows in the browser. All 34 machines build.


bloqaudio (approved):
Tested on my board with this merged onto current main `fa93340`.

**Save on hardware.** Stored config: the normal lines plus `ntp server 192.168.0.1` (unknown to the page), `port 4 name up link` and `ingress 1a 2a`. I set port 3 to accept all on the VLAN page and saved from the browser. The written config kept every stored line, including both of the ones the page does not know, and the ingress line came out as `ingress 1a 2a 3a`. The command log was cleared afterwards.

**Merge logic.** I ran `mergeConf` from main and from this branch side by side in node on the Save path. The differences are only the ones you intend: unknown stored lines are kept, a port name containing a space is kept until the log replaces it, and ingress merges per port. A global `ingress u` in the stored config correctly drops the per-port entries before it. VLAN delete and the `vlan 1` / `vlan 10` prefix case come out the same as on main. One side effect, not blocking: duplicate stored lines are no longer collapsed, because the stored config is no longer run through the merge. Only a hand-edited config would have those.

**Command log.** After 70 commands of 39 bytes, `/cmd_log` returns 2041 bytes in 51 whole lines, which is more than one TCP segment and arrives intact. The same sequence on main returns 1001 bytes. Cost on `SWTGW218AS`: XSEG +1024 (the buffer), BANK1 +18, stack unchanged.

I checked the change to Write from the editor by reading the code only.

feelfree69:
Thanks!

PR #463, opened 2026-09-15
cmd: report a rejected command as an error and keep it out of the history

DrDoof:
Hi! POST /cmd already answers 400 when err_status is set, but only the tokenizer ever set it. Every "Error:", "Invalid ..." or "Unknown command" from the parsers came back as 200, so the web page showed "Applied", marked the config as unsaved, and the rejected line went into the command history and from there into the saved config. "pvid 3 4095" is an easy way to see it.

The parsers now set err_status on every path that rejects a command, the history only records commands that went through, and a body with several commands reports the first failure instead of losing it to the next line. Warnings such as a full VLAN name table stay warnings.

Host tests pass and all 34 machines build.


vDorst:
I think this pattern below can use a helper function.
```c
  err_status = ERR_INVALID_ARGUMENT;
  print_string("Error: <Some Error Msg>\n");
```

Properly need a better name.
```c
  cmd_invalid_and_print("<Some Error Msg>");
```

DrDoof:
Good idea, that pattern was in 41 places. Added `cmd_error()`, which sets the status and prints the message, so the two can't drift apart. Bank 2 drops by 235 bytes. The seven remaining `err_status` assignments are the tokenizer paths that print nothing.

The helper itself added two warning 357, because a `__code char *` parameter isn't const. Rather than cast it away, `print_string` and its two variants now take `__code const char *`, which takes the build from 76 warnings to 70 and needed the matching signature in `test/support.c`.

cmd_parser.c
feelfree69: Comment belongs to `parse_mac_cmd`

cmd_parser.c
```diff
@@ -1620,8 +1594,7 @@ void parse_syslog(void)
 	}
 	else
 	{
-		err_status = ERR_INVALID_ARGUMENT;
-		print_string("Error: syslog [on|off|ip [ip-address]|port [number]]\n");
+		cmd_error("Error: syslog [on|off|ip [ip-address]|port [number]]\n");
 		print_string("  on/off enables or disables syslog, ip sets the syslog server IP address,\n");
 		print_string("  port sets the destination UDP port (default 514)\n");
```
vDorst: My vision was also remove the "Error: " from every error-line and let the new helper print that part.
And how long that a text message me?
Above example now has `print_string()` for every line, which can just be one, when we have no limits.

cmd_parser.c
DrDoof: Good catch, thanks. Moved the helper above the comment so it sits with parse_mac_cmd again.

cmd_parser.c
DrDoof: Done, the helper prints the prefix now. The messages dropped their own "Error: ", and the ones that said "usage: " dropped that too, so they all read the same way. Two of them started with "Error setting" and "refusing", which would have read twice, so I reworded those.

The syslog block is one string now instead of three calls.

On length: print_string has no limit of its own, but a command answered over HTTP is captured into the 2500 byte response buffer, and the firmware appends "[output truncated]" if it doesn't fit. On the serial console there is no limit at all.

vDorst:
Thanks

PR #464, opened 2026-09-15
httpd: don't let a quote in a port name break the web page

DrDoof:
Hi! Name a port `NAS"1` and the web page stops loading: the name goes into status.json as is, so the JSON no longer parses, and since it's saved in the config a reboot doesn't fix it. VLAN names and SFP strings have the same problem.

These now go through a small helper that swaps `"`, `\` and non-printable bytes for `.`, same as hostname. The length doesn't change, so the size checks still hold, and names already in a config show up fine again.

Checked it with a JSON parser on the host, and all 34 machines build.


httpd/page_impl.c
```diff
@@ -185,7 +193,7 @@ void send_sfp_info(uint8_t sfp)
 			continue;
 		uint8_t c = sfp_buf[i & 0xf];
 		if (c && c != 0xa0) // a0 is the byte read from a non-existant I2C EEPROM
```
vDorst: Do you think this check is still valid.
`if (.. && c != 0xa0) // a0 is the byte read from a non-existant I2C EEPROM`
If there is no EEPROM then I2C transaction fails and it could not go thru this point.

vDorst:
I am not sure what is better. If we just sanitize the value at parsing/on the input side, these values should be right. But if something goes wrong, it is nice to have.

mintsoft:
> I am not sure what is better. If we just sanitize the value at parsing/on the input side.

imo, this is best. Limit the input to known good characters, simplifies the handling into one place?

DrDoof:
Thanks, that nudged me to look at it properly. The input side turned out to be one place, not three: the console, POST /cmd and the saved config all end up in the command parser, so port and VLAN names get filtered there now, the same way the hostname already was. That's the second commit.

I kept the escaping for the SFP strings. Those come off the module's EEPROM and never pass through a command, so there's nothing to catch on the way in.

About the 0xa0 check: sfp_read_block gives up when the controller flags the transaction, so that byte shouldn't get that far, and the filter turns it into a dot anyway. I've taken the special case out. I couldn't try it with an empty cage though, the one SFP slot here has a module in it.

httpd/page_impl.c
```diff
@@ -854,13 +862,13 @@ void send_status(void)
...
-					outbuf[slen++] = sfp_module_vendor[sfp][s];
+					json_char_to_html(sfp_module_vendor[sfp][s]);
 				slen += strtox(outbuf + slen,"\",\"sfp_model\":\"");
 				for (uint8_t s = 0; s < 16 && sfp_module_model[sfp][s]; s++)
-					outbuf[slen++] = sfp_module_model[sfp][s];
+					json_char_to_html(sfp_module_model[sfp][s]);
 				slen += strtox(outbuf + slen,"\",\"sfp_serial\":\"");
 				for (uint8_t s = 0; s < 16 && sfp_module_serial[sfp][s]; s++)
-					outbuf[slen++] = sfp_module_serial[sfp][s];
+					json_char_to_html(sfp_module_serial[sfp][s]);
```
vDorst: I think it is better to sensitize them already when the buffers are filled, in `sfp.c:103`.
So we can even replace these for-loops with a strcpy.

Ee do have a protoype in rtl837x_common.h `uint16_t strcpy(__xdata uint8_t *dst, const char *s);` but no real code anymore.

DrDoof: Done. sfp_read_field filters while it fills the buffers, so each of the three loops is one strcpy now. NUL still ends a string, so a field shorter than its slot reads the same as before.

strcpy had the prototype and no code, so the library one was being linked: the map shows it coming from _strcpy.rel before and from rtlplayground now. Mine returns the length, which is what the pages want. Bank 1 drops 17 bytes, bank 3 gains 19 for the filter, and the common area ends up 3 bytes smaller.

json_char_to_html stays for send_sfp_info, which reads the EEPROM straight into the page.

DrDoof: sfp_read_field filters the strings as it fills the buffers now. The three loops are one call each, though not strcpy: rtl837x_common.h hides that prototype under RTLP_HOST_TEST so host builds take the C library version, and adding its result to slen does not compile there. That is what my first push broke in CI. A small static helper does the copy instead.

Bank 1 drops 96 bytes, bank 3 gains 19 for the filter.

json_char_to_html stays for send_sfp_info, which reads the EEPROM straight into the page.

I rebased on main as well, so the suites from #467 run here.

cmd_parser.c
```diff
@@ -440,6 +440,14 @@ static void cmd_error(__code const char *msg)
...
 // reboot. As a regular command a "mac ..." line in the startup config applies
 // on every boot (execute_config runs before the final static L2 entry).
+static uint8_t name_char(uint8_t c)
+{
+	if (c < 0x20 || c > 0x7e || c == '"' || c == '\\')
+		return '.';
+	return c;
+}
+
+
```
vDorst: The comment is separated from `parse_mac_cmd()` again.

DrDoof: Fixed in a7fc830 - the helper now sits above the comment, so the comment is next to parse_mac_cmd() again. Generated code is unchanged.

httpd/page_impl.c
DrDoof: Good point, and it is already gone in this branch: the check is now just `if (c)`, with json_char_to_html() doing the filtering. So the 0xa0 case no longer needs its own test - anything outside printable ASCII comes out as '.'.

One small behaviour change worth naming: a 0xa0 byte used to be skipped entirely, now it shows up as a dot. Dropping the byte instead is a one-line change if you prefer the old output.

vDorst: I think it is fine. Corrupt SFP modules is already a issue that they may not follow the spec anyway. So we are not sure what to get anyway.

vDorst:
Thanks

PR #465, opened 2026-09-16
add support for ONT-S207CW-62TS-SE and identical Binardat-2G06-04210GSM

andreas5232:
I've ported the RTLPlayground to another version of 2.5G/10G switches.
Any help or feedback welcome, as this is my first contribution here.
Currently the switch is fully working, but the LEDs have the wrong colors.
Additionally I'm not quite happy with my device documentation yet.
I'll try to look into this soon and remove the PR's draft state.

Based on the findings here: https://github.com/emesix/ONT-S207CW-62TS-SE

Cheers,
Andreas


th0m4sek:
Supported devices table is alphabetical.
If this devices use same PCB as already existed in documentation i will be better to extend info in that PCB page not creating new page.

andreas5232:
@th0m4sek thanks for looking into my PR! :-) 

As that's my first time doing some deeper chip research and SOIC8 device flashing, I'm still a little bit overwhelmed by all the different aspects.

As soon as I finally get the LED assignment right, I'll try to map it onto an existing devic to keep the PR as clean and short as possible.

Also I'll double check everything, when I'm finished, to make sure it will be a useful improvement of this project.

vDorst:
Intressing that this board has JTAG markings.
I hope you can find out with pins they are connected too.
The datasheets I have seens, don't mention anything about JTAG. It could be that JTAG is only used for boundary-scan (testing is SOC-pin make connection to the PCB)-test, and you may can't access the CPU withit.

Note about the documentation.
CPU is RTL8372. Only 72 has two SFP possebilityies.
Ethernet phy is not phys but magnetics, this is what creats the +kV high-voltage isolation in ethernet.
Ethernet phys are internal in the chip.

afbeelding

doc/devices/ONT-S207CW-62TS-SE.md
```diff
@@ -0,0 +1,207 @@
+# ONT-S207CW-62TS-SE
+
+## Overview
+ONT-S207CW-62TS-SE is a managed switch with **4x 2.5G RJ45 ports** and **2x 10G SFP+ ports**.
+
+**Note:** This device is **identical in hardware** to PCB-SWTG024AS-A-2.0.1. The only difference is the brand label.
+
+## Compatible Devices
+
+The following devices use the **same PCB-SWTG024AS-A-2.0.1 hardware** and are fully compatible with the same firmware configuration:
```
feelfree69, line 10: Sure? From what I can see, your device is VERY special.
It has SFP ports on the left, and the whole PCB is mounted upside down.

doc/devices/ONT-S207CW-62TS-SE.md
```diff
@@ -0,0 +1,207 @@
...
+| LED Color | Speed |
+|---|---|
+| **Green** | 10G |
+| **Orange** | 2.5G / 1G |
+| **Off** | No link / No SFP module |
+
+## RTLPlayground Configuration
+
+### machine.h
+```c
```
feelfree69, line 78: it doesn't make sense to copy sourcecode into documentation.

doc/devices/photos/SWTG024AS-v2.0-unmanaged/binardat-2g06-04210gsm-back.jpg
feelfree69, line 1: Why is every photo duplicated?

doc/devices/ONT-S207CW-62TS-SE.md
andreas5232: I'm still debugging and testing a lot. At least all the basic functions were fully working with old web UI, mixed up LEDs and reversed Port order in web UI. All ports have been able to ping switch ip and reach web UI.

doc/devices/photos/SWTG024AS-v2.0-unmanaged/binardat-2g06-04210gsm-back.jpg
andreas5232: Thanks for pointing this out. That is definitely not on purpose. As this PR ist still WIP, I will be doing some more polishing, as soon as I got a fully working state. :-)

doc/devices/ONT-S207CW-62TS-SE.md
feelfree69: > reversed Port order in web UI

Yes, that's probably because the PCB is mounted upside down.
Because of that, this device differs more than any other from the existing designs. This is somehow the opposite of what you wrote about "fully compatible", and it does NOT use 
"same PCB-SWTG024AS-A-2.0.1 hardware".
If the latter would be the case, you would not even need a seperate machine definition for your device.

doc/devices/ONT-S207CW-62TS-SE.md
andreas5232: I'm struggling a little bit with the project structure at this point. I was expecting either a separate config file per device and not a single file with multiple if/else parts. Also I'm not very happy with the quality of my own PR yet.

doc/devices/SWTG024AS.md
```diff
@@ -1,4 +1,5 @@
-### SWTG024AS
+# SWTG024AS
```
feelfree69, line 1: Same comment as before: I believe you device is more differenzt than any other of the existing devices. I do not see a single sign that the PCB has anything to do with the existing SWTG024AS devices. Please revert all changes in the file.

(Maybe you were mislead because you started with the config for that device; and as this worked (port forwarding, WebUI, ...) you thought it is the correct/same device? In fact you could use ANY of the machine definitions - first 4 RJ45 will probably work in any case)

doc/devices/ONT-S207CW-62TS-SE.md
andreas5232: Alright, I understand. I was trying to implement the device support with just a few changes. But if it's necessary because of the port mapping being completely different, I will of course integrate it as a separate device.

What I meant with "compatible" is the fact, that you can get an almost running version with it.

I've done some openwrt stuff and serial flashing, but this PCB based development using SOIC8 is new to me.

feelfree69: > What I meant with "compatible" is the fact, that you can get an almost running version with it.

Yes, but this you can probably achieve with ANY of the existing machine definitions. The first 4 RJ45 ports are handled (except for LEDs) entirely within the SoC, so they do not differ at all between the machines. 

> , but this PCB based development using SOIC8 is new to me.

No problem, all new contributors are welcome!

andreas5232: That's a good hint. Both of the SFP+ were also working. I'll probably have to inspect the original firmware for led config and will then try to get this PR cleaned up and polished.

Thanks for your help!

PR #472, opened 2026-09-18
httpd: account for the bytes actually sent

DrDoof:
I went looking into static files arriving corrupted when the browser is slow to read them, and it turns out `httpd.c` has been trusting `uip_mss()` to tell it how much it last sent. It cannot do that: uIP recomputes that number from the window the peer advertises in every incoming segment, and it does so before calling us. By the time we look at an ACK, the value describes the window right now, not the segment being acknowledged.

While the window stays put the two numbers are the same, which is probably why this has been quiet for so long. As soon as it moves between a segment going out and its ACK coming back, the bookkeeping drifts:

- shrinking 1460 -> 600: we send 860 bytes a second time under fresh sequence numbers, and the browser cheerfully appends them to the file;
- opening 600 -> 1460: we skip 860 bytes instead;
- and on a retransmission after a shrink we copy fewer bytes than uIP actually puts on the wire, so the tail of the repeat is whatever was still sitting in the buffer.

It is easy to watch happen on real hardware. Fetching `/app.js` with a client that reads slowly (SO_RCVBUF 4 kB, 256 bytes every 50 ms) gives back 36253 bytes instead of 35833, with the first divergence at byte 2141. With this change, three runs in a row come back byte for byte.

The fix is small: remember the length in `tx_send()` and use it in both places. `sent_len` is only ever read after `tx_send()` has set it, since the ACK path runs in `TSTATE_TX` and uIP asks for a retransmission only while a segment is outstanding, which is the very segment `tx_send()` staged. Every call site already passes at most `uip_mss()`, so uIP never clamps and `sent_len` matches `uip_connr->len`, the length it will retransmit. Pleasant side effect: folding the six `uip_send(outbuf + o_idx, ...)` sites into one helper costs 2 bytes of XDATA and hands back 255 bytes of BANK1.

Two smaller things came along for the ride. The first commit casts `uip_appdata` in three subtractions, because mixing `void *` with `uint8_t *` is something SDCC accepts and C does not, and it was what kept this file out of host tests; the generated code is identical, the SDCC output differs only in the comments carrying the source text. The last commit adds `test/test_httpd_tx`, which builds `httpd.c` and `uip.c` unmodified and drives them through a handshake and a GET from a client that moves its window, comparing what reaches the wire with the file it should be. Its first scenario holds the window still and has to pass: that is the check that the test itself is honest before the moving-window ones are believed.

Happy to split any of this out if you would rather take the pieces separately.


vDorst:
I see one warning when building the tests.
```
../httpd/httpd.c: In function ‘scan_header’:
../httpd/httpd.c:352:72: warning: suggest braces around empty body in an ‘else’ statement [-Wempty-body]
  352 |                                 dbg_string("Invalid session cookie!\n");
      |     
```

Maybe we should also add a error on warning in this step.
So we don't break the tests.

DrDoof:
Good catch, thanks. Fixed in 42bbed9.

The else only has dbg_string() in it, and that expands to nothing unless DEBUG is on, so gcc sees an empty body. The line is older than this branch - the bench is just the first thing that compiles httpd.c with a host compiler and -Wall -Wextra, which is why it shows up now.

I added -Werror to test/Makefile as you suggested. That was the only warning the four test binaries produce, so nothing else had to change.

vDorst:
Thanks

PR #473, opened 2026-09-18
config, httpd: skip an over-long line and refuse to store one

DrDoof:
While the switch replays the startup configuration, one bad line ended the
whole replay: a line too long for cmd_buffer, or one the tokenizer rejected,
stopped it right there. Everything below was silently left unapplied, while
the file itself and the editor showing it still looked complete.

I ran into this here: a 130-byte line sitting after the last real setting was
enough to lose every setting that followed it on the next boot, and nothing on
the console or in the UI said the configuration had only been half applied.

Two commits:

1. execute_config() reports the offending line and carries on with the next
   one instead of stopping. The skip flag lives across the flash page refill,
   so a long line that crosses a page boundary is not resumed halfway and run
   as a truncated command.

2. The upload path and the editor refuse a line the replay cannot take.
   config_take() checks before the sector is erased, so a rejected upload
   leaves the existing configuration in place. The editor measures the line in
   bytes rather than characters, so a port name written in a non-ASCII script
   is counted the way the switch counts it.

The limit works out at 126 bytes per line, and the parser, the upload check
and the editor all agree on it.

I checked this on hardware with a deliberately over-long line in the
configuration: before the change the setting ahead of it was applied and the
one after it was missing; afterwards both are applied, re-uploading the same
file is refused with 400, and the flash is left untouched.


bloqaudio (approved):
Tested on my board (RTL8373, `PCB_SWTG018AS_V2_1_0`) with this merged onto current main `fa93340`.

**Before.** On main I stored a config with `port 1 name before`, then a 130 byte line placed to cross the 256 byte flash page at 512 (bytes 462 to 592), then `port 2 name after`. After a reboot port 1 was named, port 2 was not, and the console stopped at `ERROR: Command too long`.

**After.** Same config left in flash, new image booted: both names are applied, the console prints `ERROR: Command too long, skipped:` and goes straight on to `after`. Nothing from the part of the line past the page boundary gets run as a command, so the skip flag surviving the refill does what you describe.

**Upload.** A 127 byte line is refused with `400` and the stored config is left exactly as it was. A 126 byte line is accepted, and after a reboot it replays without an error. So the parser and the upload check agree at 126 on hardware. I did not try the editor check in the browser. The upload check sits behind it anyway.

Cost on `SWTGW218AS` against `fa93340`: BANK1 +90, BANK2 +77. One small thing, not blocking: `skipping` is a plain `uint8_t` automatic, and SDCC gives it a byte of direct RAM (`_execute_config_skipping_10001_552`), so the stack goes from 151 to 150 bytes free. Declaring it `__xdata uint8_t skipping` brings the stack back to 151 for 11 bytes of BANK2.

feelfree69:
Thanks!

PR #474, opened 2026-09-18
temp: read the on-chip thermal sensor

DrDoof:
The chip has a pair of thermal meters that nothing was reading, so this adds a `temp`
command and a row in the System page.

Getting a number out of them took a moment. The result register at 0x340 splits a 19 bit
value across two fields - bits 18:3 sit in the low half, bits 2:0 up in bits 18:16 - with
the sign at bit 18, degrees in 17:10 and a fraction below. The nice part of that layout is
that the degrees fall out of a shift of the low half, so the conversion stays 16 bit.

The bit that surprised us: the meter comes out of reset latched. The register holds
whatever was measured when the board powered up and then never moves - four samples 40 s
apart came back identical down to the raw ADC, and 0x340 matched the power-on register at
0x350. Clearing bit 2 of the control register at 0x338 during init is what makes it live.

On an SWTGW218AS it now reads 58.7 to 59.5 C while running, with the power-on register
still showing the 49.2 C it started from. For what it is worth, the scale lines up three
ways: the threshold register reads -40 and +125, the slope from warm-up works out at
4450 LSB/C, and the slope between the two meters at 4388 LSB/C.

We only read the meter that is up after boot. The other sits in reset with no coefficients,
so it would just repeat its power-on value - happy to wire that one up too if it looks
worth having.

The page gets the formatted string rather than the raw register, since the System table
prints values as they arrive.

Costs 231 bytes in BANK2 and 186 in BANK1.


feelfree69:
> On an SWTGW218AS it now reads 58.7 to 59.5 C while running,

This is somehow expected. While it may be interesting just out of curiosity for developers like us, an end-user will not get any help from that value. So I would not expose it to the WebUI.

DrDoof:
> an end-user will not get any help from that value

I think it does help a few kinds of end users:

- A switch in a small, closed rack with no airflow. The reading is the only way to see whether it runs hot.
- 10GBASE-T SFP+ modules draw close to 3 W each right next to the chip, and their own DOM reading does not show what that does to the switch.
- A fanless box in the attic, the garage or on top of a NAS in summer. The room changes with the season, the reading shows it.
- Someone adding a heatsink or a fan, or moving the switch, can see whether it helped.
- When links drop every few hours, one number tells us whether heat is worth ruling out before anyone looks at the code.

The value is also in the info JSON, so Home Assistant or other monitoring can pick it up and warn on it, whatever the page shows.

If it stays in the WebUI, I would put it at the very top, next to the Ethernet port status, so it is always in view.

feelfree69:
Tested with KP_9000_6XH_X2 and SWTGW218AS, both ok.
Open case, Heatsink of SWTGW218AS has double height, chip runs at 48°C, while KP_9000_6XH_X2 with the smaller heatsink reaches 57°C.

feelfree69 (approved):

PR #475, opened 2026-09-18
Makefile: run machine_check under a shell that has pipefail

bloqaudio:
The `machine_check` recipe opens with `set -eo pipefail`, but the Makefile never sets `SHELL`, so make runs it under `/bin/sh`. Where that is dash, which is the default on Debian and Ubuntu, the option does not exist and the recipe dies on its first line:

```
/bin/sh: 1: set: Illegal option -o pipefail
make: *** [Makefile:156: machine_check] Error 2
```

So the check cannot be run locally on a stock Debian or Ubuntu box. CI does not catch it because it runs in a container whose `/bin/sh` accepts the option.

This asks for bash on that one target rather than for the whole file, so the dependency stays where the option is actually used and no other recipe changes shell.

With the fix all 12 machines compile again:

```
Checking KP_9000_6XHML_X2
...
Checking ZX310S_4T2XT
```

Verified on Ubuntu 24.04 with `/bin/sh` linked to dash, SDCC 4.5.0.


tofurky:
This is working in dash 26.04 and trixie, not older versions though

from `apt changelog dash`:
```
dash (0.5.12-7) unstable; urgency=medium

...
  * Upstream patch:
    - Implement pipefail option (Closes: #1071238)
...
```

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1071238

bloqaudio:
Good catch, that explains why CI is green: the container is trixie, whose dash has the patch.

The version that fails here is `0.5.12-6ubuntu5` on Ubuntu 24.04.5 LTS, one revision below the `0.5.12-7` that closed that bug:

```
$ dash -c 'set -o pipefail'
dash: 1: set: Illegal option -o pipefail
```

24.04 is the previous LTS, superseded by 26.04, and it stays in standard support until April 2029, so it will keep biting people for a while yet. Happy to switch the fix to dropping `pipefail` instead of asking for bash, if you would rather not add the dependency: the only pipeline in the recipe is inside a command substitution, whose exit status the loop never checks, so the option is not buying anything there today.

*(Edited: I originally wrote that 24.04 was the current LTS. It is not, 26.04 is. The point about the older dash still standing in 24.04 is unchanged.)*

PR #476, opened 2026-09-18
eee: accept a port again, and stop the web UI discarding other ports

### LINKED ISSUE #459: eee command appears broken (as at end of day Sept 15th, 2026)
Running 2f005d2 which includes the new UI, I'm seeing an issue with EEE.
In short, changing the setting in UI does not work and the toggle flips back to "enabled" - this is despie it appending the below to the startup config and a reboot:

```
eee 1 off
eee 2 off
eee 3 off
eee 4 off
eee 5 off
eee 6 off
eee 7 off
eee 8 off
```

I have found that nearly all eee commands (the exception being "status") fail when put into the CLI, whether following the usage instructions or flipping values as a sanity check (to see if the instructions may be out of date):

```
> eee
eee [on|off|status] [port] [100m|1g|2g5]
> eee off 6
Speed word invalid, use: [100m|1g|2g5]
> eee 6 off 100|1g|2g5
eee [on|off|status] [port] [100m|1g|2g5]
> eee 6 off 1g
eee [on|off|status] [port] [100m|1g|2g5]
> eee off 6 1g
Speed word invalid, use: [100m|1g|2g5]
> eee 6 off
eee [on|off|status] [port] [100m|1g|2g5]
> eee off 6
Speed word invalid, use: [100m|1g|2g5]
> eee
eee [on|off|status] [port] [100m|1g|2g5]
> eee status 6
Speed word invalid, use: [100m|1g|2g5]
> eee status 6 100m
Speed word invalid, use: [100m|1g|2g5]
> eee status
Port: 1: Advertising:  2.5G  1G  100M   Link Partner:                 INACTIVE 
Port: 2: Advertising:  2.5G  1G  100M   Link Partner:                 INACTIVE 
Port: 3: Advertising:  2.5G  1G  100M   Link Partner:                 INACTIVE 
Port: 4: Advertising:  2.5G  1G  100M   Link Partner:                 INACTIVE 
Port: 5: Advertising:  2.5G  1G  100M   Link Partner:                 INACTIVE 
Port: 6: Advertising:  2.5G  1G  100M   Link Partner:        1G  100M ACTIVE   
Port: 7: Advertising:  2.5G  1G  100M   Link Partner:                 INACTIVE 
Port: 8: Advertising:  2.5G  1G  100M   Link Partner:                 INACTIVE 
Port: 9: SFP
> eee 6 off 100M
eee [on|off|status] [port] [100m|1g|2g5]
> eee off 6 100M
Speed word invalid, use: [100m|1g|2g5]
> eee off 6 100m
Speed word invalid, use: [100m|1g|2g5]
> eee 6 off
eee [on|off|status] [port] [100m|1g|2g5]
> eee
eee [on|off|status] [port] [100m|1g|2g5]
> eee status 6
Speed word invalid, use: [100m|1g|2g5]
> eee status 6 1g
Speed word invalid, use: [100m|1g|2g5]
```

bloqaudio:
Fixes #459.

Three things were wrong at once, so the command failed differently depending on where you drove it from.

**The parser never accepted a port.** `parse_eee()` sniffs word 2 for a trailing `g` or `m` to tell a speed from a port number, advancing `idx` past the digits to do so, and the port branch then hands that advanced `idx` to `cmd_parse_port_separator()`. That helper wants the first character of the number and finds the separator instead, so it returns 0 and the command is rejected. Every other caller passes `cmd_words_b[]`. The effect is that no `eee` command carrying a port worked in any argument order, including the per-port lines a saved configuration replays. @tofurky diagnosed this and posted the fix in the issue; it is carried here unchanged.

**The web UI sent the arguments backwards.** The EEE toggle posted `eee <port> <on|off>`, the reverse of the documented `eee [on|off|status] [port] [100m|1g|2g5]`, so the switch answered with its usage line and the checkbox sprang back. That was mine, from #429. @tofurky spotted this too.

**Fixing those two alone would have made it worse.** The stored-configuration merge in `html/app.js` builds its drop key as a line prefix. With the port trailing, a per-port line matches the bare `eee` entry and replaces every `eee` line in the configuration rather than the one for that port, so setting one port would quietly clear the rest. The command now has its own case beside the bandwidth one: a line with a port replaces only that port, a line without one clears them all, and the optional speed word the command has always taken is accepted in stored configurations.

**One more thing the bench showed.** `port_eee_enable()` and `port_eee_disable()` reported the port with `print_byte()`, the internal index in hex, so `eee off 5` answered `EEE off for 04` while `port_eee_status()` beside it counts from one. All three messages now use `print_phys_port()`.

This restores the documented grammar rather than taking a position in the wider discussion on the issue about unifying port settings under a single `port <n> ...` form. One data point for that discussion: because of the parser bug, every per-port `eee` line any UI has ever written is a no-op on replay, so there is no installed base of working per-port `eee` configuration to preserve.

Verified on hardware, an 8x2.5G plus SFP+ board on `PCB_SWTG018AS_V2_1_0`:

```
> eee off 5
EEE off for 5
> eee on 5
EEE on for 5 speed 2g5
> eee on 3 1g
EEE on for 3 speed 1g
> eee 5 off
Error: eee [on|off|status] [port] [100m|1g|2g5]
```

and in the web UI, ports 1 and 2 switched off, saved to flash, rebooted, both came back off, which is the case that fails without the merge fix.


vDorst:
Thanks for fixing.

Issue #296, opened 2026-07-24
在Port Configuration设置端口速率的时候没有生效


EthanCheung92:
我想指定端口速率，将其设置为强制100M全双工，但是在Port Configuration设置端口速率的时候似乎没有生效。

logicog:
Please use English in this forum.
Regarding port speeds, the port speed is negotiated by both parties. Setting the speed on the switch not necessarily makes this work with the connected device, if that device does not accept the advertised speed.

bloqaudio:
The selector on the Ports tab is never given a value, so it reads `auto` on every page load whatever the port is set to. If you set a port to forced 100M full, applied, and then reloaded, it would have shown `auto` again even though the setting had been written. #477 fixes that.

That may not be the whole of what you saw, since the link speed also depends on what the other end accepts, but it is worth retesting once #477 lands and checking the link badge rather than the selector.

PR #477, opened 2026-09-18
web: show the speed a port is set to, not always auto

bloqaudio:
The speed selector on the Ports tab is built with its options and never given a value, so it reads "auto" on every load whatever the port is actually set to. Force a port to 1G, apply, reload the page, and the selector says "auto" while the link badge beside it says 1G.

This is not new with the single-page UI. The multi-page UI it replaced parsed the same advertisement field into a `pAdvertised` array and never read it back, so the selector has never reflected the setting.

The advertisement bits `/status.json` already reports are enough to fix it. `phy_set_speed()` writes `PHY_ANEG_ADV`, `PHY_ANEG_MGBASE_CTRL` and `PHY_MMD31_GBCR`, and those are the three registers the JSON reads back into `adv`, so every forced speed leaves exactly one of the six bits standing while auto leaves the lot:

| setting | `adv` | selector |
| --- | --- | --- |
| auto, 2.5G port | `111111` | auto |
| auto, 1G port | `011111` | auto |
| 2g5 | `100000` | 2.5G |
| 1g | `010000` | 1G |
| 100m full | `001000` | 100M full |
| 100m half | `000100` | 100M half |
| 10m full | `000010` | 10M full |
| 10m half | `000001` | 10M half |

A value with no entry in that table, such as the bare `100m` setting that advertises both duplex modes, leaves the selector untouched rather than claim a setting the list cannot express. The selector is also left alone while it has focus, the same guard `loadMtu()` uses, so a refresh landing mid-edit does not overwrite what is being chosen. SFP ports use a separate rate list and are not touched.

Verified on hardware, an 8x2.5G plus SFP+ board on `PCB_SWTG018AS_V2_1_0`: a port set to 1G and applied still reads 1G in the selector after a page reload, and reads auto again after being set back. `machine_check` passes for all machines and the host suite passes 30/30.

This may also be the substance of #296, where a port was set to forced 100M full and appeared not to take effect. The selector reading `auto` again after a reload looks exactly like a setting that did not stick, whether or not it did.


feelfree69 (approved):
Thanks! Tested on KP_9000_6XH_X2, works perfectly.

vDorst:
Thanks

PR #479, opened 2026-09-19
[WIP] Refactor the Machine struct

vDorst:
This PR refactors the machine struct.

Many of the port-settings can be encoded in a single byte.
So we can see and check the port-usage.

Changes:
- `.phys_to_log_port` is removed, just walk over `.log_to_phys_port`, index of the array is the logical port number.
- `.n_sfp` is removed, it can be looked-up/counted in `log_to_phys_port`.
  In many cases we are not interested in the number of sfp ports but only is the SDSx is connected to a SFP-cage or external PHY.
- Added basic infrastructure to support external phy.
- link `.sfp_port` directly to the SDS-port/MAC. No need to not do that.
- Mark unused ports as `NOP`.
- `.sfp_port[]` is renamed and converted to a struct `sds_settings` with union for the settings. So we can store `SFP-cage`/`External PHY`/`Fixed-link`-settings on one location. Based on the flags in `.useage` we can see how to use it.

Todo:
- Stop using the `sfp_slot`-concept. I just want to link it to a portnumber/sds-port.
- `.min_port` and `.max_port` can be removed. Because every entry in `.log_to_phys_port` is label properly, we can just walk over it, without that much overhead.
- `n_10g` is I think the same as `.n_sfp`.
- unify port speed with just one command. `sfp` is just to show information, not for setting-up the port.


boot.c
```diff
@@ -328,14 +328,18 @@ void init_smi(void) __banked
...
 	// Workaround for SDCC BUG 4070: SFR_DATA_U32 = 0x00005555;
 	SFR_DATA_U16_UPPER = 0x0000;
 	SFR_DATA_U16 = 0x5555;
 	if (machine.n_10g == 2) {
 		// 0x00015555, only change the bytes that differs from the default.
 		SFR_DATA_16 = 0x01;
-	} else if (machine.n_sfp == 2)
+	} else if (two_sfp)
 		// 0x00005515
 		SFR_DATA_0 = 0x15;
```
DrDoof: The MAC_TYPE half can go per SDS here:

```suggestion
	if (sds0_is_sfp)
		// 0x00005515
		SFR_DATA_0 = 0x15;
	if (machine.n_10g == 2)
		// 0x00015555, only change the bytes that differs from the default.
		SFR_DATA_16 = 0x01;
```

The else-if goes because the two writes land in different bytes and only agree today since no machine sets both. The first test drops sds1, since that byte carries the MAC_SDS0 field alone: 0x55 with that field cleared is 0x15, which is what the line already writes.

I went through the trait combinations machine.c produces and the value written comes out the same for each. It differs only for a single cage on SDS0, which no machine has, and there the new form is the right one.

n_10g has to stay for now: every SDS_EPHY line is commented out, so usage reads SDS_UNUSED on the two machines with an external PHY.

DrDoof:
Two notes outside the MAC_TYPE hunk.

I left SMI_PORT_POLLING alone. Bit 8 stays set on a machine whose only cage is on SDS1 and I cannot explain that from the code, so rewriting it per SDS would be guesswork on a register that decides which ports get polled.

The branch also conflicts with main now, but only one hunk, in port_eee_enable().

vDorst:
I made my initial try to untangled it a bit.
I think I did ok.
I see that sds are configured on multiple places and not always consistent.
So I have to do more cleanup passes.

DrDoof:
I walked both registers machine by machine. MAC_TYPE matches the old code on 37 of 38. The odd one out is LIANGUO_HYWS_SGT0108S, the only board with no usage set for SDS0, which moves from 0x5555 to 0x5515. That smells like a missing entry rather than a board where SDS0 is really spare.

SMI_PORT_POLLING shifted further. Bit 8 used to be set on every non-8373 board that was not dual-SFP, and now it follows SDS1 having an external PHY. Six boards go from 0x0001f8 to 0x0000f8: K0501W_V2_0, KP_9000_6XH_X, LIANGUO_ZX_SWTGW215AS, SWTG024AS_A_2_0_1_5C_1SFP, SWTG024AS_V2_0 and ZX310S_4T2XH.

That may well be the answer to what I could not explain before, since a cage on SDS1 has no PHY to poll. Worth a look on hardware though.

vDorst:
> I walked both registers machine by machine. MAC_TYPE matches the old code on 37 of 38. The odd one out is LIANGUO_HYWS_SGT0108S, the only board with no usage set for SDS0, which moves from 0x5555 to 0x5515. That smells like a missing entry rather than a board where SDS0 is really spare.

This was a missing entry.

> SMI_PORT_POLLING shifted further. Bit 8 used to be set on every non-8373 board that was not dual-SFP, and now it follows SDS1 having an external PHY. Six boards go from 0x0001f8 to 0x0000f8: K0501W_V2_0, KP_9000_6XH_X, LIANGUO_ZX_SWTGW215AS, SWTG024AS_A_2_0_1_5C_1SFP, SWTG024AS_V2_0 and ZX310S_4T2XH.

I think I refactor it better.
We only need to let the hardware poll the PHY if a external PHY is connected via SMI-bus.
Also the internal phy can share that same bus.
Program also the PHY_ADDR for both MAC3&8.

PR #481, opened 2026-09-19
uip: verify TCP and UDP checksums using the RX descriptor flag

DrDoof:
Turns out the RX tag does tell us whether the checksum was good - it is bit 2
of byte 1, set when the L4 checksum is bad. That is the bit the TODO above the
TCP checksum code was asking about.

So TCP checksum checking can come back, and UDP may as well come along for the
ride, because UIP_CONF_UDP_CHECKSUMS is 0 and it was flying blind too. Reading
one flag beats making the 8051 add up the whole payload.

I poked at it for a while before trusting it. Broken ICMP checksums light the
flag up every single time and clean traffic never does. Four out of four
mangled TCP SYNs got dropped without a reply, while the web UI carried on as if
nothing had happened. A zero UDP checksum, which IPv4 says is perfectly fine,
does not get flagged at all - so DHCP has nothing to worry about and there is
no special case to write. And an NTP client pointed at a test server still
managed to set the clock, so good packets clearly still make it through.

156 bytes of code, no extra XDATA.


vDorst:
Thanks

PR #482, opened 2026-09-20
rtlplayground: take several packets per loop in handle_rx()

DrDoof:
The main loop only ever picks up one packet per pass through handle_rx(), and
since it sleeps until the next tick, that quietly puts a ceiling on the CPU port
of about one packet per tick - however little work there actually is to do. Aim
a burst of broadcast or IGMP traffic at it and everything queues up behind that,
STP and LACP timers included.

So this drains a few packets per pass instead of one. There is a budget on it,
deliberately: without one a sustained flood would just spin in there and nothing
else in the loop would get a look in.

I measured it with a one-way UDP stream, counting what the CPU actually got
through:

  before          159 packets/s
  RX_BUDGET 4     245 packets/s, loop pass ~16 ms
  RX_BUDGET 8     298 packets/s, loop pass ~26 ms

I settled on 4. Eight is a bit quicker, but most of what it buys is a longer
worst case for everything else in the loop, and throughput is already most of
the way there at four. STP stayed responsive either way.

71 bytes of code, no extra XDATA.

There is a companion change to idle() that comes at the same loop from the other
end, but the two are independent and this one stands on its own.


rtlplayground.c
```diff
@@ -973,19 +1103,22 @@ void tcpip_output(void)
...
 {
-	// Check the amount of data available on the NIC/ASIC side
-	reg_read_m(RTL837X_REG_NIC_RX_BUFF_DATA);
-	if (sfr_data[2] != 0 || sfr_data[3] != 0) {
+	__xdata uint8_t budget = RX_BUDGET;
+
+	while (budget--) {
+		// Check the amount of data available on the NIC/ASIC side
+		reg_read_m(RTL837X_REG_NIC_RX_BUFF_DATA);
+		if (!sfr_data[2] && !sfr_data[3])
```
vDorst: Use SFR_DATA directly.
```c
	reg_read(RTL837X_REG_NIC_RX_BUFF_DATA);
	if (!SFR_DATA_U16)
```

rtlplayground.c
```diff
@@ -973,19 +1103,22 @@ void tcpip_output(void)
...
+			break;
 		reg_read_m(RTL837X_REG_CPU_RX_CURR_PKT);
 		uint16_t ring_ptr = ((uint16_t)sfr_data[2]) << 8;
 		ring_ptr |= sfr_data[3];
```
vDorst: Also here, use SFR_DATA directly.
```c
	reg_read(RTL837X_REG_CPU_RX_CURR_PKT);
	uint16_t ring_ptr = SFR_DATA_U16;
```

vDorst (changes requested):

rtlplayground.c
DrDoof: done

rtlplayground.c
DrDoof: done

vDorst:
Thanks, all these changes, not only this one, seems to make it more responsive.
Based on the ping-times and Firefox loading.

PR #483, opened 2026-09-20
rtlplayground: do not sleep in idle() while a packet is waiting

DrDoof:
idle() nods off before it ever looks at the receive buffer, so a packet that
turns up just after the loop went to sleep sits there until the next tick before
anything notices it. Every TCP round-trip pays that, which is why the web UI
feels sluggish even though no part of the loop is anywhere near busy.

Look at the buffer first, and only sleep when there is nothing in it.

What that came to:

  web UI, median of 15 requests for the main page   88 ms -> 76 ms
  CPU port under a one-way UDP stream               159 -> 230 packets/s

The second number caught me off guard - I expected this to help latency and
leave throughput alone. What happens is that the loop starts turning over faster
than the tick once packets are queued, which is also the easiest way to see the
change doing its job: loops/ticks sits at 0.99 on an idle switch and only climbs
past 1.0 when something is actually waiting. So it is not a busy-wait, it just
stops napping when there is work.

12 bytes of code, no extra XDATA.

There is a companion change to handle_rx() that comes at the same loop from the
other end, but the two are independent and this one stands on its own.


rtlplayground.c
```diff
@@ -1157,7 +1408,9 @@ void handle_button(void)
 //
 void idle(void)
 {
-	PCON |= 1;
+	reg_read_m(RTL837X_REG_NIC_RX_BUFF_DATA);
+	if (!sfr_data[2] && !sfr_data[3])
```
vDorst: Please use SFR_DATA directly, cheap in codesize and cpu-cycles.
```c
	reg_read(RTL837X_REG_NIC_RX_BUFF_DATA);
	if (!SFR_DATA_U16)
```
The rest is good.

vDorst (changes requested):

rtlplayground.c
DrDoof: done

vDorst:
Thanks

PR #484, opened 2026-09-20
httpd: gate /upload on authentication before it changes state, and make the length limit count

bloqaudio:
Two fixes in the same handler, both in the firmware upload path.

## The authentication gate runs too late

In the multipart branch of `handle_post()`, `/config` tests `authenticated` inline and returns 401 before touching anything. `/upload` does not. It announces the upload on the console and sets `config_upload`, `uptr`, `verify_crc`, `max_upload` and `pre_acc`, and only meets a 401 further down at the dispatch, after all of that has been written.

The reason it matters is that the five values it writes are file-scope globals while the transfer state they steer, `tstate`, is per connection, and uIP is built here with ten connections. So a request that is about to be rejected can still retarget a transfer another client has in flight: a configuration upload continues as a firmware one, and its remaining parts go to the staging area rather than the configuration sector. The victim is an authenticated administrator mid-save, and the result is a configuration sector left erased.

The fix is the test the neighbouring branch already performs, moved ahead of the assignments.

## The upload length limit never fired

`max_upload` is assigned once when the upload starts and never reduced, so `max_upload - uip_len > 0` is unsigned arithmetic that only goes false when a segment arrives whose length matches the limit exactly. The write pointer climbs from the staging address with nothing to stop it, and `flash_write_bytes()` carries the address in three bytes and drops what does not fit. The comment on that line already said the check was unfinished.

It now subtracts each segment and compares before subtracting. The limit was only ever armed for firmware uploads, so configuration uploads keep the bound their own buffer imposes and are untouched.

## Verification

On an 8x2.5G plus SFP+ board, `PCB_SWTG018AS_V2_1_0`, with a serial console attached. Each case captured separately:

| case | HTTP | `Firmware upload started.` on console |
| --- | --- | --- |
| unauthenticated `POST /upload` | 401 Unauthorized | 0 |
| authenticated `POST /upload`, headers only | connection dropped | 1 |

The negative shows the gate now stops the request before any state changes. The positive shows the path is still reached, so the negative is not a test that simply misses.

Also on hardware, because the second commit sits in the flash write path:

- Good image uploaded through the web UI: checksum verified, switch reset and came back.
- Deliberately corrupted image of the same size: rejected on the CRC16 check, no reboot.
- Configuration round trip across six TCP segmentation shapes, byte-identical each time, including one run with an unauthenticated `POST /upload` fired from a second connection into the middle of it.


vDorst:
Thanks

PR #485, opened 2026-09-20
leds: index the second LED register from the set it belongs to

bloqaudio:
`leds_setup()` walks all four LED sets and writes the high-speed nibble of each into one of two registers, `LED3_0_SET1` for sets 0 and 1 and `LED3_0_SET3` for sets 2 and 3. The second branch reuses the index expression of the first without rebasing it:

```c
sfr_data[3 - (set << 1)]   /* set 2 -> -1, set 3 -> -3 */
sfr_data[2 - (set << 1)]   /* set 2 -> -2, set 3 -> -4 */
```

`leds_dump()` reads the same two registers back and already rebases with `(set - 2)` at both of its matching sites, so the intent is not in doubt.

Two things follow. Sets 2 and 3 never receive their nibble, because the read-modify-write pair writes back exactly what it read. And four bytes below `sfr_data` are overwritten on every boot of every machine, since the loop runs over all four sets whatever the board declares in `port_led_set`.

The second one is not theoretical. From the map of a build for `PCB_SWTG018AS_V2_1_0`:

```
D: 00000ABD  _sbuf        (SBUF_SIZE 16, so 0xABD..0xACC)
D: 00000ACD  _sfr_data
```

so the four negative indices land on `0xAC9..0xACC`, the last four bytes of the serial receive ring. It is survivable today only because it happens once, early, before anything is typed, and a different declaration order or `--xram-loc` would move it onto something else.

No board in `machine.c` currently uses LED set 2 or 3, so the LED half is latent and this is not a visible-behaviour fix for any existing device. It stops the out-of-bounds write and makes the upper two sets work if a board ever declares them.

Verified on an 8x2.5G plus SFP+ board, `PCB_SWTG018AS_V2_1_0`, which uses sets 0 and 1: after a cold power-cycle the LED behaviour is unchanged, copper green with activity, SFP green at up to 2.5G and blue at 10G. `machine_check` passes for all machines and the host suite passes 30/30.


vDorst:
I tested with moving the ledset to location 2,3
```patch
diff --git a/machine.c b/machine.c
index 3cd47b5..a1c3ffd 100644
--- a/machine.c
+++ b/machine.c
@@ -257,8 +257,9 @@ __code const struct machine machine = {
 
        .reset_pin = GPIO54_ACL_BIT2_EN,
        .high_leds = { .mux = LED_27 | LED_29, .enable = LED_28_SYS | LED_29 },
-       .port_led_set = { 0, 0, 0, 1, 0, 0, 0, 0, 1},
+       .port_led_set = { 2, 2, 2, 3, 2, 2, 2, 2, 3},
        .led_sets = {
+               { 0 }, { 0 },
                { /* RJ45: Green LED */
                        LEDS_1G | LEDS_100M | LEDS_10M | LEDS_LINK | LEDS_ACT | LEDS_10G,
                        0,
```
This works.


For everysome, please stop reporting things like this:
```
machine_check passes for all machines and the host suite passes 30/30.
```
The CI-build does that for you, and `machine_check` practically only check `machine.c`, not the rest of the code.

PR #486, opened 2026-09-20
syslog: do not read the connection before there is one

bloqaudio:
`udp_callbacks()` offers every UDP connection event to `syslog_callback()`, which decides whether the event is its own by comparing the local port against `state.syslog_conn->lport`. That pointer is null until syslog is started: `syslog_init()` clears it at boot and `syslog_stop()` clears it again. So with syslog off, which is the default, every poll dereferences null and reads whatever lives at the bottom of xdata.

If that word ever matches the port being offered, the send path runs on another application's connection: it overwrites `uip_appdata`, transmits to that peer, and clears `readptr` and `line_available`, losing the log line it was holding.

Test the pointer first. When it is set, behaviour is unchanged.

Verified on an 8x2.5G plus SFP+ board, `PCB_SWTG018AS_V2_1_0`: healthy with syslog off, and `syslog ip`, `syslog port`, `syslog on`, `syslog off` all behave and report correctly with the switch staying responsive throughout. 
Unrelated to this change, and reported here only because it turned up while testing it: I could not get the firmware to emit a syslog datagram at all. With a collector on the same subnet, on port 514 and on a high port, with the log ring demonstrably being fed, a capture on the receiving host saw no syslog UDP and no ARP from the switch while seeing other traffic fine. That is not caused by this patch, since the added test only short-circuits when the pointer is null and nothing is sent in that state either way. I have not chased it further yet and will open a separate issue once I know whether it reproduces on a stock build.

Filed as #490, reproduced on an unmodified build of `0cc09e4`.


vDorst:
Thanks

PR #488, opened 2026-09-20
rtlplayground: keep the STP and TCP timers on the system tick

DrDoof:
A follow-up to #482 and #483. stp_timers() ran every fourth pass of the main loop, which only measures time while a pass is one tick long. Under an IPv4 stream a pass is now about 16 ms, so STP time ran at a third of its rate; under a flood of cheap frames the loop no longer sleeps, so passes outrun the tick and STP time runs fast, shortening the forward delay right when a loop is forming. uip_periodic() had the same problem, so the retransmit timer sped up with the loop.

Both go by the tick counter from the timer interrupt now. STP catches up on ticks a long pass missed, eight steps at most; the TCP periodics run once per tick. 48 bytes of code, one byte of xdata.

A host test drives the gate with passes every 5, 16 and 0.6 ms: 500 STP steps in ten seconds each time, where counting passes gave 500, 156 and 4166.

## Measured on hardware

SWTGW218AS, hello time 2 s, BPDU captured on the neighbour with tcpdump while a one-way UDP stream was aimed at the CPU port. Load during the runs: 5291 packets in the window, `loops/ticks` down to 0.32, so a pass was three ticks long.

```
                       idle          under load
without this change    2.000 s       no BPDU at all in 25 s
with this change       1.99997 s     1.999 s   (2.001 2.008 1.988 2.000 2.000)
```

The stopped case is worse than the arithmetic suggests: at a third of the rate I expected hello every 6.25 s, but the timer had drifted far enough that nothing went out for the whole capture. A neighbour would age the link out and rebuild the topology.

Two traps if anyone reproduces this: capture on the physical port, not the bond, or the same BPDU shows up twice with a 0.00 s gap; and STP only sends through one member of a LAG, so the other port captures nothing at all.


DrDoof:
Measured on a SWTGW218AS, with the LACP timers gated the same way (they are in #299). A one-way UDP flood at the CPU port stretched the switch's LACPDU interval from 1.02 s to 2.4-2.9 s on the image before the change, one lost frame short of the partner's 3 s timeout; with the gate it stayed at 1.00-1.04 s through floods up to 84 kpps. The catch-up after a long pass shows up as an STP phase of up to 75 ms in diag; a cap of 4 steps instead of 8 would halve that, happy to change it.

PR #489, opened 2026-09-20
port: report link role and negotiated pause per port

DrDoof:
Two per-port MAC status registers that nothing in the firmware reads, even though each answers a question you currently cannot answer from the switch at all:

```
0x6408  mstr/slv  which end is master on a BASE-T link
0x63FC  tx pause  whether pause was negotiated
0x6400  rx pause
```

`port <n> show` now ends with two extra lines:

```
Link role: master
Pause negotiated: TX RX
```

My first attempt put these in the stats table as extra columns, which was a mistake: the row went from 77 to 112 characters and wrapped on any normal terminal. The per-port view has room for full words instead of cramped ones.

The read sits in `rtl837x_port.c` rather than `phy_show()`, because these are MAC registers and `phy_show()` is skipped for SFP ports. Master/slave keys off `machine.is_sfp[]`, so it stays off the fibre ports, where the notion does not apply at all.

Bit N of each register is chip port N, which is **not** how `RTL837X_REG_LINKS_STS` indexes its bytes a few lines above in the same header. I copied that pattern first and got plausible-looking nonsense.

An earlier version of this branch also printed a medium line from `0x63EC`. feelfree69 showed that bit follows the link rather than the port: an RJ45 module in a cage reads copper while the link is down and fibre once it is up. That part is gone.


feelfree69:
Not sure how useful the medium info is.

Tested with a KP_9000_6XH_X2 device:

After power-up, a plugged DAC cable without link partner reports copper. After establishing a link it reports fibre. After unplugging the remote-side it stays at fibre.

Same behaviour with a RJ45 SFP module. Copper during LOS, fibre during link.

I would say no info about the medium is better than the wrong info.


**Another minor finding, unrelated to this PR:**

`port <n> show`  prints different strings for both AN advertisements for 2.5G rates and above:

```
Link speed: 2500M full duplex
AN enabled, advertising: 10Base-Half 10Base-Full 100Base-Half 100Base-Full 1000Base-Full 2500BaseN-Full
Link Partner advertises: 10Base-Half 10Base-Full 100Base-Half 100Base-Full 1000Base-Full 2500Base-Full
```

feelfree69:
> On my SWTGW218AS the output matches the box: pause reports `TX RX` on the two ports carrying an aggregate, `fibre` on the SFP port, and `inactive` elsewhere.

Not sure if I understood this right.

On my box. Flow Control reports "TX RX" on the copper ports (which I assume is wrong, I have not enabled any bandwidth limit or flow control), and the SFP port with the 10G DAC cable reports `Flow control: inactive`.

DrDoof:
That is a clean catch, and the DAC test is one I had no way of running here.

What you describe fits a register that tracks the link rather than the port. The vendor name for 0x63EC is MAC_LINK_MEDIA_STS, and an RJ45 module reads copper during LOS and fibre once the link is up, so the bit follows the link and not what is plugged in. Mine only looked convincing because the SFP was the one port with a link.

Worse, my else branch printed copper whenever the bit was clear, so a port with no link got one invented for it. The Medium line is gone now, and master/slave keys off machine.is_sfp[] instead.

On flow control, the register reports what was negotiated rather than anything configured, so TX RX on copper ports is autonegotiation doing its job. The label was the misleading part; it now reads Pause negotiated.

feelfree69:
> On flow control, the register reports what was negotiated rather than anything configured, so TX RX on copper ports is autonegotiation doing its job.

Ah, then it's the Autoneg feature which enables Pause on TX/RX. Setting a fixed speed disables it:

```
> port 2 auto
Logical Port: 05
 AUTO
Setting port 2 to speed auto

>
<new link: 2000050000, was 2000550000>

<new link: 2000550000, was 2000050000>
> port 2 show
Logical Port: 05
Name:
Link speed: 2500M full duplex
AN enabled, advertising: 10Base-Half 10Base-Full 100Base-Half 100Base-Full 1000Base-Full 2500BaseN-Full
Link Partner advertises: 10Base-Half 10Base-Full 100Base-Half 100Base-Full 1000Base-Full 2500Base-Full
Link role: slave
Pause negotiated: TX RX

> port 2 2g5
Logical Port: 05
 2.5G
Setting port 2 to speed 2G5

>
<new link: 2000050000, was 2000550000>

<new link: 2000550000, was 2000050000>
> port 2 show
Logical Port: 05
Name:
Link speed: 2500M full duplex
AN enabled, advertising: 2500BaseN-Full
Link Partner advertises: 10Base-Half 10Base-Full 100Base-Half 100Base-Full 1000Base-Full 2500Base-Full
Link role: slave
Pause negotiated: none

>
```

There's no explicit setting for the Pause config yet, right?

DrDoof:
Right, nothing sets it per port today. But your 2g5 run turned up something better than a missing knob: the advertisement written for forced speeds from 1G up is 0x1001, while auto and the forced 10M and 100M modes all carry bit 10, the PAUSE bit. So forcing the speed takes pause off the link rather than just reporting it differently.

I pushed a one-value change to phy/advertise-pause-above-1g, 0x1001 to 0x1401. I cannot test it here, since it needs a link partner that advertises pause. If you have a spare moment, does port 2 2g5 still say none with that branch?

The MAC side does have TX_PAUSE_EN and RX_PAUSE_EN per port in MAC_FORCE_MODE_CTRL0, so a real setting is possible later, but this looks like the actual bug.

DrDoof:
Tested it here in the end, without flashing anything. phyget and physet read and write the advertisement directly, and a port with no link is a safe place to try it.

Forcing 1g takes 7.0x10 from 0x15e1 to 0x1001, auto puts it back, and writing 0x1401 by hand sticks, so the PHY takes the value.

What settles the bit is a second reading: two ports advertising the same 0x15e1 from our side, one partner at 0xd041 without bit 10 and the other at 0xdde1 with it. Pause is active only on the second. So bit 10 is the pause capability, and forcing a speed above 100M quietly drops it.

Opened #509 with the one-value change. A reading from your board would still be welcome if you get a moment.

feelfree69:
> e. If you have a spare moment, does port 2 2g5 still say none with that branch?

No, Pause TX RX stays enabled in that branch.

feelfree69 (approved):

DrDoof:
That closes it then. Your board had the piece I could not reach here, a partner that actually advertises pause, so the chain now runs all the way from the advertisement to the negotiated result.

Adding it to #509. A second board with a real link partner is worth rather more than my port with nothing plugged into it.

On the explicit setting you asked about, in case you want to poke at it: it lives in MAC_FORCE_MODE_CTRL0, 0x6344 + port * 4, with TX_PAUSE_EN at bit 7, RX_PAUSE_EN at bit 8 and SMI_FORCE_FC_EN at bit 9 to make the forcing count. The firmware already knows that address as RTL837X_REG_MAC_FORCE_MODE and writes it for the CPU port in boot.c. regget and regset will let you try it on a live port before anyone writes a command for it.

PR #492, opened 2026-09-20
CI: Turn on error on warning also for sdcc-compiler.

vDorst:
Turn on error on warning also for sdcc-compiler.
With a extra flag `CI=1` in the makefile, so that local build just compile normal.
Enable `CI=1` stops the compiler on a warning.
Now also build (4x2/RTL8372) and (8x1/RTL8373) project.


Makefile
```diff
@@ -20,6 +20,10 @@ else
 	CC_FLAGS += -DMACHINE_$(MACHINE)
 endif
 
+ifdef CI
+	CC_FLAGS += --Werror
```
tofurky, line 24: I think it should be `-Werror`, single hyphen:
```
$ sdcc --help
SDCC : mcs51/z80/z180/r2k/r2ka/r3ka/sm83/tlcs90/ez80_z80/z80n/r800/ds390/TININative/ds400/hc08/s08/stm8/pdk13/pdk14/pdk15/mos6502/mos65c02/f8 TD- 4.5.0 #15242 (Linux)
published under GNU General Public License (GPL)
Usage : sdcc [options] filename
Options :-

General options:
      --help                Display this help
  -v  --version             Display sdcc's version
      --verbose             Trace calls to the preprocessor, assembler, and linker
  -V                        Execute verbosely. Show sub commands as they are run
  -d                        Output list of macro definitions in effect. Use with -E
  -D                        Define macro as in -Dmacro
  -I                        Add to the include (*.h) path, as in -Ipath
  -A                        
  -U                        Undefine macro as in -Umacro
  -M                        Preprocessor option
  -W                        Pass through options to the pre-processor (p), assembler (a) or linker (l)
...
```

vDorst: According to the manual https://sdcc.sourceforge.net/doc/sdccman.pdf page 35
Screenshot_20260920_223105

It is `--Werror`.

Also compiling with `-Werror` generates a warning `at 1: warning 117: unknown compiler option '-Werror' ignored`

tofurky: Oh, sorry, should've actually tested it before commenting.
```
$ sdcc --help|grep error
      --use-stdout          send errors to stdout instead of stderr
      --Werror              Treat the warnings as errors
```

Makefile
```diff
@@ -20,6 +20,10 @@ else
...
+endif
```
DrDoof: `ifdef` is true for any value, including `CI=0`, and CI services set `CI=true` in the environment of every step, so the flag also turns on in the steps that do not pass `CI=1`.

```suggestion
ifeq ($(CI),1)
	CC_FLAGS += --Werror
endif
```

DrDoof (approved):
Makes sense to me. Two notes besides the one on the Makefile.

The branch is 13 commits behind main, so the green run covers the tree as it was before #494 and #496 landed. A rebase would check the flag against current code, and it merges cleanly.

One to keep in mind rather than change: the container takes whatever sdcc debian:trixie ships, 4.5.0 today. When that moves to 4.6 the flag picks up the new warnings too, and builds fail for reasons unrelated to the change under test.

Makefile
vDorst: I just wanted a flag, it is there or not. But that seems not to exists.
I add your suggestion. Maybe other people think it is a boolean.

vDorst:
@DrDoof:  Rebased and added your suggestion

> One to keep in mind rather than change: the container takes whatever sdcc debian:trixie ships, 4.5.0 today. When that moves to 4.6 the flag picks up the new warnings too, and builds fail for reasons unrelated to the change under test.

Ideally I think we want Github to use our contailerfile, in which we control the version.
Like directly download it from SF.net.
I saw that github can use your own container-images and even cache them. https://github.com/marketplace/actions/cache-container-images-action

DrDoof:
On the container: there is already a Dockerfile in the tree, added for the dev environment, and nothing references it. It installs sdcc from apt on debian:13-slim, which is the same release the workflow pulls, so as it stands it would not pin the version either. It would move the same apt line somewhere else.

If the workflow built that image instead, the version becomes one line to change, whether that ends up being an apt pin or a download from SF.net. It would also keep the workflow and the local dev image from drifting apart, which seems worth more than the caching.

PR #493, opened 2026-09-20
rtlplayground: read the link registers when the switch reports a change

DrDoof:
The link-change interrupt has been set up at boot for a long time (mode 0x5f84, mask 0x5f34 for the ten ports), but `EX0` stayed commented out: the handler did nothing and never cleared the event in the switch, so `idle()` read the two link registers on every pass instead. On a quiet switch that is 400 register reads a second to find out that nothing changed.

This finishes it.

- `isr_ext0()` only latches a flag and masks itself. The register access macros build a write in the shared SFR data bytes, so a handler that touched the switch in the middle of a write from the main loop would corrupt it.
- `idle()` clears the event register (write-one-to-clear, `0x3ff`), re-enables `EX0` and then reads the link state, in that order, so a change that lands during the read raises a new edge. The link code itself moved unchanged into `check_links()`.
- The flag is also set once a second from the seconds block, so a board whose INT0 line is not wired keeps detecting link changes, a second late at worst.
- Boot drops the events pending from bringing the links up before the mask is set.

Measured on an RTL8373 board (SWTGW218AS), serial console timestamped on the host:

```
23:16:29.573 Setting port 3 to disabled
23:16:29.591 <new link: 2400105000, was 2400105500>
23:16:44.700 Setting port 3 to disabled
23:16:44.720 <new link: 2400105000, was 2400105500>
23:16:59.798 Setting port 3 to disabled
23:16:59.818 <new link: 2400105000, was 2400105500>
```

18 to 20 ms from the command to the report on each of three link-downs. The one-second fallback alone would spread that between 0 and 1000 ms, so the interrupt is what delivers it. `0x5f88` read as zero after every event and no interrupt flood was seen; before the change the same register held `0x3fd`, nine ports that had reported a change nobody read. STP and LACP reacted to the events as before.

SWTGW218AS: code +57 B, xdata unchanged, stack 151 -> 150 B (the flag is a bit variable). The handler compiles to `clr EX0 / setb link_irq / reti`.

---

### Second commit: one less read of the receive buffer

While looking at what else the loop reads on every pass, the receive buffer turned out to be checked twice. `idle()` reads `NIC_RX_BUFF_DATA` to decide whether to sleep, and `handle_rx()` reads the same register again a moment later to decide whether there is anything to take. Only the link check and `handle_sfp()` run between the two, and neither can make a queued packet disappear, so the first iteration of the receive loop can use what `idle()` already saw. Later iterations read the register as before.

**This part is not measured on hardware.** The interrupt change above is; this one is not. One register read per pass while traffic flows is too small to show up without a throughput run, and I would rather say so than imply a number I do not have. The argument for it is structural, not empirical: a packet already sitting in the buffer cannot go away on its own.

It costs 6 bytes in BANK1 and 6 in BANK2, BANK3 unchanged. Happy to drop this commit if you would rather keep the PR to the interrupt work alone.


vDorst:
This is already merged via #499?

DrDoof:
This one went in with #498, which carried the same change on top of #488: it is in main as f87eb70. Closing it so it does not show up as a conflict.

PR #494, opened 2026-09-20
Fix warnings seen on SDCC 4.6.0 due to missing const qualifiers

tofurky:
Building with SDCC 4.6.0 emits tons of warnings e.g.:

	cmd_parser.c:31: warning 356: object in read-only code space should be const
	cmd_parser.c:36: warning 357: pointer to object in read-only code space should be pointer to const

These of course become fatal when building with --Werror.

Comparing resultant binaries, it seems that a single string is shifted (deduplication?) but the numbers emitted at the end of the build are otherwise identical.


tofurky:
I also considered `#define __ccode __code const`, but it's just saving some bytes while being a bit obscure.

PR #495, opened 2026-09-20
Properly handle NULs when normalizing SFP fields

tofurky:
Some generic 1.25gbps DAC cables were seen to have a NUL *after* the space padding in the model field. This caused the normalization to miss the space padding in the string. The vendor field did not have the issue.

Note that this seems to be against spec per SFF-8472, which mentions only space padding or *all* NUL, not a combination of both. In the case of all NUL, we'll waste a few cycles with no-op writes.


vDorst:
Thanks

PR #496, opened 2026-09-21
sfp: do not block the main loop while a module wakes up

DrDoof:
Plugging in an SFP module froze the whole switch for half a second. Not the SFP handling - everything: STP stopped sending BPDUs, LACP stopped, packets waited. It turned out to be a single line.

`handle_sfp()` runs on every pass of the main loop, but it only touches I2C when a module actually appears. That path starts with `delay(100)`, meant to give the module time to wake up before its EEPROM is read. The problem is what `delay()` does:

```c
void delay(uint16_t t)
{
	sleep_ticks = t;
	while (sleep_ticks > 0)
		PCON |= 1;
}
```

It spins on `PCON` until the tick counter runs down. At `SYS_TICK_HZ 200` that is 100 ticks, so half a second in which the main loop does not advance at all. A diagnostic build put the `sfp` phase at up to **107 ticks** - by a wide margin the longest stall of any phase, several times anything in the receive path.

The fix keeps the wait but stops blocking for it. When a module is detected, the current tick is recorded and the loop carries on; the I2C read happens on a later pass, once the wake-up period has elapsed.

Behaviour around it is unchanged. Removing a module clears a pending read, so a module that is yanked out during the wake-up window is never read. A failed read still restores the detect bit, which means the existing retry-on-the-next-poll path works exactly as before - it simply starts a fresh wake-up period.

The tick comparison is spelled out instead of using a helper, so this does not depend on anything outside the current tree.

### Cost

Measured on SWTGW218AS, full rebuild of both trees (`make clean` alone is not enough here - stale `.rel` files survive it and the linker happily reuses them):

```
          before    after    delta
BANK1      47934    47940       +6
BANK2      43328    43334       +6
BANK3      11166    11306     +140
```

Plus 3 bytes of xdata for the two new variables.

### What is not tested

This is built but **not yet verified on hardware** - I have no spare switch to hot-plug modules on, and the one I have carries a live cluster behind it. The logic is straightforward and the failure paths are unchanged, but someone with a module to spare pulling and reinserting it would be a useful confirmation.


vDorst:
I feels like it is time for a proper timer framework/task scheduler.
It is fine that it is just a co-operative multitasking.

The SFP-handler needs to be a state-machine, with as input a event why it is called.
This could solve more issues we can't handle at the moment like:
- Add poll-feature for device had don't have a detect-pin or SFP-module don't have a detect-pin.
- SFP modules that need a longer boot-time before it can do other-things.
- Support external phy's inside the SFP-module that also needs a poll to fetch the status.
  #479 helps the code to have basic-support/foundation for it.

Also other code like LLDP and STP have there own timer, so a timer-framework is beneficial for them too.

Issues which are related to timer:
- #488
- #474
- #456
- #452
- #455

DrDoof:
A timer framework is something I would love to see here as well. The timing in
this tree really is scattered: DHCP counts down its own ticks, STP divides with
stp_clock, httpd compares against the global ticks, and delay() spins on
sleep_ticks. Four ways of saying "later" in one firmware. An SFP handler driven
by events, instead of a function that happens to be called on every pass, would
also be a far better fit for what the hardware is actually doing.

Before replying I went and looked at how much blocking is really left once this
patch is in, and it came out smaller than I expected:

- rtl837x_phy.c has ten delay(20) calls, but they all hang off phy_set_speed(),
  which only ever runs from the CLI.
- The six in boot.c live in early_boot_handle_button(), rtl8224_enable(),
  rtl8373_revision() and init_smi(), so start-up only. sds_config(), which the
  SFP path calls right after reading a module, has none at all.
- That leaves handle_sfp() at half a second, which this patch takes out, and
  /reset at five seconds, which is followed by a chip reset anyway.

So this is the last blocking wait the main loop runs into during normal
operation, which is why I would rather not park twelve lines until the bigger
design is ready. It does not box us in either: it swaps "block until the module
wakes up" for "note when it appeared and come back later", which is the shape a
state machine wants anyway, just written out by hand for the single case we
have today.

What it does not do is any of the rest of your list. Polling cages with no
detect pin, modules that need longer than one fixed wake-up, external PHYs with
their own status to fetch, none of those fold into an if-statement, and piling
more special cases into handle_sfp() would only make the eventual rewrite
harder.

I would enjoy building that with you. If #479 is meant to be the ground it
stands on, landing that first and putting the timer on top sounds saner than
running two refactors next to each other. Say which end you would like me to
take and I will start there.

sfp.c
```diff
@@ -199,6 +202,11 @@ void handle_sfp(void) __banked
 			if (sfp_pins_last & (0x1 << (sfp << 2))) {
 				sfp_pins_last &= ~(0x01 << (sfp << 2));
 				print_string("\n<MODULE INSERTED>  Slot: "); write_char('1' + sfp);
+				sfp_wake_at[sfp] = ticks;
+				sfp_wake_pending |= 0x1 << sfp;
```
feelfree69: Personally, I don't like the bit shifting here, when at the same time the other sfp-related data in xdata are all arrays.
Do we need it at all? Or can we simply decrease sfp_wake_at every tick and fire the action once it reaches zero?

feelfree69 (commented):
small comment

sfp.c
DrDoof: Changed to an array, thanks. It does sit better next to sfp_options and the rest of them.

On the countdown: I tried it. It cannot live in handle_sfp(), which runs once per loop pass rather than once per tick, so it would have to go into the timer ISR next to sleep_ticks. That builds and works, but it costs two compares in the interrupt 200 times a second on every machine, including those without a cage, and moves 31 bytes into CSEG. The timestamp costs nothing while nothing is plugged in.

So I kept the timestamp and dropped only the bit mask. BANK3 goes down by 49 bytes, CSEG unchanged.

I can switch to the counter if you prefer it, the change is small either way.

feelfree69 (approved):
Thanks! (Not tested in HW, just read the code).

vDorst:
> So this is the last blocking wait the main loop runs into during normal operation, which is why I would rather not park twelve lines until the bigger design is ready. It does not box us in either: it swaps "block until the module wakes up" for "note when it appeared and come back later", which is the shape a state machine wants anyway, just written out by hand for the single case we have today.
> What it does not do is any of the rest of your list. Polling cages with no detect pin, modules that need longer than one fixed wake-up, external PHYs with their own status to fetch, none of those fold into an if-statement, and piling more special cases into handle_sfp() would only make the eventual rewrite harder.

Yes lets merge it and make a new PR to do the rest.

> I would enjoy building that with you. If #479 is meant to be the ground it stands on, landing that first and putting the timer on top sounds saner than running two refactors next to each other. Say which end you would like me to take and I will start there.

#496 is a refactor so the software understand that we can have external devices connected to the SDSx ports which needs a different setup. And make it easier to define it in machine.c and the way to handle it in the rest of the code, like handle_sfp().
Now I am at a point to unraffel how the code is setting-up when it is a sfp port or a external phy in `init_smi(void)`. I am struggling a bit to split this per SDS and what do we need as config for the phy. Like storing in machine.c what the max-speed is.

Maybe we need move this converastion #479.


I merge this thanks, lets add a new issue about the timer-framework design.

PR #498, opened 2026-09-21
rtlplayground: run the timed work once per tick

DrDoof:
Draft while #488 and #493 are open: this branch carries their commits (the first four), so only the last commit is this change. Rebased on today's main, which already has the SFP wake-up wait off the loop (#496), so this is now only the tick block.

idle() ran everything on every pass: the seconds block, the SFP pin test, the TCP periodic timer behind a gate of its own, the STP timers behind theirs. The timed work now lives in handle_tick(), which idle() calls once per system tick from a single gate, and handle_tx() loses its private one. The link, receive and console paths stay where they were.

handle_sfp() ran on every pass as well, two GPIO reads per pass to learn that nothing changed; it now runs every ten ticks, so a module is still seen within 50 ms.

```
SWTGW218AS      before    after
code            112741   112786   (+45 B)
xdata            14573    14574   (+1 B)
stack              150      150
```

Host harness passes, no cross-bank calls, machine.c compiles for all 38 machines.


vDorst:
Thanks

PR #499, opened 2026-09-21
rtlplayground: receive on the NIC interrupt instead of polling the buffer

DrDoof:
Draft, stacked on #498 (which in turn carries #488 and #493): only the last commit is this change. Rebased on today's main; I will rebase again as those go in.

The receive interrupt was wired up but never used: EX1 was enabled, the handler re-enabled itself and did nothing, and the mask bit for it in the NIC stayed clear. So idle() read the receive buffer register on every pass to find out whether a packet had arrived, and slept until the next tick when none had. A packet that arrived during that sleep waited for the tick: 2.2 ms on average, 5 ms at worst, for every frame the CPU handles.

Now the handler latches rx_irq and masks itself, and idle() sleeps unless a receive, a link change, a console command or the tick is pending. On rx_irq the loop drains the buffer, clears the status bit and re-enables EX1, in that order. The NIC holds the interrupt line as long as its status bit is set, and the bit does not clear while the buffer still holds a frame, so a frame that lands between the last read and EX1 = 1 raises the interrupt again rather than getting lost. The other order, clearing before draining, leaves the line asserted and gives an interrupt on every pass. The mask is written after nic_setup(), which resets it.

Measured on an RTL8373 board, the same counters compiled into both images and clocked by the tick counter, traffic from a host on an LACP trunk. The board runs the LACP branch (#299), so the measurement was taken there with this same change on top; the loop code is identical on this branch:

```
                           polling    interrupt
idle, register ops/s           788           79
25 pps, RTT avg/max      2.19/4.96    0.44/0.75 ms
500 pps, RTT avg/max     2.78/5.06    0.39/5.40 ms
500 pps, loss                0.08%        0.08%
```

Only tested on RTL8373. On a board whose INT1 is not driven by the NIC this would stop receiving, and the once-a-second fallback is no safety net at all: it drains four frames a second. If someone with an RTL8372 board can run a ping through the CPU before this goes in, that would settle it.

```
SWTGW218AS      before    after
code            112786   112839   (+53 B)
xdata            14574    14574
stack              150      150
```

The handler compiles to clr EX1 / setb rx_irq / reti.


rtlplayground.c
```diff
@@ -1206,57 +1246,41 @@ void idle(void)
...
-			stp_clock--;
 		}
 	}
-	// Check whether a command is waiting in the cmd_buffer and execute
+}
+
+
+void idle(void)
+{
+	if (!rx_irq && !link_irq && !cmd_available && !TICKS_DUE(idle_tick_last, 1))
```
vDorst: Can we put all of these bits in only `flag`-byte which is also bit-addressable and keep it in SRAM because it is access a lot and want to atomic operation on it.

I think timer_tick also need a flag, so we don't need to calculate the change every time.

`link_irq` is this also shared with the other IRQ like GPIO?

DrDoof: Nice catch on both counts.

link_irq, rx_irq and tx_buf_empty are already __bit, so they sit in the bit-addressable area at 0x20 and set and clear atomically. cmd_available is the one left behind, still a byte in xdata. Making it a bit saves 11 bytes of CSEG and 8 of BANK2.

Your tick flag was the bigger win. Setting it in the timer ISR and reading it in idle(), rather than working out TICKS_DUE twice a loop, saves 36 more and lets idle_tick_last go entirely. Folding both in.

STP has to stay as it is. stp_tick_last moves by += STP_TICK_STEP with catch-up, so it needs the count, not just an edge.

On link_irq: it is only the port link change, IMR 0x5f34 masked to the ten port bits. The GPIOs here are SFP detect and RX-LOS, which we poll. IRQ2 and IRQ3 only clear EXIF.

vDorst: My idea is this.

Define a byte in SRAM or SFR that also can be bit-addressable.
Define the bits you want up-to 8 flags.
Now you just have to check 1 byte to see if a flag is active.

I hope SDCC has a better/automatic way todo this.
This code is not working yet, I am not sure about SRAM and BIT-address location.
But I hope this gives you a picture what I had in mind.

```patch
index 10700a4..4a7c970 100644
--- a/rtlplayground.c
+++ b/rtlplayground.c
@@ -137,9 +137,20 @@ __code const uint16_t bit_mask[16] = {
 
 __xdata uint8_t linkbits_last[4];
 __xdata uint8_t linkbits_last_p89;
-volatile __bit link_irq;
-volatile __bit rx_irq;
-volatile __bit tick_pending;
+
+// Place flags in the bit-addressable SRAM location or unused SFR-register which is also bit-addressable.
+__data __at (0x2F) uint8_t flags;
+// flags bits, so we can do atomic operation on it and use `jbc`-instruction.
+// Flags must active high, otherwise `jbc`-instruction can't be used.
+__bit __at (0x78) link_irq;
+__bit __at (0x79) rx_irq;
+__bit __at (0x7A) tick_pending;
+__bit __at (0x7B) reserved_1;
+__bit __at (0x7C) reserved_2;
+__bit __at (0x7D) reserved_3;
+__bit __at (0x7E) reserved_4;
+__bit __at (0x7F) reserved_5;
+
 // Last known state of the SFP detection/Loss of Signal pins
 __xdata bool button_last;
 __xdata uint8_t button_sec_counter_last;
@@ -1264,7 +1275,7 @@ static void handle_tick(void)
 
 void idle(void)
 {
-       if (!rx_irq && !link_irq && !cmd_available && !tick_pending)
+       if (!flags)
                PCON |= 1;
        if (tick_pending) {
                tick_pending = 0;
```

vDorst: I think `tick_pending` part it only should handle the time-administration. 
This should set other flags like 1-sec is leap. So other sub-systems can react to it.

DrDoof: Both work out nicely. A byte at 0x20 with named bits compiles cleanly and the check drops to a single mov. It is in now, with rx, link, tick and cmd sharing the byte, and the bit area went from three bytes down to two.

I left tx_buf_empty out, as it means the opposite of the others, 1 when there is nothing to do, so folding it in would stop the sleep test firing at all.

On the tick doing only time administration: handle_tick() already spots the second and calls handle_button() right there, so a flag has no reader yet. Did you have later subsystems in mind, or saving the button handler a second read of the seconds register?

vDorst:
Ping time is now super low. typ 300 uS via fiber.

> On the tick doing only time administration: handle_tick() already spots the second and calls handle_button() right there, so a flag has no reader yet. Did you have later subsystems in mind, or saving the button handler a second read of the seconds register?

No, I never designed a system like this.
I may going to look at the Embassy Rust Async framework. To see how they handle such things.
Or maybe RTOS.

Thanks for the improvement.

PR #502, opened 2026-09-21
Add support for Linksys LN2308

jlunz:
This series adds support for the Linksys LN2308.

Initially I just wanted to change the brightness of the power LED, then I spotted the UART pinout and let Claude run for a while with a serial adapter, a way to power on/off the device, and flashrom for the SOIC-8 chip.
It works, and I've tested everything I can and need so far, let me know if AI-assisted contributions are ok or not.

One thing to note: I currently see a 9th port, but the device only has 8 external ports, no SFP port like many other devices here. I can add an additional change, but wanted your feedback first.

Thank you for this project!


machine.c
```diff
@@ -1119,6 +1100,62 @@ __code const struct machine machine = {
 		},
 	};
 
+#elif defined MACHINE_LINKSYS_LN2308
+/*
+ * Linksys LN2308: 8 x 2.5GBit RJ45, no SFP cage. RTL8373N + RTL8224N (chip id 0x83737000),
```
feelfree69: I think we should not add lengthy comments here, but move them to the device-markdown instead.

feelfree69:
> One thing to note: I currently see a 9th port, but the device only has 8 external ports, no SFP port like many other devices here. I can add an additional change, but wanted your feedback first.

I'm not sure how to handle this. I tend to put changes for "SFP- and 10G-less"-devices into a seperate PR.

vDorst:
> > One thing to note: I currently see a 9th port, but the device only has 8 external ports, no SFP port like many other devices here. I can add an additional change, but wanted your feedback first.
> 
> I'm not sure how to handle this. I tend to put changes for "SFP- and 10G-less"-devices into a seperate PR.

For now, max_port is just 7 (which is last mac of the 8 ports)

```c
.max_port = 7,
.log_to_phys_port = {1, 2, 3, 4, 5, 6, 7, 8, 0},
.phys_to_log_port = {0, 1, 2, 3, 4, 5, 6, 7, 0},
```

In the #479, this is better handled at least on the init level, I am not yet at the part how to present it the right way.

machine.c
```diff
@@ -1119,6 +1100,62 @@ __code const struct machine machine = {
...
+ * as a plain port because the CPU port is assumed to be max_port + 1. It shows up as port 9.
+ *
+ * Case label N is chip port N-1 with no swaps, verified on hardware 2026-09-19 by walking a
+ * cable along labels 1-8 and watching the link status, so log_to_phys_port is the identity.
+ */
+__code const struct machine machine = {
+	.machine_name = "Linksys LN2308",
+	.isRTL8373 = 1,
+	.min_port = 0,
+	.max_port = 8,
```
feelfree69, line 1128: should be `max_port = 7` (thanks @vDorst)

machine.c
jlunz: I'll check and move it into doc/devices/LN2308.md

machine.c
jlunz: With that change there might be a problem with vlan_setup:

max_port = 7 stops the loop at 8, but the CPU_PORT is at 9:
https://github.com/logicog/RTLPlayground/blob/fa93340c2434e9b70f5519d94b41d07fa31978b9/rtl837x_port.c#L279-L280

feelfree69: Good catch! 

@vDorst is refactoring the machine definitions in #479. Maybe we can take this new usecase also into account there.

machine.h
```diff
@@ -71,6 +72,11 @@ struct sfp_port
 	uint8_t i2c;
 };
 
+#ifdef MACHINE_LINKSYS_LN2308
+// Dual-IO flash reads leave the LN2308 rebooting/crashing after the first banked call; use single IO
+#define FLASH_SIO_ONLY 1
```
feelfree69, line 77: Using a macro and setting this at compile-time feels wrong, as everything else is set at runtime using the machine config.

@vDorst what do you think?

PR #503, opened 2026-09-21
uip: rate-limit the dropped-fragment log

### LINKED ISSUE #491: uip logs every dropped fragment, which floods the console
`uip.c` logs `ip: fragment dropped.` for every fragment it discards. uIP has no reassembly, so on a network where anything sends fragmented traffic this is continuous rather than occasional. On a switch here it runs at roughly thirteen lines a second, 156 of 156 lines in a twelve second capture, which buries everything else on the console.

In this case the source is `usteer`, the roaming coordinator that OpenWrt access points use, which broadcasts oversized state to UDP 16720. Any fragmented broadcast traffic would do the same, and the switch is behaving correctly by dropping it. The problem is only that a correct, expected drop is logged every time.

The message is worth keeping, since a sudden burst of it says something real about the network. What it is not worth is one line per fragment forever. A rate limit on that single call, something like one line per five second window with the count, keeps the signal and loses the flood.

Note it does not reach syslog, because `UIP_LOG` is `print_string_newline_no_syslog`, so this is console only.

Happy to send that as a pull request if the approach seems right.

bloqaudio:
Closes #491.

`uip` has no reassembly, so every fragment it receives is discarded and logged. Three access points here broadcast `usteer` state as fragmented traffic, which turns an occasional message into a continuous one and buries everything else on the console. The switch is right to drop the fragments; the only problem is logging each one.

This counts the drops and reports the count once per five second window instead. The message still says something useful, a burst still shows up, and the flood is gone.

The count is printed through the `no_syslog` primitives, as the original message was. That matters: `UIP_LOG` is `print_string_newline_no_syslog`, and routing the new line through the ordinary printer would move the flood off the console and into the syslog ring instead of ending it.

### Hardware evidence

Measured on an RTL8373 board, `MACHINE=PCB_SWTG018AS_V2_1_0`, against upstream main `88454ec`.

Input rate, captured independently on the wire at the neighbouring switch so the measurement does not depend on the firmware: 380 fragmented frames in 20 s, about 19/s, from three sources.

Before, from #491: 156 of 156 console lines in a twelve second capture were `ip: fragment dropped.`

After, a 60 s console capture at 115200 8N1:

```
total console lines:      12
summary lines:            12
counts: sum=1121  mean=93  max=100
```

Twelve lines rather than roughly eleven hundred, one per five second window, and the counts sum to 18.7/s which agrees with the 19/s measured on the wire, so nothing is being lost from the count.

### Size

`BANK1` goes from 47994 to 48141 bytes, so 147 bytes, leaving 1011 free. `BANK2` grows by 6. `BANK3` and `CSEG` are unchanged.

### One thing worth a second opinion

The counter is a `u8` and saturates at 255, which is about four times the rate seen here. A window that busy would under-report rather than mislead, since the line still fires, but if you would rather it were a `u16` or printed something like `255+` when it saturates, say so and I will change it.


uip/uip.c
```diff
@@ -915,7 +919,26 @@ uip_process(u8_t flag) __banked
...
+      if(nz)
+        write_char_no_syslog('0' + d);
+      d = (fragdrop_count / 10) % 10;
+      nz |= d;
+      if(nz)
+        write_char_no_syslog('0' + d);
+      write_char_no_syslog('0' + fragdrop_count % 10);
+      write_char_no_syslog('\n');
+      fragdrop_count = 0;
+    }
```
DrDoof: One thing I noticed reading this: it is the only place that touches the counter, so a count only reaches the console when the next fragment shows up. If a burst stops, whatever is left in fragdrop_count sits there until something fragments again, and then prints as if it belonged to the window that just closed. Fragments rarer than the u16 wrap, around five and a half minutes, can fall back inside the window and be held the same way.

There is somewhere to put it, though. uip_process already has a periodic path, flag == UIP_TIMER, which handle_tx reaches every pass, and flushing a non-zero count there would cover both.

On your u8 question: at the rate you measured it has plenty of headroom and saturating fails gently, so I would keep it. What might help is making it visible, something like 255+ when it is pinned.

DrDoof (commented):
I like this one. What stands out is keeping it on the no_syslog primitives - going through the ordinary printer would have quietly moved the flood into the syslog ring instead of ending it, and that is easy to miss. The drop path also comes out cheaper than before, since most fragments now cost an increment and a compare rather than a whole line.

I went looking for a decimal printer you could have reused and there isn't one. print_byte, print_short and print_long are all hex, so spelling the digits out really was the only way short of adding one.

bloqaudio:
Both good catches, thank you. Fixed in `02aedbd`.

You are right that the counter was only ever touched on the drop path. A residual count waited for the next fragment and then printed as though it belonged to the window that had just closed, and past the `u16` wrap at about five and a half minutes it could fall back inside the window and be held again. The flush now lives in the `flag == UIP_TIMER` path, which `handle_tx()` reaches every pass, and the drop path is a bare increment. That also makes it cheaper again, which was your other observation.

On the `u8`: kept, with the marker you suggested. A pinned window now prints `255+`.

Hardware, on an RTL8373 board against upstream main:

```
60 s sample, ambient traffic only
  12 summary lines, one per five second window
  counts sum 1395, mean 116, max 119   (23.2/s)

counter pinned on purpose, 475 oversized UDP broadcasts from another host
  ip: fragments dropped: 115
  ip: fragments dropped: 255+
  ip: fragments dropped: 255+
  ip: fragments dropped: 255+
  ip: fragments dropped: 108
  ip: fragments dropped: 105
```

One thing I did not measure, so I will not claim it: the tail flush itself. The fragment source here never stops long enough for the counter to go idle, so what I can show is that every line now comes out of the timer path and that the marker works. The case you described, a count printed promptly after a burst ends, is reasoning from the code rather than something I put on a scope.

`BANK1` is now 48159 against 47994 on main, so 165 bytes, 993 free.

feelfree69:
> I went looking for a decimal printer you could have reused and there isn't one. print_byte, print_short and print_long are all hex, so spelling the digits out really was the only way short of adding one.

What about creating such a decimal-print function and using it? Otherwise I assume it won't take long until the next developer creates a similar function....

PR #504, opened 2026-09-22
Very experimental telnet console

jpsutton:
## Telnet console with optional TOTP 2FA, plus SNTP, `help`, and `save`

Adds a telnet management console that shares the existing command line
with the serial console and the httpd `/cmd` endpoint, so anything you
can type on the UART now works over the network.

### What's included

- **Telnet server** (`telnetd.c`, port 23, single session). Negotiates
  character-at-a-time mode (`WILL ECHO` + `WILL SGA`) and does its own
  echo and line editing, so the password is never echoed. Login uses the
  existing admin password.
- **Optional TOTP second factor** (`totp.c`, RFC 6238; HMAC-SHA1, 6
  digits, 30s step, ±1 window). Verified against the RFC 6238 test
  vectors. Fails closed: if TOTP is enabled but the clock is not yet
  set, login is refused rather than falling back to password-only. The
  web UI has no TOTP gate and remains the recovery path.
- **SNTP client** (`sntp.c`, RFC 4330). The RTL8373 has no RTC, so wall
  time only exists after a sync; TOTP depends on it. `ntp <server-ip>`,
  10s retry until synced, 4h resync.
- **`telnet bind <ip|any>`** accepts connections only while the switch's
  own address matches, so the service can follow a specific L3 address
  (useful with DHCP). **`telnet timeout <secs>`** sets the idle close.
- **`help`** lists all commands, grouped by topic.
- **`save` / `save show`** (`cfgstore.c`) persist the running config to
  the startup-config sector from the CLI. The firmware keeps an in-RAM
  shadow of the startup config, seeded from flash at boot and after web
  config uploads, and updated per successful config command using the
  same overwrite/delete/toggle merge rules as the web UI's save button,
  so both save paths agree on what a startup config can contain.
- **Web UI**: telnet enable/bind controls on the System tab, telnet
  fields in `information.json`, and the config-command whitelists
  extended for the new commands.

### uIP changes

- `UIP_CONF_MAX_LISTENPORTS` raised from 1 to 2 (httpd on 80, telnet on
  23). `uip_listen()` fails silently when no listen slot is free, so
  this is required.
- `tcp_appcall()` dispatches incoming connections by local port between
  the httpd and the telnet server.
- The idle reaper (`UIP_CONF_IDLE_TIMEOUT`) frees the scarce connection
  slot from abandoned httpd connections after ~30s, which also killed
  long-lived telnet sessions. Added `UIP_IDLE_EXEMPT_LPORT` to exempt
  the telnet port; the telnet server enforces its own, longer idle
  timeout instead. httpd behaviour is unchanged.

### Notes

- SHA-1 and the config-merge engine are written around the 8051's
  near-full internal RAM: xdata statics, `__reentrant` helpers, and
  byte-wise 32-bit rotates avoid DSEG/OSEG spills under SDCC.
- Idle timing snapshots the 4-byte tick counter with interrupts masked
  to avoid a torn read closing sessions early.

### Testing

Built and tested on a Horaco ZX-SWTGW218AS (PCB SWTG118AS-V2.1,
RTL8373). Verified: telnet auth, TOTP login (device codes match an
independent implementation) and fail-closed behaviour, bind
enforcement, idle timeout, `save` round-trips across reboot, `help`,
and that httpd is unaffected. TOTP and the config-merge logic also have
standalone host-compiled tests against known vectors.

This is a fairly large multi-feature branch. Happy to split it into
separate PRs (telnet, TOTP+SNTP, `help`, `save`) if you'd prefer to
review them independently.


uip/uip-conf.h
```diff
@@ -111,12 +111,18 @@ typedef unsigned short uip_stats_t;
...
+#define UIP_IDLE_EXEMPT_LPORT 23
+
 /**
  * Maximum number of listening TCP ports. TODO: increase this!
  *
  * \hideinitializer
  */
-#define UIP_CONF_MAX_LISTENPORTS 1
+/* One slot for the httpd (80) and one for the telnet server (23) */
+#define UIP_CONF_MAX_LISTENPORTS 2
```
DrDoof, line 125: One thing to watch here: UIP_CONF_MAX_CONNECTIONS is still 1 just above, so this gives two listening ports but only one connection slot. A telnet session then sits on the only slot the web UI can use, and since port 23 is exempt from the reaper it holds it for the whole idle timeout, 600s by default or 120s for a session that never logs in, where an abandoned httpd connection used to be reclaimed after 30.

On doing the transport with uIP's own facilities: psock.c and pt.h are in the tree but not in the Makefile, so reaching for them means adding a file plus the switch-based protothreads, which is awkward with banked code and a tight OSEG. From the diff only a dozen or so lines here are transport anyway. IAC negotiation, echo and auth are the bulk, and psock would not help with those.

DrDoof (commented):
That is a lot of ground for one branch, and I think the split you offer would really help. Telnet, TOTP, help and save all stand on their own, so each could land or stall on its own merits instead of waiting for the others.

One overlap you may not know about: the SNTP client already exists as #447, which has been sitting for a while now. That one carries timezones, DST, a Time tab and a time command, and it builds on the resolver in #446 so the server can be a name rather than an address. Both versions touch udp_apps.c, cmd_parser.c and rtlplayground.c, so whichever goes in second will need reworking. Dropping sntp.c here and rebasing on #447 would spare you that.

jpsutton:
> That is a lot of ground for one branch, and I think the split you offer would really help. Telnet, TOTP, help and save all stand on their own, so each could land or stall on its own merits instead of waiting for the others.
> 
> One overlap you may not know about: the SNTP client already exists as #447, which has been sitting for a while now. That one carries timezones, DST, a Time tab and a time command, and it builds on the resolver in #446 so the server can be a name rather than an address. Both versions touch udp_apps.c, cmd_parser.c and rtlplayground.c, so whichever goes in second will need reworking. Dropping sntp.c here and rebasing on #447 would spare you that.

I'll work to split the individual features and submit separate PRs for each, if that's amenable to you all.

vDorst:
I personally feels that telnet at the moment is not a feature we want. Web-console is a similar feature at that is now buggy and not ideal. Adding yet an other option to access the console, without web-console and uart-console is working properly, feels not helpful.

So I personally want to hold this back, until basic are working correctly.
I am planning to start a roadmap discussion what to add in the next release `v0.2` and what maybe put in the `v-0.next`-release.
I see of all the new features, but also that the foundation is not right to support them nicely and properly, like PHY, console, and time-handling. I think the focus towards v0.2 should be refactor the foundation of the project so we can add new features better with the stable API to interact with the core.

For SNTP we already have a PR #447.
Please look at that one first and maybe review/collaborate.

PR #508, opened 2026-09-22
build: rebuild objects when the compiler flags change

DrDoof:
This one bit someone in #352, and it took a while to work out why.

The `.d` files track headers but not flags, so adding `CI=1` to a tree that was already built recompiles nothing. The objects keep whatever flags they had, and the image ends up a mix of the two.

What that looks like depends on the flag. `DIAG=1` fails the link on undefined symbols, which is annoying but at least stops you. An opt-in whose hooks expand to empty macros is much quieter: the calls disappear rather than dangle, the link is perfectly happy, and the feature simply measures nothing. In #352 the command was there, the counters sat at zero, and nothing anywhere said why.

So objects now depend on a small stamp file holding `CC_FLAGS`, rewritten only when the flags really differ:

```
$(CCFLAGS_STAMP): FORCE | create_build_dir
	@echo '$(CC_FLAGS)' | cmp -s - $@ 2>/dev/null || echo '$(CC_FLAGS)' > $@
```

Building `boot.rel`, then again with `CI=1`, then once more with nothing changed:

| | first | after flag change | repeated |
|---|---|---|---|
| before | compiled | skipped | skipped |
| after | compiled | compiled | skipped |

A changed flag set rebuilds, an unchanged one still does not, so incremental builds stay worth having.


feelfree69:
Tested on top of current main. Works as expected. Approving.

feelfree69 (approved):

DrDoof:
Thanks for review.

PR #509, opened 2026-09-22
phy: advertise pause on the forced speeds from 1G up

DrDoof:
The advertisement written for a forced speed of 1G and above is `0x1001`, while auto and the forced 10M and 100M modes all carry bit 10. Forcing a port to 1g or faster therefore takes the pause capability off the link.

Measured on a SWTGW218AS, on a port with no link so nothing was disturbed:

```
auto          7.0x10 = 0x15e1
port 1 1g     7.0x10 = 0x1001
port 1 auto   7.0x10 = 0x15e1
```

Writing `0x1401` by hand with `physet` sticks, so the PHY takes the value.

That bit 10 is the pause capability shows in a second reading. Two ports advertise the same `0x15e1` from our side, and only the partners differ:

| port | partner advertises | pause in 0x63FC |
|---|---|---|
| 5 | `0xd041`, bit 10 clear | inactive |
| 3 | `0xdde1`, bit 10 set | active |

Pause needs both ends to offer it, and it is active exactly where the partner does.

Confirmed on a second board by feelfree69, who found this in #489: with this branch a forced 2g5 keeps pause at TX RX, where before it fell to none. That board has a link partner advertising pause, which mine did not.


feelfree69 (approved):
Tested in hardware, TX RX Pause setting stays enabled also when a port is set to fixed speed setting.

PR #511, opened 2026-09-23
rtl8224: wait for the PHY to answer instead of a fixed 2.5 s

DrDoof:
While timing how long the switch takes to boot, I noticed that `rtl8224_enable()` waits a flat 2.5 s after releasing the RTL8224 reset. On my SWTGW218AS the chip was already answering with `0x8224` the first time I asked, about 265 ms after the release, so it looked like most of that wait could go.

The change remembers when the reset was released and, once SMI is up, asks for the chip id until it answers. If it never does, it gives up after the same 2.5 s as today, so a slower PHY on another board shouldn't get any less time than it does now.

Since I was there anyway, I also put a limit on the busy-bit wait in `phy_config_8224()`. As far as I can tell, a PHY that never answered would otherwise stop the boot before the network is up; now it prints a message and carries on.

A rough before/after on my board, timed from `reset` on the console, two runs each:

```
                      before     after
version banner        3.21 s     0.70 s
first ping answered  27.34 s    24.78 s
```

So the boot ends up about 2.5 s shorter. As a sanity check I read back the two PHY registers set during init after each boot (`0x7b20` = `0x0bed`, `0x0a90` = `0x00fc`) and they match what I get without the change. Link speeds stayed the same, and traffic went through ports in both groups of four.

Following vDorst's note that the related RTL8372 and RTL8221B need about 10 ms, the reset pulse itself also goes from 0.5 s to 20 ms. To check that the chip still gets reset and is not just left configured from before, a test build printed the MDI reverse register (MMD 30, `0x0a90`) right before the init writes it: it reads the power-on value `0x00f3` with 20 ms on three boots, same as with 0.5 s. That brings the banner to 0.20 s; the first ping gains another 0.3 s or so (24.5 s), as STP decides when it comes.

It costs BANK2 +195, BANK3 +22, CSEG +21 and XSEG +4 bytes on the SWTGW218AS. It builds with `CI=1` for the CI machines and the host tests pass.

I couldn't find any published reset timing for the RTL8224, so I'd rather not read too much into the 265 ms - it is just when this board was first asked, and the chip may well be ready earlier. That's also why it polls with a timeout instead of picking a shorter fixed delay. Happy to change it if you'd prefer another approach.


vDorst:
> I couldn't find any published reset timing for the RTL8224, so I'd rather not read too much into the 265 ms - it is just when this board was first asked, and the chip may well be ready earlier. That's also why it polls with a timeout instead of picking a shorter fixed delay. Happy to change it if you'd prefer another approach.

RTL8372 N and non-N, RTL8221B need around 10mS. So it can be reduced if that is useful.
The rest is looking good.

vDorst (approved):

DrDoof:
That was the missing piece. I shortened the reset pulse from 0.5 s to 20 ms, twice your 10 ms, and tried it on the SWTGW218AS.

To make sure the chip really resets and is not simply still configured from before, I printed the MDI reverse register (MMD 30, 0x0a90) right before the init writes it. With 20 ms it reads the power-on value 0x00f3 on all three boots, same as with 0.5 s. After init it is 0x00fc, 0x7b20 is 0x0bed, the LACP trunk comes up and traffic goes through both port groups.

The console banner now shows 0.20 s after `reset`, against 0.70 s with the 0.5 s pulse and 3.21 s before this PR. The first ping only moves by about 0.3 s, since STP decides that one.

It is folded into the same single commit.

vDorst:
Thanks

PR #513, opened 2026-09-23
cmd: "port <n> name" without a name clears it instead of reading stale words

DrDoof:
The name branch of `port <n> name` read the fourth word without checking that the line had one, and the tokenizer does not reset unused word slots. So `port 5 name` took the name from wherever the previous command's fourth word had been: `port 5 name hello` followed by `port 5 name` gave `hello` again, and after a longer line it gave the tail of that line.

The web UI never sends the bare form, the console can. With no argument the name is now empty, which also gives the console a way to remove a name. One condition added to the copy loop; `bank-audit` clean.

Checked on SWTGW218AS from the console: before, `port 5 name hello` then `port 5 name` printed `Name set to: "hello"`; after, it prints `Name set to: ""` and `port 5 show` reports an empty name. A name with an argument is set as before.


feelfree69 (approved):

vDorst:
Thanks

